PR #6, opened 2019-05-11
Fix Null Reference Exception

md7marcu:


sbwalker:
Can you explain how to reproduce the null reference exception? I have never encountered this myself and I would like to understand if there is bigger issue before we introduce code to handle the exception

sbwalker:
So this happens when the Url is modified manually and it does not match a page name in the system.

sbwalker:
I am going to close this pull request and log it as an Issue related to 404 handling

PR #11, opened 2019-05-18
ThemeService Improvements

### LINKED ISSUE #8: Cleanup Reused/Duplicated Code Relating to Themes
Currently, in the Client library, multiple .cshtml files are performing the same actions with the results from the IThemeService (Formerly ISkinService).

These operations should be condensed and moved to a method implementation inside of the IThemeService to prevent fragility in management of the code-base going forward.

Items to consider moving to a method implementation:

* Available Themes
* Available Containers

This will substantially reduce the on-view code levels greatly. 

I'll submit a PR for this, after my other PR is processed, but wanted to get this documented.

mitchelsellers:
Centralized the code used to determine the available themes, panes, and containers.  This isn't the most "ideal" way to handle this.  However, it does improve the management of the code by centralizing the logic for theme selection.

Future PR's development might improve this more.

Resolves #8


sbwalker:
This is a nice refactoring to consolidate the theme logic. The only request I would make is a change to the method naming convention used. I think the "CalculateSelectable" prefix is confusing as it implies extra processing or filtering which is not really accurate. I think it would be more intuitive for developers if they were simply named GetThemes, GetContainers, GetPaneLayouts. Please let me know your thoughts.

mitchelsellers:
My reason for calculate prefix is that right now it is iterating the collection, splitting things out etc. it isn’t a simple data retrieval. 

But I can easily change it.

sbwalker:
You are correct that the internals of the methods are not just doing simple data retrieval - however the goal of a framework should be to hide those internal implementation details from module developers. A module developer should simply be able call a method to get a list of themes/containers/panelayouts and not care about how the framework stores them internally. Similarly, I am not sure what the term "Selectable" means in this context - it implies some sort of user interface activity and also hints that there may be items which have been excluded from the list as they are not selectable. In reality the items returned from these methods represent all of the themes/containers/panelayouts that are installed and available in the system.

mitchelsellers:
The reason for selectable is that we are giving a list of themes that could be used. The existing “GetThemes” method for example is more of a theme package rather than a theme that could be used. 

Maybe rename that one to “getThemePackage” and the others become named as you mention?

sbwalker:
Good point... the current GetThemesAsync calls the server which iterates the assemblies in the /bin to find all of the themes. In reality a Theme is comprised of multiple Theme components, Container components, and optionally PaneLayout components. The components are identified by their type name. So when a developer asks for a list of Themes they are actually requesting a list of Theme types. So maybe the new methods should be suffixed with Types for clarity ( which would be consistent with the naming convention used in the entity models as well )

mitchelsellers:
I can adjust as desired. That might be the best. 

What would your recommended names be?

sbwalker:
My preference is GetThemeTypes, GetContainerTypes, GetPaneLayoutTypes - and just leave the original method name the same for now.

mitchelsellers:
Ok, names have been changed as requested

sbwalker:
Thank you Mitch!

PR #27, opened 2019-06-13
Add alias for running in vs-code

iJungleboy:
This enables it to run with VS Code :)
The alias is required because otherwise an endpoint fails to respond


sbwalker:
there was a better way to solve this problem by updating the lanchsettings.json file so that the port number was consistent for both IIS Express and the .NET Core web server. It does not require a database update.

PR #51, opened 2019-08-14
Support for third party modules, improved error handling, standardardized enum naming, reorganized interface definitions, support for DB script upgrades, added Settings entity

sbwalker:


iJungleboy:
wow! that's a big one ;) 
awesome 👍

PR #58, opened 2019-08-16
Added Swagger support for API

sbwalker:


ADefWebserver:
Thanks for this! So important for people like me who use Microsoft Flow a lot. Being able to call an API to consume data and resources and trigger actions is *a requirment*. A Swagger page is also essential for debugging issues and discovering functionality.

PR #87, opened 2019-09-19
Add basic support for SoftDeletable

fileman:


sbwalker:
I like the concept of centralizing soft delete support in the framework; however, I have a few questions/comments. After reviewing your suggestion and considering the desired functionality, doesn't it make sense to consider Delete activities to also be "Audit" activities? So rather than creating a new interface as you have suggested, could we not simply expand the existing IAuditable interface to include the additional deletion information? I guess the problem would be that if we took this approach, all existing entities which implement IAuditable would need to be updated at the same time. But in the bigger picture wouldnt this approach be more optimal for developers in the future? One potential downside is that developers would need to include these fields in their entities even if those entities did not support the soft delete concept - but is there ever a case to not support the soft delete concept?

sbwalker:
I am also interested on hearing your thoughts about querying the soft deleted entities... should the framework only include return "active" entities when it does a Get ( this is something that can be controlled centrally as well in DBContext ). Or should it return all entities and it should be up to the developer to do supplemental filtering using Linq to remove deleted items? Personally I am more in favor of the latter as I prefer to have more control, but I would like to hear your thoughts.

fileman:
As have you sayd, add another interface can give at the developers the possibility to choose if entity can be deleted or softdeleted depending on application domain requirements, it's a little more complex but with more flexibility, instead adding to IAuditable those properties add a minimal DB size increase even if they are nullable, but I don't think this can be a reason to exclude this solution.

As example in #88 all entities was returned only to user in admin role (better approach is with a policy like "CanRestorePages") by excluding efcore globalqueryfilter, so other users only gets the active pages, in wasm client less network resource required when hosted in cloud instead on intranet.

PS: DbContextBase should be improved to skip setting to null properties when the value of IsSoftDeleted wasn't modified, something like:
`if ((bool)item.CurrentValues[nameof(ISoftDeletable.IsSoftDeleted)] 
                        && !(bool)item.OriginalValues[nameof(ISoftDeletable.IsSoftDeleted)])`

sbwalker:
Sorry for the delay in getting back to this. After some consideration, my preferred approach would be to introduce a new interface ( like you suggested ) in the Shared Models and I believe it should be called IDeletable ( I dont think it is necessary to include the word "soft" as a "hard" delete essentially means the record no longer exists ). So I would use the property IsDeleted rather than IsSoftDeleted as well. 

As far as querying entities is concerned I would prefer to not introduce any filters at the repository level - I would prefer to return the full collection of entities and allow the developer to filter them if necessary based on their specific requirements. I realize this will create a bit more payload however the broader performance goal should be to reduce the number of service calls in the application rather than worrying about payload. Fpr example in the case of Pages we are storing the full collection of Pages for the Site in the PageState which is essentially a cache that can be used by any module so that they do not need to make a service call to retrieve them.  

If you are agreeable to this approach, please modify your pull request and I will merge it.

ADefWebserver:
> I would prefer to return the full collection of entities and allow the developer to filter them if necessary based on their specific requirements

I fully agree with this. It allows all use cases to be covered where the alternative would not.

fileman:
In this PR there isn't modify to repository, since is in #88 however I have closed it because there some conflict to resolve.

sbwalker:
Sorry I just realized that my guidance was not very clear... I relocated IDeletable to the Oqtane.Models namespace in Models to be consistent with IAuditable. Thank you so much for your help... do you plan to add a new implementation for Pages?

fileman:
I didn't say anything because I thought you wanted to move IAuditable too.
Yes I'm going to implement this for pages, without repository changes

PR #103, opened 2019-09-29
IDeletable implementation for Page

fileman:


fileman:
In Delete.razor, PermissionGrid should be:
1. removed;
2. readonly add a div with style="pointer-events: none; backgroud-color: same_color_of_disabled";

sbwalker:
Good question. I think the permission grid should probably be removed from the Delete component, as it does not really serve any useful purpose.

fileman:
After changing IsDeleted value, I've to reload from browser to hide/show page from menu, there is a method to reload the Menu.razor component or we need it?

sbwalker:
I think the problem might be that the Menu component is currently using the OnInitialized event rather than the OnParametersSet event. You can try changing it and test to see if it resolves the problem.

fileman:
You are right, changed in #104

PR #111, opened 2019-10-04
Added ability to install modules and skins at run-time directly from Nuget

sbwalker:


ADefWebserver:
This is a historical moment... yes I know many won't understand that right now, but, I plan to start blogging about what just happened here on BlazorHelpWebsite.com in November...

ADefWebserver:
From: Shaun Walker 
Sent: Friday, October 04, 2019 1:38 PM
Subject: Nuget Integration

You will like this PR: “Added ability to install modules and skins at run-time directly from Nuget”

Basically the way it works is it queries the public Nuget repository for any packages which contain the “oqtane” tag. Then depending on if you are in the Module or Theme area in the application, it filters the results based on whether the package has a “module” or “theme” tag. It displays the list in order by popularity ( download count ) and allows you to Download any package. The package is downloaded from Nuget and placed in the appropriate framework folder where it can be immediately installed by clicking the Install button. This makes it really easy to install extensions.

PR #122, opened 2019-10-12
Fix and Improve DeletePage

fileman:
Fix UpdatePermissions instead of DeletePermissions in DeletePage and DeleteModule

Improve Page Delete to remove even modules, permissiones and pagemodules


sbwalker:
Although its not fully implemented, the architecture supports the ability for a Module to be displayed on multiple Pages. So when a Page is deleted you cannot assume that the Module should also be deleted - as that Module may be linked to another Page in the site.

sbwalker:
The latest PR I submitted contains an enhancement to the Control Panel for adding an existing module to a page. You will now be able to test your code against all expected scenarios.

PR #133, opened 2019-10-16
Page module deletable

fileman:
Enable and Disable Page Module for SoftDelete.
In Page Edit Mode show gray (css) background if disabled.

### DIFF
--- a/Oqtane.Client/Shared/ModuleInstance.razor
+++ b/Oqtane.Client/Shared/ModuleInstance.razor
@@ -67,7 +67,7 @@
 
     public void HideProgressIndicator()
     {
-        progressindicator = true;
+        progressindicator = false;
         StateHasChanged();
     }
 }
--- a/Oqtane.Client/Shared/Pane.razor
+++ b/Oqtane.Client/Shared/Pane.razor
@@ -117,7 +117,7 @@
                 }
                 else
                 {
-                    foreach (Module module in PageState.Modules.Where(item => item.Pane.ToLower() == Name.ToLower()).OrderBy(x => x.Order).ToArray())
+                    foreach (Module module in PageState.Modules.Where(item => item.Pane.ToLower() == Name.ToLower() && !item.IsDeleted).OrderBy(x => x.Order).ToArray())
                     {
                         // check if user is authorized to view module
                         if (UserSecurity.IsAuthorized(PageState.User, "View", module.Permissions))
--- a/Oqtane.Client/Themes/Controls/ModuleActions.razor
+++ b/Oqtane.Client/Themes/Controls/ModuleActions.razor
@@ -23,7 +23,7 @@
 }
 
 @code {
-    List<ActionViewModel> actions;
+        List<ActionViewModel> actions;
 
     protected override void OnParametersSet()
     {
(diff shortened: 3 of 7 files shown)


sbwalker:
The files that are in the Oqtane.Client/Themes folder are "themes" which are intended to be simple templates for rendering the look/feel of a page. There should be no programming logic in these theme files - they should only contain markup and components. The filtering of PageModules should occur in the Oqtane.Client/Shared/Pane.razor component - where the modules are dynamically injected into a pane. Personally I do not think there should be any connection to EditMode - once a PageModule is deleted it should never be rendered into a pane, regardless of Mode. In the future it would make sense for there to be a Recycle Bin in the Admin Dashboard which allows you to see a list of deleted Pages and a list of deleted PageModules, and allows you to undelete them or hard delete them.

fileman:
I'll try to create di "recycle bin" for deleted page modules, for you is better to place in Page/Edit.razor a control with her deleted modules or an independent section in admin dashboard with deleted modules of all pages?

sbwalker:
From a conceptual perspective, think of the Windows file system. When you delete a file from the file system it disappears from all user interfaces which expose files. However, you can go to the Recycle Bin on your desktop and there you can see all of the deleted files. The Recycle Bin provides the option to restore a file or permanently delete a file. I can envision the same behavior for the Page and Module assets in Oqtane ( and in fact this is the approach used in DNN - I can upload a screen shot of its Recycle Bin for more context )

sbwalker:
recyclebin

sbwalker:
Perhaps it would help if I explained the expected behavior ;)

Pages
- View Pages - should only display pages which are not deleted.
- Delete Page - when a user is in this component and they select the Delete option, the system should set the IsDeleted property to True. It should not modify or delete any dependent objects such as permissions, page modules, settings, etc...  
Edit Page - there is no way to edit a deleted page, so there is no need to display the Deleted? field in this view. Which means you do not need to deal with delete/undelete logic in this component.
- Recycle Bin - there should be a list ( it could be a pageable grid as you likely want to display the Name, DeletedOn, and DeletedBy columns ) which displays only the deleted pages. There should be an option to Restore a page or Delete a page. Restore would set the IsDeleted property to False. Delete would actually hard delete the object and would need to also remove any dependent objects.

Modules
- View Page - should only display modules for the page which are not deleted.
- Delete Module - when a user selects the Delete Module option from the Action menu when viewing the page in Edit Mode, the system should set the IsDeleted property on the PageModule to True. It should not modify or delete any dependent objects.
- Recycle Bin - there should be a list ( it could be a pageable grid as you likely want to display the Page Name, Module Title, DeletedOn, and DeletedBy columns ) which displays only the deleted modules. There should be an option to Restore a module or Delete a module. Restore would set the IsDeleted property to False for the PageModule. Delete would actually hard delete the PageModule object and would need to also remove any dependent objects - *Note* - due to the many-to-many relationship between Pages and Modules via the PageModule table, there are a few additional scenarios to consider. When deleting a module in the Recycle Bin you need to check if there are any remaining pages which are associated to the ModuleId. If there are none then it is safe to hard delete the module from the Module table ( and all dependent objects ). If it is linked to other pages then the only record you can hard delete is the PageModule record which was selected - leaving the Module record untouched.

Issue #105, opened 2019-09-30
Need ability to soft delete core entities


fileman:
In addition to User, Roles and Permissions, are these the other core entities?
- [ ] Sites
- [x] Pages
- [x] PageModules
- [ ] Modules
- [ ] Users

Something to add or something to remove?

sbwalker:
Sites, Pages, PageModules, Modules, Users

fileman:
For Users, you prefer `[Required]` attribute in Model or `if(string.IsNullOrWhiteSpace(user.DisplayName))
{
user.DisplayName = user.Username;
            }`

fileman:
#114 I've opted for handle in the pages, it looks more aligned to other properties in oqtane

sbwalker:
Moving this comment from a PR to this issue so that it does not get lost:

Perhaps it would help if I explained the expected behavior ;)

Pages

View Pages - should only display pages which are not deleted.
Delete Page - when a user is in this component and they select the Delete option, the system should set the IsDeleted property to True. It should not modify or delete any dependent objects such as permissions, page modules, settings, etc...
Edit Page - there is no way to edit a deleted page, so there is no need to display the Deleted? field in this view. Which means you do not need to deal with delete/undelete logic in this component.
Recycle Bin - there should be a list ( it could be a pageable grid as you likely want to display the Name, DeletedOn, and DeletedBy columns ) which displays only the deleted pages. There should be an option to Restore a page or Delete a page. Restore would set the IsDeleted property to False. Delete would actually hard delete the object and would need to also remove any dependent objects.
Modules

View Page - should only display modules for the page which are not deleted.
Delete Module - when a user selects the Delete Module option from the Action menu when viewing the page in Edit Mode, the system should set the IsDeleted property on the PageModule to True. It should not modify or delete any dependent objects.
[...]

PR #145, opened 2019-10-19
Improved recycle bin

fileman:
this should complete the #105 with the expected behavior

### DIFF
--- a/Oqtane.Client/Modules/Admin/RecycleBin/Index.razor
+++ b/Oqtane.Client/Modules/Admin/RecycleBin/Index.razor
@@ -60,9 +60,7 @@
 
 @code {
     public override SecurityAccessLevel SecurityAccessLevel { get { return SecurityAccessLevel.Admin; } }
-
-    bool loadentities { get; set; }
-
+        
     List<Page> pages { get; set; }
     List<PageModule> pageModules { get; set; }
 
@@ -70,9 +68,6 @@
     {
         pages = new List<Page>();
         pageModules = new List<PageModule>();
-
-        loadentities = true;
-
         await LoadEntities();
     }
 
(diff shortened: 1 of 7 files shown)


sbwalker:
Thank you for the PR... a few comments:

- currently the /controls folder is intended for generic components which are applicable for use in any module. The TabControl and TabPage components do appear to be generic so this seems like a good approach.
- So far the user experience for Oqtane does not utilize any Tabs however it makes sense in scenarios where there are a lot of options to represent in a component that it would be useful to organize/group them. 
- Oqtane already has an entity named "Page" and it creates confusion if the "Page" term is used for other purposes. Another suitable alternative in this case may be "TabPanel"?
- Oqtane relies on Bootstrap 4 for its global styles. By utilizing such a widely used style system it makes the product much simpler for designers and developers. However, this also means that the product should implement the Bootstrap standards as closely as possible and not introduce custom options when a suitable Bootstrap option already exists. In this regard it would make sense for the generic Tab to be based on the Bootstrap Tabs nav https://getbootstrap.com/docs/4.0/components/navs/#tabs ( with perhaps a Parameter so that a custom class could be substituted - "nav-tabs" or "nav-pills" )

fileman:
I've tried with bootstrap nav-tabs, but without success, they are not rendered even if the elements are present in DOM.
I've added Tab* controls as a workaround to show the result of recyclebin behavior, if it is ok I'll try to use only bootstrap.

sbwalker:
Sorry... the problem is that Bootstrap tabs relies on another JavaScript plugin https://getbootstrap.com/docs/4.0/components/navs/#javascript-behavior .- I thought it only depended on jQuery. So... we have a dilemma... add another JavaScript reference, or create a Blazor native Tab control...

sbwalker:
In general my opinion is to use native Blazor functionality rather than JavaScript when possible as this is one of the big advantages of Blazor.

sbwalker:
So perhaps just rename TabPage to TabPanel

fileman:
I've experimented RecycleBin with bootstrap card ... I don't like it very much, but if you want I can submit a new PR to show the result

fileman:
You can find it [here](https://github.com/fileman/oqtane.framework/tree/RecycleBinWithBootstrap)

sbwalker:
I moved the controls to the Modules/Controls folder as these components are intended to be used in Modules - not Themes.

PR #152, opened 2019-10-23
fix for logmanager nullexception when values is null

fileman:
I've found it when I try to add a page with null parent

### DIFF
--- a/Oqtane.Server/Infrastructure/LogManager.cs
+++ b/Oqtane.Server/Infrastructure/LogManager.cs
@@ -97,7 +97,7 @@ namespace Oqtane.Infrastructure
                         {
                             if (values[names.Count - 1] != null)
                             {
-                                message = message.Replace("{" + names[names.Count - 1] + "}", values[names.Count - 1].ToString());
+                                message = message.Replace("{" + names[names.Count - 1] + "}", names[names.Count - 1] + ":" + values[names.Count - 1].ToString());
                             }
                             else
                             {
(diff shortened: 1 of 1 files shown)


sbwalker:
Good catch! I have been thinking about this scenario and the problem is that null is sometimes a meaningful value. For example in the case you identified, having a parentid of null means that a page is at the top of the page hierarchy. The problem with the current code is that it tries to perform a ToString() on properties which are passed to the logger - which throws an exception if the property value is null. However null is actually a valid value in JSON ( ie. {"property": null} ) and null is a distinct and different value than empty string "". So I think the correct fix may be to check if the value is null and if so, substitute its value with "null" in the message and serialize it as null in the properties. I will experiment with this this morning.

fileman:
A simple solution can be this, in event log you will see "Page Order Updated 1 with Null ParentId" and "Page Added {"PageId":0,"SiteId":1,"ParentId":null,"Name":"Page2",....}"

fileman:
with last commint log messagge should be rewrited as example from "Page Order Updated {SiteId} {ParentId}" to "{SiteId} {ParentId} Pages Order Updated"

sbwalker:
When using structured logging it is up to the developer to construct the message as they see fit - the framework should not add any additional words, etc... - it should only substitute the {tokens} with the property values. So I like the changes you made however it should not add the word "with" or "token:". When populating the Message, if the property is null then it should substitute the {token} with the string "null" - this will eliminate the exception caused by calling ToString() on a null object and it will provide the context to the person viewing the log that the value was null. In the logic where it rebuilds the property dictionary it is also calling ToString() on values and should not - if a value is null it should add it to the dictionary and then rely on the JsonSerializer to serialize the dictionary properly.

PR #155, opened 2019-10-23
added security attribute to TenantController Get methods and resolved TenantId on server during Installation

sbwalker:


Oqtane.Server/Infrastructure/LogManager.cs
```diff
@@ -95,27 +95,29 @@ private Log ProcessStructuredLog(Log Log)
                         names.Add(message.Substring(index + 1, message.IndexOf("}", index) - index - 1));
                         if (values.Length > (names.Count - 1))
                         {
-                            message = message.Replace("{" + names[names.Count - 1] + "}", values[names.Count - 1]?.ToString() ?? "null");
+                            if (values[names.Count - 1] == null)
```
fileman: Why you prefer an "IF" instead of "??" operator, readability or technical reason?

sbwalker:
Good question. I prefer the longer conditional form for readability, debugging, and maintenance. If I am skimming a chunk of code it is really easy to follow the flow if you see standard if conditions - the shorter form is more concise but its really easy to miss the embedded conditional logic ( and not all developers are familiar with the shorter syntax  - whereas all developers are familiar with the standard form ). That being said, I do use the shorted (x == y) ? 1: 0; sometimes ( usually in very simplistic scenarios ) - so even I am not 100% consistent.

fileman:
ok, I will try to remember it for the next PR

PR #183, opened 2020-01-03
Oqtane service collection extensions

hishamco:
This include:

- Add Oqtane `ServiceCollection` extension methods such as: `AddOqtaneModules()`, `AddOqtaneThemes()` .. etc

- Add missing theme types to the modules assemblies

### DIFF
--- a/Oqtane.Server/Extensions/AssemblyExtensions.cs
+++ b/Oqtane.Server/Extensions/AssemblyExtensions.cs
@@ -1,0 +1,35 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace System.Reflection
+{
+    public static class AssemblyExtensions
+    {
+        public static IEnumerable<Type> GetInterfaces<TInterfaceType>(this Assembly assembly)
+        {
+            if (assembly is null)
+            {
+                throw new ArgumentNullException(nameof(assembly));
+            }
+
+            return assembly.GetTypes(typeof(TInterfaceType));
+        }
+
+        public static IEnumerable<Type> GetTypes(this Assembly assembly, Type interfaceType)
+        {
+            if (assembly is null)
+            {
+                throw new ArgumentNullException(nameof(assembly));
+            }
+
+            if (interfaceType is null)
+            {
+                throw new ArgumentNullException(nameof(interfaceType));
+            }
+
+            return assembly.GetTypes()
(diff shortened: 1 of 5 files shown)


sbwalker:
this is a very nice refactoring to clean up and separate the startup logic

hishamco:
Thanks, hope you see more contribution in the upcoming future ;)

hishamco:
@sbwalker I forgot to tell you that during doing this refactoring I notice that the client-side project has part of what I did in the server-side project, is this by intention? I will submit another PR for fixing that too

sbwalker:
Yes this is intentional. Oqtane can be run in both server-side and client-side hosting models. To run in client-side you need to change your Solution Configuration from Debug to Wasm then recompile and run. This uses conditional compilation which is mostly restricted to the Startup methods

sbwalker:
After merging the changes as part of this PR I am having difficulty with my local environment. It is throwing errors when I run related to:

An unhandled exception of type 'System.ArgumentException' occurred in Microsoft.Extensions.DependencyInjection.dll
Cannot instantiate implementation type 'Microsoft.Extensions.Hosting.BackgroundService' for service type 'Microsoft.Extensions.Hosting.IHostedService'

sbwalker:
So the problem is that the refactored code:

var serviceTypes = assembly.GetTypes(hostedServiceType);

does not have the same behavior as the previous code:

                Type[] servicetypes = assembly.GetTypes()
                    .Where(item => item.GetInterfaces().Contains(typeof(IHostedService)))
                    .ToArray();

The GetTypes() overload used in the refactoring returns a much larger collection of types than the former code.

sbwalker:
resolved in #185

hishamco:
Sorry I just prepared a fit for that two days ago, unfortunately the Internet cable has an issue, so that's why I didn't push the pr

hishamco:
> Yes this is intentional.

So I need to do similar thing in the client-side as well

hishamco:
@sbwalker I just notice in that client-side project that there's no notion of registering themes, modules, hosted services. Is this intentionally or a bug?

Please confirm before push my pr

sbwalker:
This is intentional. The client-side project does not need to register themes, modules, or hosted services - only the Server project needs them, and the Server project always runs on the server ( whereas the Client project will either run on the server in server-side Blazor or in the browser in client-side Blazor ).

PR #186, opened 2020-01-12
Generic repository

### LINKED ISSUE #184: Introduce IRepository<TType>
IMHO introducing a `IRepository<TType>` will save time instead of creating `I{Type}Repository` for every entity

For instance `IJobRepository` will be:

```csharp
public interface IJobRepository : IRepository<Job>
{
}
```

instead of

```csharp
public interface IJobRepository
{
    IEnumerable<Job> GetJobs();
    Job AddJob(Job Job);
    Job UpdateJob(Job Job);
    Job GetJob(int JobId);
    void DeleteJob(int JobId);
}
```

or we could implement the generic repository directly from within `JobRepository`

Sometimes we may need to create custom on such `IUserRepository`:

```csharp
public interface IUserRepository<User> : IRepository<User>
{
    User GetUser(string Username);
}
```

hishamco:
Fixes #184

I just introduce the abstraction APIs for the repository pattern and implement it in two entities `Page` & `Tenant`

/cc @sbwalker


sbwalker:
When I review the code changes I am having a difficult time seeing the benefits. The amount of code in order to implement repository based data access has not been decreased or simplified. Perhaps the problem is that two examples ( Page, Tenant ) are not the most practical entities to demonstrate this concept?

hishamco:
> Perhaps the problem is that two examples ( Page, Tenant ) are not the most practical entities to demonstrate this concept?

Seems you are right, coz there are some extra logic to be done in those entities

Let me try to find suitable once and update the pr once the Internet become stable in my country

sbwalker:
Perhaps Role would be a good candidate as it is one of the the most straightforward entities.

hishamco:
You 're right it is very straightforward, but again we need to create a custom interface `IRoleRepository` because it has two custom methods, but in case the entity doesn't require custom methods it will be much simpler

hishamco:
I just find `JobRepository` prove what I mean in the last phrase, so `IJobRepository` is not required anymore

I will update my PR soon

hishamco:
Shall I revert Page & Tenant repositories changes or shall they go in the same PR if it looks good to you?

sbwalker:
The original hypothesis for this enhancement was that "introducing a generic repository will save time instead of creating I{Type}Repository for every entity". In order to prove this theory I suggested that a sample implementation be submitted for review. Based on the sample implementation it showed that the majority of entities do not conform to a simple CRUD pattern - almost every entity has additional methods or business logic which still require the creation of an I{Type}Repository interface. So the generic repository seems to have limited benefit as it will not save much time. So I am not sure if it makes sense to accept this pull request and complete the integration across the entire framework?

hishamco:
> The original hypothesis for this enhancement was that "introducing a generic repository will save time instead of creating I{Type}Repository for every entity"

Yes it is but not for every entity, perhaps that some entities are still required `I {Type}Repository` but more than 80% of the code is required is shifted to `Repository<TType>` this of sure has a lot of benefits to us, coz you don't need to write the code for the basic CRUD operations again and again, also if your repository has a custom methods you will see yourself calling the core methods which are already implemented with minor changes instead of duplicating the code

Hope my answer satisfying the benefits of introducing this pattern

sbwalker:
Thank you for your response - I think you are correct that the genetic repository offers some additional benefits other than time savings. I am not sure if you saw the latest blog about Oqtane philosophy https://www.oqtane.org/Resources/Blog/PostId/538/oqtane-philosophy but the Practical Engineering is an area where I am trying to focus to ensure the framework is intuitive and not over engineered. So it’s a matter of being deliberate about changes and really understanding the impacts and benefits. I also believe that consistency is really important - meaning that pattens and techniques should be implemented across the entire framework, and that coding styles and guidelines should also be followed.

hishamco:
Great post Shaun, devs are really need "Less Ceremony", on other hand we need to understand the impacts & benefits as you mentioned earlier

In term of consistency I'm agree with you, but we as framework architect we should think about many things too maintainability, simplicity and extensibility

Don't forget DRY principle, and let dev do what he is supposed to do and leave the rest to the framework

I'm very transparent to hear you opinion as the framework owner about the pros and cons of the old and current repository pattern and let us talk loudly while Oqtane is OSS

My regards

hishamco:
It would be nice if we have a gitter room for discussion with the community

sbwalker:
Excellent suggestion about creating a gitter room for collaboration - I will do that as soon as I get home tonight ( I am traveling back to the U.S. from Europe today ). I will also put together my thoughts on pros and cons - although at this point I think there are more pros and very few cons with the generic repository pattern you submitted.

hishamco:
BTW I hope to be a core contributor to this project, I still remember the first DNN  versions (VB.NET) I was hope to be a contributor at that time, but I didn't get a chance there, but I got a chance here to do some contributions

hishamco:
BTW I hope to be a core contributor to this project, I still remember the first DNN  versions (VB.NET) I was hope to be a contributor at that time, but I didn't get a chance there, but I got a chance here to do some contributions

PoisnFang:
I would prefer to set it up how PiranhaCMS does their Db with an [Api class](https://github.com/PiranhaCMS/piranha.core/blob/master/core/Piranha/Api.cs). You can add all the services into the Api class and just call the Api, you can also set it up to work with the Generic IRepository<TType> and remove all the repository and services files. That way you do not need to specify all the repositories individually and it is dynamic.

hishamco:
Each CMS has it's own thoughts, in Piranha CMS seems the `Api` class aggregate all the service it's similar to `UmbracoContext` where all the services are there, IMHO injecting the required services may better while the AspNetCore offers all the underlying APIs

Again each approach has prons and cons

sbwalker:
Closing for consideration after V1

hishamco:
@sbwalker, can we revise this one more time

sbwalker:
@hishamco This is a good example of an approach which is based on individual preference - some people like generic repositories and others do not. I do not believe refactoring the core framework to use a generic repository provides enough product benefit to warrant the change. However, developers can certainly use a generic repository pattern in their own custom modules.

hishamco:
At least we could provide the abstraction APIs for it instead of implementing it from scratch

PR #196, opened 2020-02-07
Soft Delete Module

### LINKED ISSUE #105: Need ability to soft delete core entities

fileman:
I don't know if it's optimal but it works and fixes #105 , I wait for any change requests

### DIFF
--- a/Oqtane.Client/Modules/Admin/ModuleDefinitions/Index.razor
+++ b/Oqtane.Client/Modules/Admin/ModuleDefinitions/Index.razor
@@ -46,7 +46,7 @@ else
     List<ModuleDefinition> moduledefinitions;
     List<Package> packages;
 
-    protected override async Task OnInitializedAsync()
+    protected override async Task OnParametersSetAsync()
     {
         try
         {
--- a/Oqtane.Client/Modules/Admin/ModuleDefinitions/Index.razor
+++ b/Oqtane.Client/Modules/Admin/ModuleDefinitions/Index.razor
@@ -51,6 +51,7 @@ else
         try
         {
             moduledefinitions = await ModuleDefinitionService.GetModuleDefinitionsAsync(PageState.Site.SiteId);
+            moduledefinitions = moduledefinitions.Where(item => !item.IsDeleted).ToList();
             packages = await PackageService.GetPackagesAsync("module");
         }
         catch (Exception ex)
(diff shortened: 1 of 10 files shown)


sbwalker:
The purpose of the recycle bin is to allow an Administrator of a site to restore a module "instance" to a specific page in a site - not to delete all instances of a module across all pages/sites.

fileman:
> The purpose of the recycle bin is to allow an Administrator of a site to restore a module "instance" to a specific page in a site - not to delete all instances of a module across all pages/sites.

If I have not misunderstood, the instance of the module on a specific page is "page module", which is already implemented in the recycle bin in previous commit.

This allow host role users to:
- soft delete -> no more listed in page available modules;
- restore -> show in page available modules;
- hard delete -> if module was not actively used in any page of any site erase its .dll files

sbwalker:
In regards to #105 PageModules and Modules were on the list of items which should support restoration from the recycle bin - but not ModuleDefinitions. In Oqtane, ModuleDefinition records are created automatically during startup based on the assemblies/types it finds in the /bin. ModuleDefinitions only exist so that there is a way to define the permissions of who is allowed to use specific modules(apps) in a site. If you want to restrict users from utilizing a specific module(app) you would adjust the permissions in the Edit Module Definition screen. You can also uninstall a module by deleting it through the application or manually deleting the DLL from the /bin. Note that there is no foreign key relationship between ModuleDefinition and Module - this is on purpose. As a result there is a use case for softdeleting/restoring ModuleDefinitions.

A Module represents an "instance" of a module{app} that is related to a site. The Module can be exposed on one or more pages in a site - which is where PageModule comes in. From a recycle bin perspective a user can delete a module on a page. The assoxiated PageModule record will be flagged as IsDeleted which means it will appear in the Recycle Bin and can be restored or hard deleted. If the PageModule is hard deleted, the system needs to check if its ModuleId is associated to any other PageModules in the site ( softdeleted or active). If it is not, then the system should hard delete the Module record as well ( as otherwise it would be orphaned ).

mikecasas:
Are the modules in the core framework different concerning permissions? I noticed in the Weather Forecast module there are no permissions set, but yet it is listed in the Select module dropdown.

My goal would be to have ten modules in the bin folder but only site 1 can use modules 1-5 and site 2 could only use modules 6-10. I am assuming this is possible?

sbwalker:
@mikecasas yes this is the purpose of defining permissions for each module definition. You go into the Admin Dashboard and Module Management area, you can see the list of installed Modules and you can click Edit next to any of them to adjust the permissions.  Currently the framework assumes that a module is available to be utilized on all sites unless specific permissions have been configured otherwise.

sbwalker:
Closing this PR based on my earlier comments about ModuleDefinitions

PR #209, opened 2020-02-15
Unable to deleted the root site

### LINKED ISSUE #202: Root site can be deleted
Result is white screen, unrecoverable.

hishamco:
Fixes #202

### DIFF
--- a/Oqtane.Server/Repository/SiteRepository.cs
+++ b/Oqtane.Server/Repository/SiteRepository.cs
@@ -142,10 +142,13 @@ namespace Oqtane.Repository
         }
 
         public void DeleteSite(int siteId)
-        {
-            Site site = db.Site.Find(siteId);
-            db.Site.Remove(site);
-            db.SaveChanges();
+        {       
+            if (db.Site.Count() > 1)
+            {
+                var site = db.Site.Find(siteId);
+                db.Site.Remove(site);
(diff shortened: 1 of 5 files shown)


sbwalker:
Instead of adding a Disabled property to the ActionDialog, can you add the validation logic to the Index.razor for the Site.

hishamco:
I was planned to do that first, but I see value to have such property in `ActionDialog` coz there 're some case to have the `ActionLink` disabled

sbwalker:
Good point! Can you also add this feature to ActionLink as well? Also there is a merge conflict in ActionDialog.

hishamco:
Ok, I will handle this ASAP

sbwalker:
Thank for addressing the merge conflict

Oqtane.Server/Repository/SiteRepository.cs
```diff
@@ -142,10 +142,13 @@ public Site GetSite(int siteId)
         }
 
         public void DeleteSite(int siteId)
-        {
-            Site site = db.Site.Find(siteId);
-            db.Site.Remove(site);
-            db.SaveChanges();
+        {       
+            if (db.Site.Count() > 1)
```
chlupac: Site.Count is not enough. Root(first) site have to be protected. Strange things happen when you do it.

PR #212, opened 2020-02-16
RecycleBin Page Review

fileman:
Remove LoadEntities method in favor of List's Remove method

### DIFF
--- a/Oqtane.Client/Modules/Admin/RecycleBin/Index.razor
+++ b/Oqtane.Client/Modules/Admin/RecycleBin/Index.razor
@@ -36,7 +36,7 @@
         @if (pageModules.Count == 0)
         {
             <br/>
-            <p>No Deleted Modules</p>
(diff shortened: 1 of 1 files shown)


sbwalker:
The Pages collection which is part of PageState is essentially a client-side cache which is used so that repeated calls do not need to be made to the server for the same data. That being said, if you can imagine that the system is servicing multiple users simultaneously, it is possible that another user may add/update/delete pages while the current user is interacting with the system. If this occurs then you would want the current user to get access to the most up to date list of pages - it it should make a service call to the server to retrieve them. So in general, PageState should be treated as read-only data - its content should not be manipulated directly, as doing this only affects the current and does not take into consideration other changes by other users of the system.

sbwalker:
I am inclined to close this PR based on my earlier comments that utilizing the List Remove method for manipulating PageState properties is not a good approach

PR #214, opened 2020-02-18
authorization changes

sbwalker:


sbwalker:
This is a large pull request as there were extensive changes needed to complete the authorization review.

sbwalker:
The most significant changes in this PR were removing some items from the PageState object so that internal system information is not leaked to the client for requests from anonymous users. In addition, all Get() Controller methods now have authorization implemented so that they only return items for which the current user has access rights. Where applicable, processing logic was migrated from the client to the server, as the server is the ultimate place where security needs to be enforced.

PR #244, opened 2020-02-28
Creation of site fix

chlupac:
-add ~ to site folder name
-check if alias endpoint is alive with ping request

### DIFF
--- a/Oqtane.Client/Modules/Admin/Sites/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Sites/Add.razor
@@ -6,8 +6,9 @@
 @inject ISiteService SiteService
 @inject IThemeService  ThemeService
 @inject IUserService UserService
+@inject IPingService PingService
 
-@if (tenants == null)
+@if (_tenants == null)
 {
     <p><em>Loading...</em></p>
 }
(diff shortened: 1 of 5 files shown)


sbwalker:
I do not want to users to have to add a "~" to their folder names

chlupac:
it's handled automatically

sbwalker:
It is not 100% clear what this pull request was trying to resolve. A few observations:

- the ping service you included to test if an alias is a valid url has security issues. The controller is available for anyone to call and pass any url they desire. This could be abused ( ie. a malicious user could initiate a DOS attack on a url and use your site as a proxy ).

- although it seems like a good idea to validate that an alias is valid, there are a number of use cases where this functionality is not desired. For example, when a user is setting up a site they may want to include a number of aliases which have not yet been configured through DNS yet - as they may plan to do this step later. The system should not restrict the user from entering them. In addition, you should be able to use Oqtane on your local machine without an Internet connection - so requiring an alias to be validated may pose an unnecessary restriction.

sbwalker:
#249 has a number of improvements for creating/editing/deleting sites which should help address some of the items you were trying to resolve in this PR. For example. if you enter an alias value incorrectly, you are now able to access the Edit Site dialog in Site Management to fix the alias value.

chlupac:
Ok, i see than my code is obsolete now, i will close this PR and i will test new update. 

1. This Ping controller is available as other publicly accessible controllers and do not pass any url. I do not see any security problem. Controller returns only "Pong" to any request. How you can use it as proxy?

2. But you are switching to that alias after creation of site and this action  fail, if it not available.

PR #252, opened 2020-03-04
Fix naming conventions for private fields

hishamco:
Fixes a part of #245

PR #259, opened 2020-03-06
ModuleBase fix

chlupac:
Fix of #252 and some cosmetics

### DIFF
--- a/Oqtane.Client/Modules/ModuleBase.cs
+++ b/Oqtane.Client/Modules/ModuleBase.cs
@@ -10,11 +10,12 @@ namespace Oqtane.Modules
 {
     public class ModuleBase : ComponentBase, IModuleControl
     {
-        public Logger logger { get; set; }
+        private Logger _logger;
+
+        protected Logger logger => _logger ?? (_logger = new Logger(this));
 
         public ModuleBase()
         {
-            logger = new Logger(this);
         }
 
         [Inject]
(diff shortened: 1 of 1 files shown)


Oqtane.Client/Modules/ModuleBase.cs
```diff
@@ -10,11 +10,12 @@ namespace Oqtane.Modules
 {
     public class ModuleBase : ComponentBase, IModuleControl
     {
-        public Logger logger { get; set; }
+        private Logger _logger;
+
+        protected Logger logger => _logger ?? (_logger = new Logger(this));
```
chlupac: Lazy loading of logger, it was instantiated on every module loaded from database

Oqtane.Client/Modules/ModuleBase.cs
```diff
@@ -10,11 +10,12 @@ namespace Oqtane.Modules
...
 
         public ModuleBase()
         {
-            logger = new Logger(this);
```
chlupac: Ah, I forgot delete this

sbwalker:
Can you explain why it is necessary to do lazy loading? Since this code was working fine before, I would assume that a functional change would not be required to resolve an issue caused by renaming variables?

chlupac:
It is not necessary, it's optimization and cosmetic. You instantiated all modules during boot, not only used ones. Constructor was called for all modules. I can delete it if you want, it's my coding style how I solve this patterns.

sbwalker:
Sorry... it is difficult to do code reviews when fixes and enhancements are included in the same pull request. I was expecting that the PR was to fix a bug introduced by the renaming, so the other code change did not make sense to me. However it is a good optimization so I will merge it.

PR #261, opened 2020-03-06
Add unit test project

### LINKED ISSUE #230: Add unit tests project
@sbwalker it would be nice to add unit tests to ensure that any added feature is tested before merge, could you please handle #227 then I can start on this

hishamco:
Fixes #230

Just adding an empty unit test project with the required dependencies, hopefully if we can add unit tests for the entire platform as well as for the bugs that we fixed if it's possible


hishamco:
@sbwalker could you please merge this, then @chlupac can start to add some tests for the fixes that he made in #230

PR #262, opened 2020-03-06
Html text

chlupac:
HtmlText fix + little cosmetics


ADefWebserver:
Good job :)

sbwalker:
There is a bug in this PR which can be reproduced if you add a new Html/Text module to a page ( which initially has no content ) and you click Edit. It displays the both rich text editor and raw text editor side by side.

sbwalker:
Note that the ability to switch between a rich text editor and raw text editor should be self contained in the RichTextditor.razor component - then any module can take advantage of it simply by including the component.

PR #265, opened 2020-03-08
Get files and folder by folder path

chlupac:


sbwalker:
I do not understand what this PR is trying to accomplish. Can you please explain the purpose of these new API methods which retrieve a file or folder based on a path - there does not appear to be a consumer of them. I also do not want Microsoft.Extensions.Logging introduced into the project as we do not need 2 logging frameworks.

chlupac:
I have module which displays a random picture on site header from folder. During development I realize than will be handy to have this little extension. It can be realized with bunch of calls to current api (get list of folders, select one, get list of files). This is reason why I made this little extension.

logger - your logger api is nice for tracing user activity, but not for development and diagnostic. And it's tied up to ModuleBase class, you are not able to log events at controller level (or I do not know how)
All .core events (for example EF database calls) are already in this sink, you can see them immediately in console and are very helpful in debugging time.  And only thing which you have to do is inject ILogger service to class. Or use Console.Writeline().

sbwalker:
One of the benefits of open source is that you get participation from people with very different backgrounds, experience levels, and use cases which they would like to implement. So I definitely appreciate what you bring to the table.

The Folder/File API is designed to be an abstraction. Currently the underlying implementation only supports storage on the server itself however in the future the goal would be to support other types of storage ( ie. cloud services, etc... ). In order to embrace the storage of files and support future storage locations, it is important for Oqtane modules to use the API abstraction. So it is recommended that that modules utilize immutable concepts like FolderId/FileId rather than names/paths which can change. For example, you would not want to store a file name or folder path within your module, as this could easily be broken in the future. So this is the reason why I was asking about the purpose of these new API methods - as although they make conceptual sense, they may also send the wrong message to developers and encourage them to use file names/paths which could have negative consequences. Perhaps you could explain your use case in more detail - ie. why you dealing with file names/paths rather than Ids.

From a logging perspective, you can take a look at any of the many Controllers in the Server project to see an example of how to use the logger. There are logging methods in the Client project in ModuleBase for convenience when building Blazor components, and there is a LogManager class in the Server project for logging in Controllers/Repositories. Both classes log events to the same storage location and they add custom enrichments which are specific to Oqtane and critical in identifying concepts like tenant, etc...

chlupac:
Ok, thank you, it makes sense. It's another perspective to the problem and I will respect it.

sbwalker:
I was actually interested in hearing about your use case for the new API methods just to ensure I am not missing a need?

chlupac:
It's simple banner which randomly select image from folder. User can upload images to this folder
'''c#
using System;
using System.Linq;
using System.Threading.Tasks;
using Microsoft.AspNetCore.Components;
using Oqtane.Modules;
using Oqtane.Services;
using Oqtane.Shared;

namespace Oqtane.Theme.Mondi.Controls
{
    public partial class PageBanner : ModuleBase
    {
        [Inject] private IFileService FileService { get; set; }
        private string _image;
        private readonly Random _rnd = new Random();

        [Parameter]
        public string Image
        {
            get => _image;
            set => _image = value;
        }

        [Parameter] public RenderFragment ChildContent { get; set; }

        public async Task<string> GetImage(string path)
        {
            var files = await FileService.GetFilesAsync(PageState.Site.SiteId, path);
            if (files != null && files.Count > 0)
            {
                return ContentUrl(files[_rnd.Next(files.Count - 1)].FileId);
            }
            return "Themes/Oqtane.Theme.Mondi/images/Subheader_1920x400.png";
        }

        {
            Image = await GetImage("Banners\\");
        }
    }
}

'''

sbwalker:
When you say that users know where to upload... are you saying that they would upload them outside of the Oqtane framework such as via FTP or ...? 

Regardless, in this case I would suggest that you could create a Settings component for your module ( see the Blog module for an example of a Settings component ). The Settings component would have a FileManager component on it where the user can select a Folder. The selected FolderId would be stored in Settings - which could then be used in the PageBanner component, rather than hardcoding a folder path.

That being said, I can understand there may be cases where it may be more convenient to reference a folder by its path. Obviously, the developer would need to be aware that this could be affected by an Administrator going into the File Management area and renaming the folder - which would break their module. However, if you have full control over your site and will mitigate this risk yourself, then it is a valid use case.

I will reopen this PR - but I would prefer if you updated the logger references to the Oqtane logger before I merge it.

chlupac:
ok. Upload using Oqtane. Using Ftp will not help because i'm using virtual oqtane folder, not physical

PR #267, opened 2020-03-10
completed client state invalidation in multi-user environment

sbwalker:


sbwalker:
Note that the original POC implementation utilized a querystring parameter ( "&reload=" ) to instruct the client that it needed to refresh items in PageState. This solution only works for a single user, as it does not take into consideration any changes that other users may be performing concurrently.

Issue #275, opened 2020-03-11
host user can forbid access to module for himself, no way back 


chlupac:
If I press Save now, Files module is lost and I can't reverse this action.
![image](https://user-images.githubusercontent.com/7838468/76431273-a366b580-63b1-11ea-878b-6fb7be3afa51.png)

sbwalker:
the fix here should be to prevent a host user from removing the Administrators role from the permissions ( if the permissions grid component does not handle this already ). A user should never be able to to remove the Administrators role from any permission.

sbwalker:
fixed in #279

Issue #274, opened 2020-03-11
After site creation no modules in control panel and management


chlupac:
No default "Utilize" permission is set on modules in new site during site creation. 
 

![image](https://user-images.githubusercontent.com/7838468/76429618-934dd680-63af-11ea-8c8e-7f9c5fd4e1c2.png)

chlupac:
Why not to give host implicit Utilize permissions?

sbwalker:
As a general rule in Oqtane I do not want to use any implicit/assumed security permissions - all security rules need to be explicitly defined. This is a security best practice and reduces the risk of unexpected vulnerabilities.

sbwalker:
This issue is fixed in #278 - the code was not synchronizing module definition permissions correctly for new sites

PR #276, opened 2020-03-11
Add implicit Utilize permissions for host users

chlupac:
GetModuleDefinitions optimalisation - one call for getuser() instead of one call per module
Cosmetics and naming fixes

If Host User is king no permissions rules is necessary

solution for #275 workaround for #274


sbwalker:
As a general rule in Oqtane I do not want to use any implicit/assumed security permissions - all security rules need to be explicitly defined. This is a security best practice and reduces the risk of unexpected vulnerabilities.

Issue #280, opened 2020-03-12
Move magic strings to constants


chlupac:
Specially Permission strings to extra class

sbwalker:
There is a Utilities.Constants class where I have tried to consolidate these items. Most of the permission strings are in the SiteRepository constructor and I plan to introduce a more robust and extensible solution in this area so that new sites can be created based on custom templates.

chlupac:
This i what I am thinking about, Site templates. 
Maybe simplest solution is serialize existing site to Json.

sbwalker:
One of the philosophies behind Oqtane is encapsulation, so as much as possible I am trying to avoid external resources ( DNN relied on a ton of external resources which was very difficult to maintain/support as the platform evolved over time ). What I am envisioning here is a simple interface ( ie. ISiteTemplate ) which has a Name property and a generic method for CreateSite(Site site). Within this method a developer would have full control over the site creation process. They could created any entities they desire ( ie. users, roles, pages, modules, folders, files, jobs, etc.... ). During installation and site creation the UI would provide an option for the user to choose from a list of templates which are available ( based on classes implementing ISiteTemplate ).

chlupac:
yes, its second option which im thinking about, complex site solution including themes modules and site template

chlupac:
I'm working on similar thing now. Thinking about some helper methods and extensions to build site easily and refactoring friendly, without need of magic strings.
Question is how to describe module typeof(what?).Name
```c#
public static void AddModule<T>(this Page page, ....) where T is IModule
{
}
```


```c#
    public interface ISiteTemplate
    {
        void InstallSite();
        string Name { get; }
    }
```

sbwalker:
I have a partial implementation for the site template feature

chlupac:
great
Look at Permission Builder class, it can help with permission strings on site builder

sbwalker:
[...]

PR #282, opened 2020-03-13
Define Permissions as enum, Permissions builder

chlupac:
Step 1 in case #280 
Shall I continue?


Oqtane.Shared/Enums/Permissions.cs
```diff
@@ -0,0 +1,11 @@
+﻿﻿namespace Oqtane.Shared
+{
+    public enum Permissions
+    {
+        Browse,
```
chlupac: Browse is not actual permission, forgot to delete it

Issue #297, opened 2020-03-20
Master.sql should be embeded as resource. 


chlupac:
Master.sql is not part of publish and after site publishing is not copied to correct place.

dbUp library is capable to do whole database installation process including database creation and variable substitution. I have tested:

```c#
                
            EnsureDatabase.For.SqlDatabase(connectionString);
            
            var dbUpgradeConfig = DeployChanges
                .To
                .SqlDatabase(connectionString)
                .WithVariable("ConnectionString",connectionString)
                .WithVariable("Alias", silentInstall.Alias)
                .WithScriptsEmbeddedInAssembly(Assembly.GetExecutingAssembly(),s => s.StartsWith("Master")) 
                .WithScriptsEmbeddedInAssembly(Assembly.GetExecutingAssembly(),s => !s.StartsWith("Master"));
            var dbUpgrade = dbUpgradeConfig.Build();
            if (dbUpgrade.IsUpgradeRequired())
            {
                var result = dbUpgrade.PerformUpgrade();
                if (!result.Successful)
                {
                    Console.WriteLine(result.Error.Message);
                }
                else
                {
                        UpdateAppSettings(connectionString);
                }
            }
```

sbwalker:
If you have tested this thoroughly and it works then I think this is a good improvement,.

chlupac:
Solved in #306

PR #306, opened 2020-03-25
Database Manager

chlupac:
DONE:
+ master.sql as resource
+ dbUp sql script variables implemented
+ improved database handling and creation code
+ simpified database creation
+ almost all Database and Tenant creation moved to DatabaseManager.cs (rest code marked with TODO)
+ Improved IsInstalled checking
+ Possibility of unattended install of master. Can be performed by SilentInstall.json config file

TODO:
- tenant based on database schema almost implemented, does not work at site creation point (dbcontext default schema not set correctly), requires further examination of problem
- have to find place where to call SilentInstall, probably on IsInstalled check or in Startup procedure.


sbwalker:
This is very large refactoring which impacts a lot of use cases. Have you tested this PR with standard SQL Server as well as LocalDB? Have you tested creating an additional Tenant and then adding a Site to the new Tenant? Have you tested installing a module package ( ie, the Blog module )? I am asking because it is very difficult to assess from the code changes if any of the test scenarios might have been adversely affected.

chlupac:
I know this is huge core refactoring. I'm kindly asking community to fork my chlupac:DatabaseManager branch and test it before we merge it.
I test almost all cases you mentioned. I have to test some error cases (db permissions not set, db inaccessible, ...),  Let this PR sit here some time, I will try to solve schema issue and do more tests.

sbwalker:
OK, I will hold off on merging until you are done. A few comments based my initial code review of the current PR:

1. Embedding master.sql as a resource ( #297 ) appears to be good
2. You can easily get the current Alias and Tenant from your server code by using TenantResolver. You should not need to pass alias name.
3. Unattended install is a good feature however based on the Oqtane philosophy I am trying to minimize the number of external resources. I would prefer if the unattended install options were included in the appsettings.config rather than creating another config file. 
4. The ability to separate your tenants by Schema was never fully implemented. It was included in the Tenant entity as a placeholder for the future. I see that you are attempting an implementation however I would recommend not spending too much time on it as I don't think it will be a use case which many people will utilize. I am actual wondering if it is a worthwhile feature at all. DNN supported this concept but it meant that all database scripts for modules had to be coded in a specific way to support multiple schemas - which was a learning curve and resulted in many issues over the years. MY personal feeling is that we should remove the concept of tenant per schema because it actually creates more problems than benefits.

chlupac:
ad 3. I can move code appsettings. 
ad. 4.
I agree. It's complicated. Personally I like have things as simple as possible. I can remove schema code at all and in this case I do not have any problems and testing will be easy.

sbwalker:
4. I am glad we are on the same page. I 100% agree that simple is better. After looking at the changes you made to the install scripts to add $schema$ it brought back memories of all of the issues which developers had in DNN creating scripts for their modules. It may be possible that if Oqtane used EF Core Migrations someday ( and not scripts ) then this would not be a problem - but that is a big effort as well and probably best reserved for a future release. Do you want to do the work to remove DBSchema from the Tenant entity ( and all related changes ) as part of this PR?

chlupac:
Yes, I will do it.

sbwalker:
I should also mention that it was quite complicated to get the multi-tenant installation process to work. In particular if you look at the InstallationController it has its own InstallationContext defined at the end of the file because at the time this class is executed, most of the entities have not yet been created so the standard Context objects do not work properly as EF Core can't bind the entities to the physical database tables. In addition, you will notice that when creating a new Tenant through the UI it sets an IsInitialized property to false, which it then uses when you try to create a new Site for that Tenant in the UI ( ie. it asks for the host username/password so that it can complete the installation ). Getting all of this working took a LOT of effort, so I am not sure if refactoring the logic will be very straightforward or produce better results.

chlupac:
My goal is

1. Consolidate database creation and initialization to one specialized class, preferably independent on web UI based on Single-responsibility and Separation of concerns principle. This separation helps in future move to EF database migrations.

2. Allow silent install config for easy deployment to production and for help to repeatable installs during development cycle and testing

3. Elimination of try-catch checks, which are annoying during debugging

sbwalker:
In my initial implementation I tried to move the logic from the Installer.Razor to create the initial site and user to the InstallationController - however I ended up giving up on that approach because of the complicated DBContext issues I ran into. So I left them separated - and that approach seemed to work better for the other scenario of creating new Tenants/Sites as well. 

However I do understand the benefit of having an automated installation option. During an automated installation the connection string and host account details would be explicitly provided in the config file so its possible the DBContext issues would not be a problem - ie. simply introducing another method in the InstallationController for the automated scenario may work just fine.

chlupac:
Hi, I have rethink my implementation and go further. Migration should be used to move database from one state to another by ADDING new script to resource. current implementation does not make it correctly, especially on master database.
I take some inspiration here:
https://andrewlock.net/series/running-async-tasks-on-app-startup-in-asp-net-core/

Migration script (DbUp) should be called at:

Application startup
1. at application startup to be sure than database is in correct version, also for all tenant databases and module scripts. If DbUp find here is if all is correct IsInstalled is set to true. It is not necessary to check database state at every page reload. 
2. If IsInstalled is false try to run silent install and if successful set IsInstalled to true

Master Install
3. App.razor check IsInstalled and  if it's false raises Installation form
4. Install controller collect data at submit and start migration.

new tenant install
1. at Tenant management database information was collected and send to controller
2. controller tries to create tenant - call Database manager for migration in tenant database
3. Database Manager installs tenant db and calls module scripts

sbwalker:
I tried the same approach when I was building the POC for Oqtane. I even tried the technique in the  blog post from Andrew Lock as I ran into the same problem with calling async tasks on startup and I still had a lot of difficulty. I spoke to Dan Roth about it at Microsoft and he basically told me it was not supported in .NET Core. However last week he told me that in .NET 5 it will be supported - which will be helpful for a number of scenarios.

Overall my recommendation is to not do a deep refactoring in this area right now. The current implementation works and covers a lot of scenarios... I am sure you are the familiar with the saying "if it ain't broke, don't fix it". What we can do right now is:

- embed master.sql as an embedded resource
- remove DBSchema from the Tenants table and related code
- implement a new method in InstallationController which handles silent installation.

chlupac:
Database Manager
done:
+ master.sql as resource
+ implemented incremental database changes also for Master
+ dbUp sql script variables implemented
+ improved database handling and creation code
+ simpified database creation
+ almost all Database and Tenant creation moved to DatabaseManager.cs (rest code marked with TODO)
+ Unattended install of master can be performed by settings in appsettings.json
+ Improved IsInstalled checking
+ Removed DBSchema field from Tenant
+ Default database and site creation moved to Program.Main


@sbwalker from your comment ... - and I still still had a lot of difficulty and had to abandon the idea. ....

challenge accepted. 
problem solved. (with help of a very small hack in tenant resolver)
solution tested.

ready to hear complains ...

sbwalker:
one scenario which is not easy to test at the moment but is an important requirement and is functional in the current code is in-place upgrade of the framework. There is an Upgrade Service option in the Admin Dashboard which allows you to either upload an Oqtane Nuget package or download a Nuget package from Nuget.org. The system places this package in the wwwroot/Framework folder and then extracts the DLLs and runs any upgrade database scripts. It uses the Oqtane.Upgrade projct to accomplish this as it needs to be performed a different process. It all works - but do you think the changes which are part of this PR will affect it?

sbwalker:
The core database upgrade scripts were handled within the startup code - they would compare versions and automatically perform any core database upgrades required. I will have to look closer if there was any changes in this area with your PR. But if you have any thoughts based on your knowledge of the changes you made, please let me know.

chlupac:
If you restart application after install new version, all updates on all tenants will be handled by startup migration. No need to call upgrade scripts in upgrade project.

PR #308, opened 2020-03-26
Added logic to check for a valid email address.

aubrey-b:
There is no check on the email address when a user registers, other than the address is not empty. 

A bad email address caused an error when the email job was turned on.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Register/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Register/Index.razor
@@ -1,4 +1,5 @@
-﻿@namespace Oqtane.Modules.Admin.Register
+﻿@using System.Net.Mail;
+@namespace Oqtane.Modules.Admin.Register
 @inherits ModuleBase
 @inject NavigationManager NavigationManager
 @inject IUserService UserService
@@ -42,13 +43,16 @@
     string _confirm = "";
     string _email = "";
     string _displayName = "";
+    bool _isEmailValid = false;
 
     private async Task Register()
     {
         try
         {
             _message = "";
-            if (_username != "" && _password != "" && _confirm != "" && _email != "")
+            ValidateEmailAddress();
+
+            if (_username != "" && _password != "" && _confirm != "" && _isEmailValid)
             {
                 if (_password == _confirm)
                 {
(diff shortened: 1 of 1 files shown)


chlupac:
Whether it is working or not using try catch to validate input is not recommended practice. Regex is definitely better way to go.

sbwalker:
This is a good idea, as it is important to validate email addresses. The challenge in client-side Blazor is that the code runs in the browser and does not have access to resources like System.Mail. In addition, Oqtane is a client/server application so you would this validation to occur when people are interacting with the client as well as when they call the API directly. So the logic you suggested would need to be implemented on the server - most likely in the UserController.

sbwalker:
@chlupac  is correct that Regex would be preferable for client-side validation of email addresses - which would eliminate the need for System.Mail ( which would not work in client-side Blazor )

aubrey-b:
OK, I will look into that...I forgot about the client side implications of using the mail namespace. I like to avoid regex, but I will find the pattern.

Would it be better to use a model in the shared project and use it on both the client and server?

aubrey-b:
@chlupac Why is the try-catch pattern not a good one here?

sbwalker:
I would focus on @chlupac 's suggestion to simply add client-side validation for now using RegEx.

chlupac:
@aubrey-b : It's not good practice to use try catch as logical operator. 
more in discussion here:
https://softwareengineering.stackexchange.com/questions/107723/arguments-for-or-against-using-try-catch-as-logical-operators

ctrl-c ctrl-v from this discussion:

Debug and maintenance work is very difficult when control flow is performed using exceptions.

Inherently, exceptions are designed to be a mechanism for altering the normal control flow of your program - of performing unusual activities, causing unusual side-effects, as a way to get out of a particularly tight bind that can't be handled with less complicated means. Exceptions are exceptional. This means that, depending on which particular environment you're working in, use of exception for regular control flow can cause:

    Inefficiency The additional hoops that the environment has to jump through to safely perform the relatively difficult context changes required for exceptions require instrumentation and resources.

    Debug difficulties Sometimes useful (when trying to debug a program) information gets thrown out the window when exceptions occur. You can lose track of program state or history that's relevant for understanding run-time behavior

    Maintenance problems Execution flow is hard to follow through exception jumps. Beyond that, an exception may get thrown from inside black-box type code, which may not behave in easy to understand ways when throwing an exception.

    Poor design decisions Programs built in this manner encourage a frame of mind which, in most cases, doesn't easily map to solving problems elegantly. The complexity of the final program discourages the programmer from understanding fully its execution, and encourages taking decisions which lead to short-term improvements with high long-term costs.

aubrey-b:
@chlupac  The example on Microsoft docs for an email regex uses a try catch.

https://docs.microsoft.com/en-us/dotnet/standard/base-types/how-to-verify-that-strings-are-in-valid-email-format

chlupac:
Yes, for checking errors, not for validation. In normal circumstances exception is not thrown. You thrown exception on every bad user input.

Issue #256, opened 2020-03-05
Validators should be used on forms


chlupac:
We have to implement validation. Not necessary for MVP but think about it.
simple one  DataAnnotationsValidator 
or something like 
https://github.com/ryanelian/FluentValidation.Blazor
I'm using it and it's easy.

```c#
            RuleFor(x=>x.Email).NotEmpty().EmailAddress().MaximumLength(255)
                .Must(BeUniqueEmail)
               .WithMessage("Adresa je již zaregistrovaná");

            RuleFor(q => q.UserName)
                .NotEmpty()
                .Must(BeUnique)
                .WithMessage("Uživatel s tímto jménem již existuje");

            RuleFor(q => q.Password)
                .NotEmpty();
                RuleFor(q => q.Compare)
                    .Equal(q => q.Password);
            RuleFor(q => q.Phone)
                .NotEmpty()
                .MinimumLength(9)
                .Matches("^[+]?[()/0-9. -]{9,}$")
                .WithMessage("Telefonní číslo není správně vyplněno");

```

mikecasas:
I agree with the validation, but is there not a library that picks up the "rules" from my model using data annotations? I don't want to define the requirements on my my model and then again for the validation with rules. Maybe it's me, but I like seeing the validation logic on my model.

```csharp
public class Customer  
{   
    [StringLength(maximumLength: 50  , MinimumLength = 10,  ErrorMessage = "The property {0} should have {1} maximum characters and {2} minimum characters")]  
    public string Name { get; set; }  

} 
```

chlupac:
DataAnnotationValidator does this job. But it'd not as flexible as FluentValidator. And on DataAnotations is some minor problems with localization. 

'''c#
using System.ComponentModel.DataAnnotations;

public class ExampleModel
{
    [Required]
    [StringLength(10, ErrorMessage = "Name is too long.")]
    public string Name { get; set; }
}
'''
'''c#
<EditForm Model="@_exampleModel" OnValidSubmit="HandleValidSubmit">
    <DataAnnotationsValidator />
    <ValidationSummary />

    <InputText id="name" @bind-Value="_exampleModel.Name" />

    <button type="submit">Submit</button>
</EditForm>

@code {
    private ExampleModel _exampleModel = new ExampleModel();

    private void HandleValidSubmit()
    {
        Console.WriteLine("OnValidSubmit");
    }
}
'''
[...]

PR #309, opened 2020-03-27
Check for a valid email.

aubrey-b:
Use a regex to check if an email is valid.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Register/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Register/Index.razor
@@ -97,7 +97,7 @@
     {
         NavigationManager.NavigateTo(NavigateUrl(""));
     }
-}
+
     private void IsValidEmail()
     {
         if (_email != ""){
--- a/Oqtane.Client/Modules/Admin/Register/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Register/Index.razor
@@ -1,4 +1,5 @@
 ﻿@namespace Oqtane.Modules.Admin.Register
+@using System.Text.RegularExpressions;
 @inherits ModuleBase
 @inject NavigationManager NavigationManager
 @inject IUserService UserService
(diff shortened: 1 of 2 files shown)


Oqtane.Client/Modules/Admin/Register/Index.razor
```diff
@@ -94,4 +97,16 @@
     {
         NavigationManager.NavigateTo(NavigateUrl(""));
     }
+}
+    private void IsValidEmail()
```
chlupac: make bool function and call it directly
private bool IsEmailValid(string email)  {}

chlupac:
It's much better. 
Make this validation function static bool and put it to Oqtane.Shared.Utilities class. Because this is not only place where we need to email validation.

aubrey-b:
I added the logic to the Oqtane.Shared.Utilities class.

Oqtane.Shared/Shared/Utilities.cs
```diff
@@ -217,5 +219,18 @@ private static string RemapInternationalCharToAscii(char c)
                 return "";
             }
         }
+
+        public static bool IsValidEmail(string email)
+        {
+            if (email != "")
+            {
```
chlupac:
```c#
if(string.IsNullOrEmpty(email)) return false; 
```
should be here i think.

aubrey-b: I kept the logic the same from the razor file because that's what you said to do. I will make the change. I would have done it like this too, but I was following your instructions and others.

aubrey-b: @chlupac yes I did test it.

hishamco:
IMHO we can simplify this by using `EmailAddress` data annotation attribute

chlupac:
@hishamco This was discussed in #256. Requires deeper refactoring of forms.

hishamco:
If is that I think we should TRUST a 3rd-party library specifically if we want to deal with emails, but it's fine to start with this

chlupac:
I think standard Data Anotation validator should be enough for basic forms. And it's in line with Shaun's policy about 3rd party libraries.

Issue #245, opened 2020-02-28
Naming conventions 


chlupac:
This is what I miss in Oqtane code. 
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines

https://github.com/ktaranov/naming-convention/blob/master/C%23%20Coding%20Standards%20and%20Naming%20Conventions.md

with help of refactoring tools is easy to put code to shape

sbwalker:
This is a good suggestion as I am a firm believer in coding standards and consistency. I am concerned that a global refactoring could destabilize the application but the right time to do this is probably before the MVP release. I am noticing that some PR's contain massive amounts of code changes which are not functional but mainly cosmetic - which makes it really difficult to do a code review and identify potential issues. Hopefully implementing some common coding standards will improve this situation.

hishamco:
I already have some doubts about the code conventions, I may push some minor PRs to resolve the issue, instead of a HUGE one ;)

sbwalker:
The main code conventions I would like Oqtane to follow are the guidelines specified in https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines ( as mentioned by @chlupac ). So this is primarily related to naming conventions ( upper/lower case, local variables using "_", etc... ). This is not related to changes to the architecture. Smaller PRs would be better than massive ones so that they are easier to review.

hishamco:
Ok, I will start on this one soon ..

chlupac:
Strange Logger (again) problem after #252 
modulebase is null when logger is trying log successful login. 
![image](https://user-images.githubusercontent.com/7838468/76095143-1643eb00-5fc4-11ea-8c50-67246f2a0a53.png)

sbwalker:
Sounds like a regression issue related to #252 - someone will need to track it down. This is why I was hesitant to introduce such a widespread change to the codebase.

chlupac:
I'm trying to debug it now

chlupac:
I have it. Preparing PR now

sbwalker:
Due to the inconsistent upper/lower case naming of local variables and parameters in the original code, it is likely that there may be a few regression issues. That being said, it should be possible to look at the history and code differences to help identify the problems.

hishamco:
@sbwalker don't worry I will have a look too, that's why I suggested CI/CD and unit tests

chlupac:
[...]

PR #314, opened 2020-03-30
Refactoring

hishamco:


sbwalker:
"Refactoring" does not provide any clue of what this PR is trying to accomplish. If there was any specific goals behind the PR it is good to provide the context so that it helps with the review of the code changes. Is this related to #245 or #280?

hishamco:
Both, it would be nice if we can merge this after setup CI/CD to ensure it doesn't break anthing

hishamco:
I will defer the conflict resolve until we provide CI/CD coz every time I fix it new PR is merged ;)

hishamco:
Again, moving to code behind model will simplify the things and makes the views clear

sbwalker:
@hishamco it sounds like you are saying that you have not tested these changes with a clean install ( ie. remove connectionstring and install new database ). If so, then I will definitely wait to merge them.

hishamco:
> you have not tested these changes with a clean install ( ie. remove connectionstring and install new database )

Yep, I just build the entire solution to make sure that I didn't make something wrong, because if you look closely, what I did is very simple refactoring

sbwalker:
It is a fairly simple refactoring - but this PR affects almost every UI in the application - which means every UI interaction should be regression tested. CI/CD would verify that the application installs successfully but at this point it will not be capable of any regression testing. It might be a good idea to attempt a clean local install and manually test each UI affected?

hishamco:
> It might be a good idea to attempt a clean local install and manually test each UI affected?

I'm trying a fresh installation everything till now is working as expected ..

sbwalker:
Are you able to deal with the latest conflicts and then I can merge this? Its a large PR and I don't want it to sit on the shelf too long or else it will become more difficult to merge

hishamco:
Sure

hishamco:
@sbwalker please squash & merge

sbwalker:
This PR introduced 56 compilation errors when it was merged. It appears that a global search and replace was done for string.Empty which caused issues in every case where an escape sequence was being used to build a string ( ie. \" ). 

ie.            _openLabel += " for=\string.Empty + For + "\string.Empty;

I am curious how you were able to perform a clean install when the code will not even compile?

hishamco:
Please let me check ...

sbwalker:
I have fixed all of the compilation errors and will check in shortly. It appears they may have occurred when you were dealing with the merge conflicts.

hishamco:
May be .. If you want me to send a PR for regression bug no problem I can do it now ... again I request you to setup CI/CD to avoid such issues

Thanks

PR #319, opened 2020-03-31
Prevent deleting master tenant

### LINKED ISSUE #317: Why there's ability to delete master database

hishamco:
Fixes #317

### DIFF
--- a/Oqtane.Client/Modules/Admin/Tenants/Edit.razor
+++ b/Oqtane.Client/Modules/Admin/Tenants/Edit.razor
@@ -9,7 +9,14 @@
             <label  class="control-label">Name: </label>
         </td>
         <td>
-            <input class="form-control" @bind="@name" />
+            @if (name == Constants.MasterTenant)
+            {
+                <input class="form-control" @bind="@name" readonly />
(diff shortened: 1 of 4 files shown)


sbwalker:
The fix handles the UI but it also needs to be handled in the API as you do not want the master tenant to be deleted via an API call.

hishamco:
I forgot that ;) I will update it soon ..

hishamco:
Done!!

sbwalker:
after further review, this change does not really accomplish the goal. The "master" tenant is very important because it is the registry which is used to access any of the other tenants. If you consider a scenario where someone installs Oqtane and then adds an additional tenant "Tenant2", you code you added would not prevent them from deleting the "master" tenant ( as the number of tenants > 1 ). It would seem that we need a bit flag on a Tenant for IsMaster which can be checked, or we need to use a constant for the name of the master tenant which can be used for verification.

hishamco:
> this change does not really accomplish the goal

You 're right, let me have a look again ...

hishamco:
@sbwalker could please check it again

Issue #318, opened 2020-03-31
Tenant is null


mikecasas:
Getting an exception when starting the app that the tenant is null after adding an additinal site.

Steps to reproduce.
1) Pull latest commit c9b
2) Clean install
3) Start app - everything is fine
4) Add new site - everything is fine
5) stop app - I am using VS 2019, IIS express
6) start app - see error message below

![tenant](https://user-images.githubusercontent.com/11261164/78018448-e3b7b480-731b-11ea-8c62-d8714b1b5dae.PNG)

chlupac:
Thanks for report, fix is on the way.

chlupac:
done.

mikecasas:
Closing as fix is tracked in #320.

PR #320, opened 2020-03-31
Startup migration bug

chlupac:
#318 solution


sbwalker:
One question I have is is related to the name of the new config section - calling it "Oqtane" seems redundant as the entire config file belongs to the Oqtane framework? Wouldn't it make more sense to name the new section "Installation" - as this section is simply replicating the fields which are usually entered via the UI. There is already an installation config setting called "SiteTemplate" which could be moved to this section ( see SiteRepository.cs for _config.GetSection("SiteTemplate") ).

  "Installation": {
    "HostUser": "",
    "HostPassword": "",
    "HostEmail": "",
    "SiteTemplate": ""
  }

Note that there is another section used by the framework currently - but it is not included in the current default config file ( see LogManager.cs ):

  "Logging": {
    "LogLevel": {
      "Default": "Information"
    }
  }

sbwalker:
we should remove the modulecreator modification from this PR as it is not related

chlupac:
Ok, I can change appsetting.json this way, it's more descriptive, good idea.

chlupac:
@sbwalker Done. you can merge it

Oqtane.Server/Extensions/PermissionExtension.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Text;
+using System.Text.Json;
+using Oqtane.Models;
+
+namespace Oqtane.Extensions
+{
+    public static class PermissionExtension
+    {
+        public static string EncodePermissions(this IEnumerable<Permission> permissionList)
+        {
```
chlupac, line 12: Copy from _PermissionRepository Function is static utility and for call is not necessary repository instance.
Refactored as extension.

chlupac: Refactoring of all calls will be done in next step.

sbwalker:
One comment I have is related to this code:

                   //TODO this is wrong. Decide default password / email or throw exception ??
                    Password = GetInstallationConfig(SettingKeys.HostPasswordKey, "oQtane123"),
                    Email = GetInstallationConfig(SettingKeys.HostEmailKey, "nobody@cortonso.com"),
                    DisplayName = GetInstallationConfig(SettingKeys.HostUserKey, Constants.HostUser),

I would think that that Password, Email, and DisplayName should be set to "" by default. And only if a person wants to use the "silent install" option would they specify values for these settings. Anytime an application includes a default password - especially for a highly privileged account - it has the potential to cause problems as it gives an easy attack vector for malicious users.

chlupac:
Agree, I want to discuss this. I will check all required properties and when is not defined, system will fallback to GUI setup

chlupac:
Default password removed, fallback to IDE

aubrey-b:
I get the tenant is null error even after this PR.

sbwalker:
What does your appsettings.json look like?

chlupac:
@aubrey-b How to reproduce bug ?

chlupac:
@aubrey-b solved in #329

aubrey-b:
@sbwalker appsettings:

```
{
  "ConnectionStrings": {
    "DefaultConnection": "Data Source=(LocalDb)\\MSSQLLocalDB;AttachDbFilename=C:\\Users\\aubrey\\source\\repos\\oqtane.framework\\Oqtane.Server\\Data\\Oqtane-202004021802.mdf;Initial Catalog=Oqtane-202004021802;Integrated Security=SSPI;"
  },
  "Installation": {
    "DefaultAlias": "",
    "HostUser": "",
    "HostPassword": "",
    "HostEmail": "",
    "SiteTemplate": "",
    "DefaultTheme": "",
    "DefaultLayout": "",
    "DefaultContainer": ""
  }
}

```

aubrey-b:
@chlupac Sorry to respond so late, but I just add a second site and then when I restart the application I get the error message similar to #318

PR #324, opened 2020-04-02
Use string Interpolation for constructing Urls

### LINKED ISSUE #300: Use C# String Interpolation for constructing Urls
This is an improvement suggestion. For a cleaner code I suggest how the urls are constructed in the Client project services can improved by using interpolation or string.format.

hishamco:
Fixes #300

### DIFF
--- a/Oqtane.Client/Services/AliasService.cs
+++ b/Oqtane.Client/Services/AliasService.cs
@@ -36,7 +36,7 @@ namespace Oqtane.Services
 
         public async Task<Alias> GetAliasAsync(int aliasId)
         {
-            return await _http.GetJsonAsync<Alias>(Apiurl + "/" + aliasId.ToString());
+            return await _http.GetJsonAsync<Alias>($"{Apiurl}/{aliasId.ToString()}");
         }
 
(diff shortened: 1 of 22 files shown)


sbwalker:
Looks good!

mikecasas:
@sbwalker Going forward as a pattern should we use string interpolation? Also if we see string concatination do you want PRs to fix?

sbwalker:
I think string interpolation is a good practice going forward. As far as refactoring existing cases of string concatenation I think we need to limit the modification of code/functionality which is already working or else we will spend all of our time dealing with regression and never get to a 1.0 release.

hishamco:
But there are few cases may it need improvement to make code cleaner and readable .. please check my PRs

sbwalker:
Indeed there is always room for improvement - however “clean” and “readable” are subjective concepts based on individual developer preference. What I was saying is that we need to focus on stability at this stage rather than refactoring code that is already working.

hishamco:
Sound good nothing but, we can do improvements if it's possible, because we can't break APIs if we shipped `1.0.0` until `2.0.0`. IMHO we can plan about `RC` before `1.0.0`, so we will have enough time for test-ability and stability before GA

PR #325, opened 2020-04-02
Add ModuleDefinition to IModule

### LINKED ISSUE #322: Avoid magic strings in IModule
As I have seen that `IModule` contains a `Properties` which is a set of module properties that the Oqtane will use, but the problem is they are magic string, so:

- User may miss some of the properties

- Casing may be another issue

- Spelling is another issue

to avoid such things I suggest to introduce a `ModuleAttribute` or `ModuleMaifestAttribute` that contains a set of all the required attributes that we have, so:

- It's strongly typed

- Will fix all of the above issues

- It will be useful in modules discovery if the modules are come from other assemblies

I can work for a prototype for this is sound looks good to you

/cc @sbwalker

hishamco:
This fixes #322

I marked `Properties` as `Obsolete` for backward compatibility, we can then remove it before ship `1.0.0`

### DIFF
--- a/Oqtane.Client/Modules/Admin/ModuleCreator/Module.cs
+++ b/Oqtane.Client/Modules/Admin/ModuleCreator/Module.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using Oqtane.Models;
+using System.Collections.Generic;
 
 namespace Oqtane.Modules.Admin.ModuleCreator
 {
@@ -18,5 +19,13 @@ namespace Oqtane.Modules.Admin.ModuleCreator
                 return properties;
             }
         }
+
+        public ModuleDefinition ModuleDefinition => new ModuleDefinition
+        {
+            Name = "Module Creator",
+            Description = "Enables software developers to quickly create modules by automating many of the initial module creation tasks",
+            Version = "1.0.0",
+            Categories = "Developer"
+        };
     }
 }
--- a/Oqtane.Client/Modules/Counter/Module.cs
+++ b/Oqtane.Client/Modules/Counter/Module.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using Oqtane.Models;
+using System.Collections.Generic;
 
 namespace Oqtane.Modules.Counter
 {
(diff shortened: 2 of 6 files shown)


sbwalker:
The main thing I was concerned about in changing IModule to a strongly typed object is backward compatibility. Has this been tested to ensure the approach is backward compatible? ie. have you used the changes in this PR to create a compiled module which implements the new form of IModule, and then added a property to ModuleDefinition and tested if the module continues to function properly?

hishamco:
I will try add a unit tests if its possible, but again I'm asking to setup CI/CD ;)

sbwalker:
This is not something that can be tested via CI/CD or unit tests. In order to test this it requires the creation of an external module ( you could use the Blog Module - https://github.com/oqtane/oqtane.module.blogs ) and modifications to the core framework which will ultimately not even be checked in ( ie. adding a property to the ModuleDefinition class to test impact on backward compatibility ). CI/CD is not a replacement for hands-on development testing.

hishamco:
> This is not something that can be tested via CI/CD or unit tests

Functional testing, anyhow I will have a try to make sure it works as expected, but the tests will prove to everyone of you that changes are working for both

sbwalker:
This pull request comoiled but threw run-time errors... there were numerous issues that needed to be resolved in ModuleDefinitionRepository in order to get this concept working. I worked through them in #343. Once the concept was implemented I upgraded the Blog module to use the new IModule interface. I installed the new Blog module into the framework. Then I modified the ModuleDefinition class by adding an additional property. I built/ran the framework and the Blog module continued to function properly. This proves that the concept maintains backward compatibility. So I committed the changes as part of #343 and also committed the changes to the Blog module. I am closing this PR.

hishamco:
> Once the concept was implemented I upgraded the Blog module to use the new IModule interface.

I'm not sure how it's backward compatible, while your removed the `Properties` property?!!

sbwalker:
I said the new IModule interface is backward compatible.

hishamco:
AFAIK backward compatibility don't allow to remove the `public` members

sbwalker:
I think we are having a gap in communication. Oqtane had an IModule interface definition which was based on a dictionary. You pointed out the weakness of that approach and I explained it was implemented that way because backward compatibility will be important for external modules. You said that a strongly typed object should be backward compatible as well. So I tested the concept... and you were correct. So I took the opportunity prior to 1.0 to change the IModule interface so that it is strongly typed. There is no need to preserver backward compatibility on the old IModule interface because we have not released 1.0 yet. I hope this is clear.

PR #326, opened 2020-04-02
Add ModuleDefinition to ITheme

hishamco:


sbwalker:
The ModuleDefinition class should not be used for Themes - they are different types of entities. If we are able to prove that moving to a strongly types object preserves backward compatibility, then this PR will need to be modified to introduce a ThemeDefinition class with only the properties which are valid for Themes. But its not worth making those changes until the concept is proven.

sbwalker:
Modified in #344 to use existing Theme model and new interface approach. Closing PR.

hishamco:
I don't much difference with your PR, furthermore your pr is not backward compatible as you asked before, that why my PR looks different little bit because support both

Thanks

PR #327, opened 2020-04-02
Database manager

chlupac:


chlupac:
Strange, I was sure this was already commited

sbwalker:
@chlupac I am testing the installation of the external Blog module. The way this used to work is that you could copy the nupkg package to the /wwwroot/Modules folder and the next time you started Oqtane, it would install the module which included running the database installation scripts. This is now broken based on the changes in this PR.

chlupac:
Ok i will repair it

Issue #328, opened 2020-04-02
Site creation bug 


chlupac:
The instance of entity type 'Tenant' cannot be tracked because another instance with the same key value for {'TenantId'} is already being tracked. When attaching existing entities, ensure that only one entity instance with a given key value is attached. Consider using 'DbContextOptionsBuilder.EnableSensitiveDataLogging' to see the conflicting key values.

   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.ThrowIdentityConflict(InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.Add(TKey key, InternalEntityEntry entry, Boolean updateDuplicate)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.Add(TKey key, InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.Add(InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.StartTracking(InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState oldState, EntityState newState, Boolean acceptChanges, Boolean modifyProperties)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState entityState, Boolean acceptChanges, Boolean modifyProperties, Nullable`1 forceStateWhenUnknownKey)
   at Microsoft.EntityFrameworkCore.ChangeTracking.EntityEntry.set_State(EntityState value)
   at Oqtane.Repository.TenantRepository.UpdateTenant(Tenant tenant) in W:\Oqtane\oqtane.framework\Oqtane.Server\Repository\TenantRepository.cs:line 48
   at Oqtane.Controllers.TenantController.Put(Int32 id, Tenant tenant) in W:\Oqtane\oqtane.framework\Oqtane.Server\Controllers\TenantController.cs:line 60
   at lambda_method(Closure , Object , Object[] )
   at Microsoft.Extensions.Internal.ObjectMethodExecutor.Execute(Object target, Object[] parameters)
[...]

PR #329, opened 2020-04-02
Tenant repository bug

chlupac:
Database Manager bug

### DIFF
--- a/Oqtane.Server/Infrastructure/DatabaseManager.cs
+++ b/Oqtane.Server/Infrastructure/DatabaseManager.cs
@@ -63,6 +63,7 @@ namespace Oqtane.Infrastructure
             if (_isInstalled && !IsDefaultSiteInstalled(defaultConnectionString))
             {
                 BuildDefaultSite(password,email);
+                
             }
         }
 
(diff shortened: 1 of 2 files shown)


mikecasas:
@chlupac I pulled your branch that you want merged and when I tried adding a new site I got the error below in SiteRouter.razor.

I am not sure if this is related to .NET Core 3.2 preview or something else.


chul1

chlupac:
Did you fill site form correctly ? I think it's not related to my changes.

mikecasas:
You are right. When I create the new site using the Oqtane theme it works fine, but when I use the Blazor theme it crashes because the drop down for Default Layout is empty/

chlupac:
put this bug to issues

mikecasas:
I just did. See #330

sbwalker:
@chlupac is the bug/fix related to the refactoring you did for DatabaseManager or is it related to the change @hishamco made for disallowing changing the master tenant name?

chlupac:
Here is two bugs One is mine, I forgot set flag on tenant, and one is @hishamco with disallowing master tenant name

sbwalker:
Ok, I am not sure about the fix for UpdateTenant - having to use .AsNoTracking() seems like a workaround for a different problem.

hishamco:
> Here is two bugs One is mine, I forgot set flag on tenant, and one is @hishamco with disallowing master tenant name

Could you please let me know what's the bug, so I can fix it now

chlupac:
No, problem is than you cannot have two different entities with same key tracked.
AsNotracked loads detached entity. It's correct.

chlupac:
@hishamco it's fixed

sbwalker:
The key is the TenantId?

chlupac:
yes, look at #328 
when you assign entry state you actually attach entity

sbwalker:
Ok, I understand. The call in GetTenant() uses _db.Tenant.Find(tenantId) which attaches the object to the context with the TenantId... and then later in the UpdateTenant method we are trying to update a different object with the same TenantId

PR #332, opened 2020-04-03
No more magic strings in module definition

chlupac:


sbwalker:
This is a good approach for the core framework. However if someone was creating a SiteTemplate as a custom assembly, the modules it includes may not yet be installed in the environment so they types would not yet be loaded. In this case they would still need to specify a typename as a string value - with the expectation that the modules would be installed later. This would still work as ModuleDefinitionName is a string and there is no dependency or tight coupling with the core framework.

PR #334, opened 2020-04-03
Fix regression bugs

### LINKED ISSUE #330: DefaultLayoutType value is "string.Empty"
I know there was some refactoring done, so now in Client/Modules/Admin/Sites/Add.razor when the layout type is null the DefaultLayoutType contains the value "string.Empty". Was this the desired result?

add-site

When adding a new site with the Blazor theme and no default layout I get this error.

chul1

hishamco:
Fixes #330 which is address `string.Empty` regression bugs that was occurs from #314

### DIFF
--- a/Oqtane.Client/Modules/Admin/Pages/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Pages/Add.razor
@@ -118,7 +118,7 @@
             </td>
             <td>
                 <select class="form-control" @onchange="(e => ThemeChanged(e))">
-                    <option value=string.Empty>&lt;Select Theme&gt;</option>
+                    <option value="">&lt;Select Theme&gt;</option>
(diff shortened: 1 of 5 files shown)


hishamco:
I think `@string.Empty` will render `value=` while the intend to render `value=""`, `@string.Empty` is useful in C# code not in HTML unless there's a proper places

Am I right?

chlupac:
you have to use "@string.Empty"

sbwalker:
I explained it in my post on #330 - adding the @ symbol works fine, I tested it.

<option value=@string.Empty>&lt;Select Layout&gt;</option>

sbwalker:
Actually I am not sure how much value there is in using @string.Empty over "" in razor components. This is markup which is going to the browser and executing code to produce "" seems like overkill.

PR #336, opened 2020-04-04
Simplify NavigateUrl using UriBuilder

hishamco:
FYI I found a little tiny bug with the old implementation when we have url like `https://contoso?pageId=4` it should be `https://contoso/?pageId=4` to meet the standard

### DIFF
--- a/Oqtane.Shared/Oqtane.Shared.csproj
+++ b/Oqtane.Shared/Oqtane.Shared.csproj
@@ -18,6 +18,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" Version="2.2.0" />
     <PackageReference Include="System.ComponentModel.Annotations" Version="4.7.0" />
     <PackageReference Include="System.Text.Json" Version="4.7.1" />
   </ItemGroup>
--- a/Oqtane.Shared/Oqtane.Shared.csproj
+++ b/Oqtane.Shared/Oqtane.Shared.csproj
@@ -18,7 +18,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" Version="2.2.0" />
     <PackageReference Include="System.ComponentModel.Annotations" Version="4.7.0" />
     <PackageReference Include="System.Text.Json" Version="4.7.1" />
   </ItemGroup>
--- a/Oqtane.Shared/Shared/Utilities.cs
+++ b/Oqtane.Shared/Shared/Utilities.cs
@@ -14,30 +14,17 @@ namespace Oqtane.Shared
             var assemblyName = assemblyFullName.Substring(0,  assemblyFullName.IndexOf(",", StringComparison.Ordinal));
             return $"{type.Namespace}, {assemblyName}";
         }
+
         public static string NavigateUrl(string alias, string path, string parameters)
         {
-            string url = "";
-            if (alias != "")
-            {
-                url += alias + "/";
-            }
-            if (path != "" && path != "/")
-            {
-                url += path + "/";
-            }
-            if (url.EndsWith("/"))
-            {
-                url = url.Substring(0, url.Length - 1);
-            }
-            if (!string.IsNullOrEmpty(parameters))
-            {
-                url += "?" + parameters;
-            }
-            if (!url.StartsWith("/"))
-            {
-                url = "/" + url;
-            }
-            return url;
+            var uriBuilder = alias == string.Empty
+                ? new UriBuilder()
+                : new UriBuilder(alias);
+
+            uriBuilder.Path = path;
+            uriBuilder.Query = parameters;
+
+            return uriBuilder.Uri.AbsoluteUri;
         }
 
         public static string EditUrl(string alias, string path, int moduleid, string action, string parameters)
--- a/Oqtane.Shared/Shared/Utilities.cs
+++ b/Oqtane.Shared/Shared/Utilities.cs
@@ -2,6 +2,8 @@
 using System.Globalization;
 using System.Text;
 using System.Text.RegularExpressions;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Extensions;
 
 namespace Oqtane.Shared
 {
@@ -17,14 +19,29 @@ namespace Oqtane.Shared
 
         public static string NavigateUrl(string alias, string path, string parameters)
         {
-            var uriBuilder = alias == string.Empty
-                ? new UriBuilder()
-                : new UriBuilder(alias);
+            if (!alias.StartsWith("/"))
+            {
+                alias = $"/{alias}";
+            }
 
-            uriBuilder.Path = path;
-            uriBuilder.Query = parameters;
+            if (!path.StartsWith("/"))
+            {
+                path = $"/{path}";
+            }
 
-            return uriBuilder.Uri.AbsoluteUri;
+            var pathPaseValue = alias == string.Empty
+                ? default
+                : new PathString(alias);
+
+            var pathValue = path == string.Empty
+                ? default
+                : new PathString(path);
+
+            var queryStringValue = parameters == string.Empty
+                ? default
+                : new QueryString($"?{parameters}");
+
+            return UriHelper.BuildRelative(pathPaseValue, pathValue, queryStringValue);
         }
 
         public static string EditUrl(string alias, string path, int moduleid, string action, string parameters)
--- a/Oqtane.Shared/Shared/Utilities.cs
+++ b/Oqtane.Shared/Shared/Utilities.cs
@@ -2,8 +2,6 @@
 using System.Globalization;
 using System.Text;
 using System.Text.RegularExpressions;
-using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Extensions;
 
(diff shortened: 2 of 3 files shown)


hishamco:
@sbwalker just a question do we really need an `alias` to be empty?

sbwalker:
Yes we need the alias parameter. In the case where you create a microsite ie. “www.domain.com/microsite” the alias parameter represents the alias path ie. “microsite”. This value must be part of the url to maintain the site context.

sbwalker:
I have my doubts that this change would support microsites or has been tested with that scenario.

hishamco:
@sbwalker have a look to my first unit test I used `contoso` as alias, but in case to support `www.domain.com/contoso` for instance I write a unit test for it, but I expect the `www.domain.com/contoso` as alias, because if the alias is `contoso` there's a missing piece here, where did the domain is attached to the generated URL?

sbwalker:
The domain should never be returned by NavigateUrl() as that would break the client-side routing behavior. The example I provided which included the domain was simply to explain the concept of microsites. A microsite is purely virtual - and it must be included as part of the relative path. Have you actually created a microsite in your local installation ( ie a site with localhost:44358//site2 ) and tested that the NavigateUrl() links are created properly?

hishamco:
> A microsite is purely virtual - and it must be included as part of the relative path. Have you actually created a microsite in your local installation ( ie a site with localhost:44358//site2 ) and tested that the NavigateUrl() links are created properly?

I will create one soon, but the unit tests are fine to guarantee that nothing is breaking ..

sbwalker:
So I took some time this afternoon to test the code change in this pull request to verify the concerns I expressed above. 

First I tested the default site of with an alias of ""localhost:44357". The alias path for this site is "" and I want to create a link to navigate to a page I created with a path of "page1". The call to this NavigateUrl method would be the following NavigateUrl("","page1",""). The result should be a relative path of "/page1". However when I use the code in this pull request I get a result of "http://localhost/page1". 

I the created a Site with an alias of "localhost:44357/site2". The alias path for this microsite is "site2" and I want to create a link to navigate to a page I created with a path of "page1". The call to this NavigateUrl method would be the following NavigateUrl("site2","page1",""). The result should be a relative path of "/site2/page1". However when I use the code in this pull request I get a result of "http://site2/page1". 

There are multiple problems here... 

1) it is returning an absolute Url rather than a relative url - which disrupts the entire SPA client-side routing model as it will force a full page refresh for every request ( note that the absolute url generated does not include the port either which makes it invalid but this behavior can be ignored as this method should not generate absolute urls anyways )
2) it treats a microsite alias path as a domain rather than as a path. 

Unit tests are only useful if they compare the results against valid values. In this case both the code changes and unit tests are incorrect.

hishamco:
Let me change the code to return the relative Url, coz if you look closely to the `UriBuilder` I returned the absolute URL, that's why I asked you where the domain is attached

if the concern is to support SPA, the change that I made need to be modified to get the relative URL instead of abosulte URL

hishamco:
> Unit tests are only useful if they compare the results against valid values. In this case both the code changes and unit tests are incorrect.

The unit tests are correct according to my changes, let me do a little modification and I will test them after and before the change without need to run the app

sbwalker:
A Blazor app is a SPA app. Regardless of whether you are running it server-side or client-side it is using a client-side routing model.

The unit test expected results are all wrong

sbwalker:
The alias parameter value can only contain 2 possible values:
1. If its a standard site (ie. "localhost:33457") it will be ""
2. If its a microsite ( ie. "localhost:33457/site1" ) it will be the first segment of the url path ( ie. "site1"
The alias parameter value will never contain a protocol, domain , port, etc...
The NavigateUrl method will always produce relative urls - it will never produce absolute urls

If you look at the current tests you will see they are all wrong.

sbwalker:
This is a perfect example of what I commented on earlier. The current NavigateUrl() method works fine - there are no issues with it. This PR is attempting to refactor the method, and if merged, it would have created significant regression issues. And in the process it has consumed a considerable amount of time and effort evaluating the PR, testing it, explaining the expected results and regression issues, etc... It would be best of we could avoid these types of situations.

hishamco:
> A Blazor app is a SPA app.

Right, because the changed to get the absolute URL, just few minutes, I will update the PR ..

hishamco:
> This PR is attempting to refactor the method, and if merged, it would have created significant regression issues

You right, but I just asked you first to check if the absolute URL valid or just I did it wrong, that's why I revert the absolute path change

hishamco:
@sbwalker could please check the current unit tests with old implementation and new implementation, they will be same except 3 unit tests because you are missed a `/` in the old implementation

sbwalker:
So the hypothesis of this PR is that a url should have a "/" preceding the "?" identifying the querystring ( ie. path/?id=3 ). This does not seem to be accurate based on this Stack Overflow post https://stackoverflow.com/questions/1617058/ok-to-skip-slash-before-query-string/42193734#42193734 but it is important to note that every major browser supports urls without the "/" ( ie. path?id=3 ). So this PR does not fix an issue - the NavigateUrl method was already working properly. 

During my testing of the code in the original PR to verify the results, I noticed that the UriBuilder does not produce urls in the format you are suggesting either. I modified the logic slightly so that it would produce the correct urls based on the parameters:

        string alias = "site1";
        string path = "page1";
        string parameters = "id=3";
        var uriBuilder =  new UriBuilder();
        uriBuilder.Path = (!string.IsNullOrEmpty(alias)) ? $"{alias}/{path}" : $"{path}";
        uriBuilder.Query = parameters;
        string url = uriBuilder.Uri.PathAndQuery;

This results in a url value of "/site1/page1?id=1". Notice that there is no "/" preceding the "?". So UriBuilder does not include a "/" before the "?". This would lead me to believe that the original hypothesis is not correct.

Looking at the updated unit tests, I still see 2 which do not include the "/" before the "?":

        [InlineData("contoso", "login", "returnUrl=/admin", "/contoso/login?returnUrl=/admin")]
        [InlineData("", "login", "returnUrl=/admin", "/login?returnUrl=/admin")]

If you are saying that these 2 unit tests are passing your new logic, then I guess the new logic is not correct?

hishamco:
@sbwalker I'm not sure if you notice the latest changes, I realize that the `UriBuilder` will not work nicely with our implementation, that's why I changed

Thanks

sbwalker:
Why does UriBuilder() not work? The test code I posted above produces valid Urls for all scenarios:

    var uriBuilder =  new UriBuilder();
    uriBuilder.Path = (!string.IsNullOrEmpty(alias)) ? $"{alias}/{path}" : $"{path}";
    uriBuilder.Query = parameters;
    string url = uriBuilder.Uri.PathAndQuery;

hishamco:
The problem with the `UriBuilder` is generate an absolute URL not a relative one .. I tested the unit tests with old and new implementation, result looks the same

sbwalker:
The code with UriBuilder that I posted above produces a relative url.

hishamco:
I see, you are using `PathAndQuery` instead, looks good, so if this looks good I can revert my last commit and change the PR to what you wrote above if the commits not looks good to you

sbwalker:
Yes please change the PR as it is not necessary to add another package reference to the Oqtane.Shared project for this purpose. You will probably also want to adjust your tests as UriBuilder does not include a "/" before the "?" for a querystring.

hishamco:
I will update this soon ..

hishamco:
> it is not necessary to add another package reference to the Oqtane.Shared project for this purpose.

The reference is required for the `UriBuilder`

hishamco:
I don't think there's an issue if we add `/` before querystring while each path should be ended with `/` but I'm not sure how the above coded works with you without the `/` before the querystring

I will push the update PR now ..

hishamco:
@sbwalker seems you didn't look to my earlier comment ;) two of the unit tests are failing because I removed the `/` before querystring as you suggested

Again I think there's no problem if we have a `/` before a querystring, if is that I need to fix two unit tests to make them pass

sbwalker:
in the case where alias = "" and path = "" and there are parameters the url generated should be in the form "/?id=1". My earlier comment about a problem with your unit tests was related to the scenarios where it was expecting results such as "/contoso/?pageId=4&moduleId=10" which are not correct. So yes, you can fix these 2 unit tests:

        [InlineData("", "", "pageId=4", "/?pageId=4")]
        [InlineData("", "", "pageId=4&moduleId=10", "/?pageId=4&moduleId=10")]

hishamco:
> My earlier comment about a problem with your unit tests was related to the scenarios where it was expecting results such as "/contoso/?pageId=4&moduleId=10" which are not correct

These are failed unit tests

```csharp
[InlineData("contoso", "", "pageId=4", "/contoso?pageId=4")]
[InlineData("contoso", "", "pageId=4&moduleId=10", "/contoso?pageId=4&moduleId=10")]
```

a `/` is required to make them pass

sbwalker:
Please change the unit tests, a "/" is not required before a "?" unless the path = ""

hishamco:
Seems there's confusion here ;) the problem is not in unit tests according to your latest suggested modification

This require a modification in the code, hope my explanation is clear

Thanks

sbwalker:
Yes, there must be some confusion here. The current code works fine and produces the correct results. If the results do not match the unit tests, then the unit test values are wrong.

hishamco:
Right  .. adding `/` to above mentioned unit tests will make the unit tests happy and make me happy too ;)

I will send a PR now

PR #337, opened 2020-04-04
Introduce INavigator

hishamco:


sbwalker:
Sorry but this PR does not solve any problems - it is a refactoring to use default properties rather than overloads which is a personal developer preference and does not improve the framework. In fact, this refactoring breaks existing functionality where NavigateUrl() is expected to use the PageState.Page.Path property value.

Oqtane.Client/Modules/ModuleBase.cs
```diff
@@ -45,19 +46,13 @@ public string ModulePath()
...
-            return NavigateUrl(PageState.Page.Path);
-        }
-
-        public string NavigateUrl(string path)
-        {
-            return NavigateUrl(path, "");
-        }
+            if (path == string.Empty && parameters == string.Empty)
+            {
+                path = PageState.Page.Path;
```
hishamco, line 53: The use of `PageState.Page.Path` is here, if you notice earlier it was used on the `NavigateUrl()` overload which is the similar things here .. nothing breaking while it uses an optional params

sbwalker:
There is a bug in the code in this PR - the path should default to PageState.Page.Path if the path parameter is “” - not if both path and parameters are equal to “”. 

I am not going to merge this PR as the current framework code works fine. Also I prefer multiple overloads to default parameter values - I feel they are more intuitive for developers. Closing this PR

hishamco:
No problem, my intend was to reduce the amount of allocation that's introduced by tons of string concatenations

hishamco:
I revert the above comment, because I confused with another PR ;)

PR #339, opened 2020-04-04
Icon constants + some refactoring

chlupac:


chlupac:
From css, + some block edit + some replaces

PR #342, opened 2020-04-05
Webassembly problem solved

chlupac:


sbwalker:
Part of the problem is solved but the site still does not load on WebAssembly. I will re-open the issue.

PR #345, opened 2020-04-05
made IModule implementations consistent

sbwalker:


sbwalker:
Something is wrong with this PR - can’t be merged. Will need to close and sort out the problem.

Issue #211, opened 2020-02-16
Question - custom _Host.cshtml


chlupac:
Is it possible to have custom _Host.cshtml per site ?

sbwalker:
No it is not possible as Blazor expects a single entry point to the application ( and when running on Wasm the entry point is not _host.cshtml - it is index,html in the wwwroot of the Client project ). I think the more important question is why you are asking this question - and I am guessing it is because you would like to reference additional Javascript and CSS files. The optimal solution for this problem is to use JavaScript Interop and dynamically load JavaScript and CSS files based on the needs of the application. This keeps the static dependencies to a minimum.

chlupac:
Yes, I will do it this way. Thank you.

chlupac:
Actually I solved this problem with new project with custom _Host.cshtml page which loads Oqtane.  
I also need to register extra services and modify little bit startup.
I do not need to use wasm in this project. 
The advantage of this solution is that I can have references to my modules and themes here, which simplifies their development and debugging. 
My main is here:
```c#
using FluentValidation;
using Mfl.Data;
using Mfl.Models;
using Microsoft.AspNetCore.Hosting;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;
using Microsoft.Extensions.Logging;
using Oqtane;
using Syncfusion.EJ2.Blazor;
using Syncfusion.Licensing;

namespace Mfl.Boot
{
    public class Program
    {
        public static void Main(string[] args)
        {
            SyncfusionLicenseProvider.RegisterLicense("");
            CreateHostBuilder(args).Build().Run();
        }

        public static IHostBuilder CreateHostBuilder(string[] args)
        {
            return Host.CreateDefaultBuilder(args)
                .ConfigureLogging(logging =>
                {
                    logging.ClearProviders();
                    logging.AddConsole();
                })
                .ConfigureWebHostDefaults(webBuilder =>
                {
                    webBuilder.UseStartup<MyStartup>();
                    webBuilder.ConfigureServices(x =>
                    {
                        x.AddSyncfusionBlazor();
                        x.AddDbContext<MyDbContext>(
                            ServiceLifetime.Transient);
                        x.AddTransient<IValidator<RegistrationModel>, RegistrationModelValidator>();
                    });
                });
        }
    }

    public class MyStartup : Startup
    {
        public MyStartup(IWebHostEnvironment env) : base(env)
        {
        }
    }
}
```

[...]

PR #346, opened 2020-04-06
Fix of tenant migration and module installation

chlupac:


sbwalker:
I think I am missing something here. When creating an external module, the SQL scripts are embedded in the module assembly. In addition, modules maintain their own Version information which is independent from the Oqtane core framework. So when an external module is installed it needs to get the version from the module and run all of the scripts <= the version. On module upgrade it needs to get the module version and run all scripts from the currently installed version up to the new version. The installation/upgrade for modules was working this way previously but in the DatabaseManager refactoring I dont see how this behavior was preserved?

chlupac:
For all tenants I call
'''c#
     foreach (var assembly in assemblies) ModuleMigration(assembly, connectionString);
'''

sbwalker:
thank you for explaining... I was trying to understand the code changes in the PR and how they related to module installation. I guess the answer is that they relate indirectly to module installation - the entire concept of upgrading tenants was accidentally removed in the DatabaseManager refactoring, and this PR resolves it. Am I correct?

chlupac:
Exactly.

sbwalker:
This fix still does not resolve the problem when installing modules. To reproduce you need to have an existing installation of Oqtane. Then you copy the Oqtane.Module.Blogs.0.0.5.nupkg package into the Oqtane.Server\wwwroot\Modules folder. You then run Oqtane. 

First it calls Startup:
        public void ConfigureServices(IServiceCollection services)
            services.AddOqtaneModules();
( this scans the /bin for assemblies which need to be manually loaded, as .NET Core does not load all assemblies by default. At this point the Blog assemblies have not been unpacked ).


Then it calls Program.cs:
                manager.StartupMigration(); 
( This is where module scripts are noe executed - but Blog module not unpacked yet )

Then it calls Startup:
       public void Configure(IApplicationBuilder app, IWebHostEnvironment env, IInstallationManager installationManager)
            // install any modules or themes
            installationManager.InstallPackages("Modules,Themes", false);

( this is where is unpacks the Blog module assembly=ies into the /bin ).

The result is that once the app is up and running, the Blog module is not loaded in the AppDomain and its install scripts were not executed. If you close down the app and re-run it will load the DLLs ( because they are in the /bin now, and run the scripts ). 

This is not acceptable as the expectation is after the initial restart the module should be installed and available. This was working properly prior to the recent refactorings.

chlupac:
1. It's a bug, In current version nuget is unpacked after migration. Scripts are called after app restart

2. If I call install manager before migration dll's are correctly unpacked but without app restart will NOT be loaded in app domain. It's same as before, statement "This was working properly prior to the recent refactoring." is wrong, bug was presented already here.

 I tested latest version before DM. 
image
files unpacked
but in app domain not presented
image
zero length of assemblies array

3. To load assemblies we can
+ restart app again in startup to ensure all references loaded (easy way)
+ or load assemblies manually. some solution is here but i'm not sure it's not easy and required some testing.
https://stackoverflow.com/questions/658498/how-to-load-an-assembly-to-appdomain-with-all-references-recursively

sbwalker:
This was definitely working before the recent changes, as this is how I tested the Blog module when I was developing it.

sbwalker:
The order of operations changed in the DatabaseManager refactoring - the call to manager.StartupMigration() did not use to occur in Program.cs. I will go back through the history today to try and figure out where this logic was executed previously.

chlupac:
I test version before database manager today to see how it works, and this work only in style unpack > restart. You do not load assemblies dynamically and here is no way how to load them without restart.
If you do not have assemblies in bin before app start they will not be loaded.

sbwalker:
Does the PR that you submitted today resolve the issue?

chlupac:
Yes, but with restart after module unpack. 
unpack nuget > restart > migration > run
You are also restarting app when you install module with GUI

chlupac:
no problem with IIS, problem with kestrel, which do not restart automatically.

sbwalker:
I suppose it might be possible to load the new assemblies into the app manually after the DLLs are unpacked - which would theoretically avoid the need for a restart. The problem is that if the module has controllers it also needs to register them ie. AddOqtaneApplicationParts

chlupac:
If we have method for loading dll to app we can call migration script on assembly in moment when assembly is loaded. Unpacking method can be static and run before buildwebhost.

sbwalker:
The main problem is that any real module will have Controllers... and they need to be added ie. mvcBuilder.PartManager.ApplicationParts.Add() - but I am not sure if this can be called anywhere except in startup. There is no point in making the other changes to unpack DLLs and call scripts unless the Controllers can also be registered.

sbwalker:
I know that this was working at some point because I spent a lot of time on it, so I dug back through the history and I found some differences in the approach which was taken in the past. I am not saying this was caused by the DatabaseManager refactoring - it seems to have happened before that. 

First the Startup ConfigureServices was called which used the installation manager to unpack any module or theme packages, and then executed the logic for loading the assemblies into the app domain:

https://github.com/oqtane/oqtane.framework/blob/2607d4dbb35251bb1c902488975d4f2fd2569e39/Oqtane.Server/Startup.cs

        public void ConfigureServices(IServiceCollection services)

            ServiceProvider sp = services.BuildServiceProvider();
            var InstallationManager = sp.GetRequiredService<IInstallationManager>();
            InstallationManager.InstallPackages("Modules,Themes");

            ...

            // get list of loaded assemblies
            Assembly[] assemblies = AppDomain.CurrentDomain.GetAssemblies();
            string path = Path.GetDirectoryName(Assembly.GetEntryAssembly().Location);
            DirectoryInfo folder = new DirectoryInfo(path);
            List<Assembly> moduleassemblies = new List<Assembly>();

            // iterate through Oqtane module assemblies in /bin ( filter is narrow to optimize loading process )
            foreach (FileInfo file in folder.EnumerateFiles("*.Module.*.dll"))
            {
                // check if assembly is already loaded
                Assembly assembly = assemblies.Where(item => item.Location == file.FullName).FirstOrDefault();
                if (assembly == null)
                {
                    // load assembly from stream to prevent locking file ( as long as dependencies are in /bin they will load as well )
                    assembly = AssemblyLoadContext.Default.LoadFromStream(new MemoryStream(File.ReadAllBytes(file.FullName)));
                    moduleassemblies.Add(assembly);
                }
            }

Then the App component kicked in and called IsInstalled:

https://github.com/oqtane/oqtane.framework/blob/2607d4dbb35251bb1c902488975d4f2fd2569e39/Oqtane.Client/App.razor

Which called the InstallationController which then called InstallationController. It contained the logic for executing any module scripts ( see InstallModule(assembly, connectionString); ):

https://github.com/oqtane/oqtane.framework/blob/2607d4dbb35251bb1c902488975d4f2fd2569e39/Oqtane.Server/Controllers/InstallationController.cs

This approach called the items in the correct order so that a deployed module package could be fully installed.

chlupac:
What should be goal of this exercise? 
Install packages without restart application? 
I will test this method and if it works I have idea how to load packages dynamically without need of system restart.
```c#
                // load assembly from stream to prevent locking file ( as long as dependencies are in /bin they will load as well )
                assembly = AssemblyLoadContext.Default.LoadFromStream(new MemoryStream(File.ReadAllBytes(file.FullName)));

```

sbwalker:
The goal is to be able to place a nupkg package in the wwwroot/Modules folder ( which can happen by manual copying, by uploading through the UI, or downloading from Nuget through the UI ) and then when the framework is restarted, the module is fully installed and available for use in the UI.

sbwalker:
It was my understanding that #351 currently requires 2 framework restarts which does not provide a good experience on all web servers. So I was trying to find the approach which worked with one restart. I will test #351 and provide more detailed feedback.

chlupac:
if you upload nuget by interface, nuget is unpacked by installation manager and after restart migration is finished. 1 restart.
If you place nuget to folder manually, here is possibilities
1. restart > unpack >restart > install 
2. Create job  which scan install folder > automatic unpack > restart > install
3. Unpack with module installer > restart

chlupac:
I believe here is way to do it safely without restart. I did some research and it's interesting theme.
In Nuget loading we probably have to: 

1. collect assembly names from nuget
2. unload old version of assemblies from domain
3. Unpack, replace old assemblies 
4. load new version of assembly to domain
5. register services and db contexts from assembly
6. register controllers
7. database migration

But I'm not sure if it's possible under user load Maybe here must be some "maintenance mode"
during this task.



interesting but not actual
https://shazwazza.com/post/custom-assembly-loading-with-aspnet-core/
or here
https://weblog.west-wind.com/posts/2019/Nov/05/Dynamically-Loading-Assemblies-at-Runtime-in-RazorPages
https://github.com/ExtCore


In my NET 4.x apps i'm using Managed Extensibility Framework for managing plugins and modules with great success.

sbwalker:
I have done some work with the new appdomain loading/unloading in .NET Core and had inconsistent results. For me it resulted in overly complicated code which was difficult to debug/maintain. So my preference is to keep this simple for now.

Let me provide some more context in regards to the module package deployment story. I mentioned it in #348 but once we ship 1.0 the main type of development which will be done on Oqtane is creating custom modules. In my opinion, the preferred approach for this will be the "external" model where modules are created outside of Oqtane. In this model there needs to be a streamlined way to deploy the module package for installation/testing. 

Currently the Blog module has a Package" project which in Release configuration automates the creation of the nupkg package and copies it to the wwwroot/Modules folder. The concept is that when you run Oqtane, it sees the module package and fully installs it for you so that it is immediately available. This is much more efficient than forcing a developer to have to run Oqtane and navigate to the Module Management UI, upload the package manually, and then re-run Oqtane to force the installation.

Can you think of a technical reason why the approach I described earlier in this thread ( which was functional at one point ) will not work anymore? Does DatabaseManager need to be called in Program.cs?

chlupac:
Base of problem is than new version of assembly from package is not loaded to domain before configuration starts.

Database Manager can be called anywhere. for example is called when new tenant is created.

Reason why is database manager called in main is than i want to do all maintenance task before web server start. I do not have in mind this developer scenario. But this also requires load libraries to app domain. 
Maybe can be possible to unpack and load dlls to domain before any config happens. Only missing parameter which i do not know in this moment is where is placed wwwroot folder. All other functions can be static without references to services.
 
For development of modules I'm using extra project (loader) which references all oqtane libraries and developed module. Module is placed to bin during build. I place module in start page of site and I can instantly debug it. I described this partially in #211.

sbwalker:
Your development approach is the opposite of what I was expecting developers to use. I was expecting Oqtane to be the host application and developers to create Blazor class library extensions which would be installed into Oqtane. This is the approach in DNN and many other frameworks. Your approach requires a developer to build a host application and then include Oqtane as a class library. If your approach works for you that is great as the framework was not developed with this use case in mind. I am wondering how your approach would support Blazor WebAssembly - or maybe that is not a requirement for you.

sbwalker:
So I managed to track down where the "breaking" change to module installation occurred - and I am responsible. The code in Startup - ConfigureServices used to call:

            // install any modules or themes
            ServiceProvider sp = services.BuildServiceProvider();
            var InstallationManager = sp.GetRequiredService<IInstallationManager>();
            InstallationManager.InstallPackages("Modules,Themes");

Which allowed the functionality to work properly as it unpacked the DLLs to the /bin BEFORE the logic executed to load them into the app domain. However this code generates a compiler warning message when you build. The warning comes from this line:

            ServiceProvider sp = services.BuildServiceProvider();

and the warning is:

"Calling 'BuildServiceProvider' from application code results in an additional copy of singleton services being created. Consider alternatives such as dependency injecting services as parameters to 'Configure'"

I got tired of seeing this warning, so I moved the logic to the Configure() method as it suggests. However this broke the module installation functionality, I am going to move it back, as it is important that the packages are unpacked before the DLLs are loaded.

chlupac:
Give me few days, i'm on good way to solve this problem. You do not need to make instance of installation manager, when you make core methods static. 

I will also test my loader with webassembly. But I do not expect any problems here.

sbwalker:
the modifications were not difficult. I have already made the changes and I am testing various scenarios now and they are all working.

thabaum:
`InstallationManager.InstallPackages("Modules,Themes");`

Could this relate to the installation of the theme package as well?

sbwalker:
I don’t think so. I tried to recreate #356 in my installation before I made this change and I was not able to.

PR #374, opened 2020-04-14
Fix register

mikecasas:
This just hides the UI if the site does not allow registration.  I will get the controller later. You can wait until I add that, or accept this as-is.

### DIFF
--- a/Oqtane.Client/Services/UserService.cs
+++ b/Oqtane.Client/Services/UserService.cs
@@ -11,12 +11,14 @@ namespace Oqtane.Services
         private readonly HttpClient _http;
         private readonly SiteState _siteState;
         private readonly NavigationManager _navigationManager;
+        private readonly ISiteService _siteService;        
 
-        public UserService(HttpClient http, SiteState siteState, NavigationManager navigationManager)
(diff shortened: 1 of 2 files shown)


mikecasas:
I added the check on the controller.

sbwalker:
@mikecasas rather than using CSS properties which simply make content invisible, it may be better to simply wrap the markup in a @if (PageState.Site.AllowRegistration) { registration markup } else { < div>Registration is Disabled< /div> }. The benefit is that the markup will not even be rendered by the component.

sbwalker:
@mikecasas I can make the change I suggested after I merge this PR if that is your preference?

mikecasas:
Sure your preference, you are the boss :)

Oqtane.Client/Services/UserService.cs
```diff
@@ -36,6 +38,13 @@ public async Task<User> GetUserAsync(string username, int siteId)
 
         public async Task<User> AddUserAsync(User user)
         {
+            Site site = await _siteService.GetSiteAsync(_siteState.Alias.SiteId, _siteState.Alias);
```
chlupac, line 41: Here was bug, repaired.

sbwalker:
I just fixed it in #381

PR #380, opened 2020-04-16
Alias is unknown at time of installation.

mikecasas:
Use the other overloaded method when initially installing the site.


sbwalker:
I do not know what this is fixing? I the installer is working fine in the current code.

sbwalker:
closing - resolved in #381

PR #384, opened 2020-04-17
changed ServerAssemblyName to ServerManagerType to optimize processing  

sbwalker:


sbwalker:
Note that this is a breaking change

PR #387, opened 2020-04-17
OS independent file paths

SPSeanLong:
Resolve issues for unix file paths '/' by using System.IO.Path for building file paths and checking for both DirectorySeparator & AltDirectorySeparator characters where applicable


SPSeanLong:
Tested windows/linux

- Install

- New Site

- New/Edit Page

- New Tennant

- Image upload/download

PR #389, opened 2020-04-18
allow users to modify default module names, descriptions, and categories and improve control panel behavior

sbwalker:


Oqtane.Server/Scripts/Master.00.00.00.sql
```diff
@@ -40,6 +40,9 @@ GO
 CREATE TABLE [dbo].[ModuleDefinition](
 	[ModuleDefinitionId] [int] IDENTITY(1,1) NOT NULL,
 	[ModuleDefinitionName] [nvarchar](200) NOT NULL,
+	[Name] [nvarchar](200) NULL,
```
chlupac, line 43: This should be Master.00.00.03.sql

sbwalker:
We are not at the stage where we are trying to provide upgrades yet. In fact it explicitly says this on the repo home page. I created an issue this morning titled Upgradeable and Backward Compatible giving people guidance that we will he transitioning to that phase soon.

sbwalker:
I am glad we are ready for it once we transition though ( thanks to the improvements you made in DatabaseManager )

chlupac:
But this is nice use case to test it. :)

sbwalker:
We are also going to have to make some changes to the framework to support upgrades. We are missing some concepts to provide the full solution. I will provide an explanation shortly.

sbwalker:
Oqtane needs to support install/upgrade for the core framework as well as modules. It is impossible to imagine every requirement a custom module might need so the install/upgrade capability needs to be flexible. Currently Oqtane supports upgrades at the database level. This is definitely part of the overall solution but it is not comprehensive enough to cover all use cases. There will be many install/upgrade scenarios which can only be handled through code ( ie. updating config files, managing files or folders, calling core APIs, etc... ). 

One of the first examples which comes to mind is for the core framework. Once 1.0 released there may be a need to add an additional admin page. The SiteTemplate handles this well when creating new sites but it does not handle sites that already exist. There needs to be a mechanism during the upgrade process to call a method based on a version# and execute some code logic.

In order to support this concept we need a few enhancements:

1. We need to track the currently installed version of upgradeable assets. Currently this exists in a few different places,  but probably not the right places. At the very least we should be storing the version in the ModuleDefinition table.

2. We need an interface  which allows us to execute version specific upgrade/install logic. It would be a simple interface such as IInstallable with a couple generic methods Install(version) and Uninstall(). The framework would use the currently installed version# of an asset ( ie. either the core framework or a module ) and it would iterate through each subsequent version until it reached the latest version which is stored internally in the asset itself. Each iteration would execute whatever logic is needed which is specific to that version.

3. The IModule interface would need an additional property for "ReleaseVersions" which would contain a comma delimited list of every version number of the module which was ever officially released. This list will be used by #2 above to know how to iterate through the versions. The core framework would also introduce a similar constant to track official release versions.

4. Install/Upgrade activities should be logged in the standard logging system for each tenant that they occur in. A new "function" enum value could be added for Install so that it is easy to identify the events.

5.  The is the most controversial item on the list, but DBUp does not fit well into the requirements above. It works well for database scripts however it controls the entire upgrade process and does not allow you to execute only a single script for a version. In reality, the internal logic of DBUp basically does the same thing described above in #2 - but is limited to databases. So the question is if we should consider moving away from DBUp in favor of our own install/upgrade logic? Based on the goal to move towards a more code-focused migration approach, it seems to make sense. Also one of the guiding principles is to minimize third party dependencies, so it makes sense from this perspective as well. However, this would be the highest risk change of the items on this list. 

6. Uninstall would need to change and leverage the IInstallable interface method. This would provide a looser coupling and a lot more flexibility.

7. Themes are another extensibility point of Oqtane and could also be enhanced to support upgrades in a consistent manner.

sbwalker:
my post above should probably be posted in issue #390 so that it is visible

PR #404, opened 2020-04-26
Update external module packages to preview 4.

mikecasas:


sbwalker:
I just merged the PR to move to Preview5 - so the module templates will also need to be moved to Preview5

mikecasas:
Updated to Preview 5.

PR #411, opened 2020-04-27
Theme base components refactoring

chlupac:


sbwalker:
@chlupac this PR resulted in a bunch of warnings: warning CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread.

chlupac:
Yes, i know. Some of methods used by delegate Action are not async. I have to make further investigation how to clean up this warning.  for example this:
```
 Action = async (u,m) =>  Settings(u, m)
```
Settings is not async method, is called synchronously.

Issue #410, opened 2020-04-27
Custom theme not being picked up as the default.


mikecasas:
When you do a new site install and set a custom theme in appsettings, it doesn't get picked up. The ```installer.razor``` file just uses the defaults that are in the constants class. There is logic sitting in ```DatabaseManager.cs```, to read the config settings, but it is not called on the initial install.

Should logic be added to SiteRepository to read the Theme in app config? Or should that logic be placed somewhere else?

https://github.com/oqtane/oqtane.framework/blob/master/Oqtane.Client/UI/Installer.razor#L190-L192

chlupac:
Settings in Install section are used during automatic (silent) installation without gui.

sbwalker:
when installing through the install wizard it picks up the SiteTemplate from config so it would make sense that it could also pick up the DefaultTheme, DefaultLayout, DefaultContainer if they were specified.

chlupac:
Solved by #416, #415 is obsolete.

PR #415, opened 2020-04-28
Get custom theme and layout on a default install.

mikecasas:
This solves #410


chlupac:
This is not good place for this change.  It should be done on repository level. Send null values down and check them at site creation repository function.

chlupac:
Solved by #416

sbwalker:
This PR does not align with client/server architecture. You cannot access config files directly from the browser - so this solution will not work in Blazor WebAssembly. Closing and merging @chlupac 's solution instead.

PR #418, opened 2020-04-29
Bootswatch themes in Oqtane.Themes.OqtaneTheme

chlupac:
Theme is set in Oqtane.Client\Themes\OqtaneTheme\Default.razor as parameter of  `LoadBootswatchTheme("superhero");`

TODO : persistence and changing of current theme


hishamco:
Seems the changes are too much to review, IMHO we need to file an issue regarding Theming & Styles and decide what should we do as an open discussion

PR #421, opened 2020-04-30
Fixed add profile property

SPSeanLong:
Adding a new profile property was going through the UpdateProfile method instead of AddProfile 
 method. Now uses the AddProfile method when there was no pre-existing id as already used earlier in the same function for same destinction

### DIFF
--- a/Oqtane.Client/Modules/Admin/Profiles/Edit.razor
+++ b/Oqtane.Client/Modules/Admin/Profiles/Edit.razor
@@ -144,7 +144,7 @@
             {
                 profile = new Profile();
             }
-            
+
             profile.Name = _name;
             profile.Title = _title;
             profile.Description = _description;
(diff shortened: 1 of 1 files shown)


Oqtane.Client/Modules/Admin/Profiles/Edit.razor
```diff
@@ -154,8 +154,14 @@
             profile.DefaultValue = _defaultvalue;
             profile.IsRequired = (_isrequired == null ? false : Boolean.Parse(_isrequired));
             profile.IsPrivate = (_isprivate == null ? false : Boolean.Parse(_isprivate));
-            profile = await ProfileService.UpdateProfileAsync(profile);
-            
+            if (_profileid != -1)
+            {
+                profile = await ProfileService.UpdateProfileAsync(profile);
+            }else
```
hishamco, line 160: Format

hishamco: we can use

```csharp
profile = _profileid != -1
    ? await ProfileService.UpdateProfileAsync(profile)
    : await ProfileService.AddProfileAsync(profile);
```

sbwalker: My preference is to use the traditional conditional logic as it is much easier for developers to read and understand - which makes the code easier to maintain

sbwalker: the else should be on its line to be consistent with the rest of the framework code, but I will merge this PR anyways.

SPSeanLong: To be sure I comply correctly in the future, the desire is

`}

else

{`

sbwalker: Correct ( without the extra line breaks before and after the “else” of course ). If you look at the other framework code you will are the standard is:

If ()
{
...
}
else
{
....
}

PR #426, opened 2020-05-01
Page editor fix & Script check

chlupac:


sbwalker:
Themes are not required for pages. You need to be able to leave them unassigned so that they inherit the theme from the site.

chlupac:
Yes I know. but current version complains about it and do not let me save. 
This PR fixed it.

sbwalker:
But it is not the correct fix. Do you want me to take a look or do you want to?

sbwalker:
My apologies. I now see what you corrected - I was not reading this correctly. Thank you for fixing this.

Issue #437, opened 2020-05-04
Log Details Device Friendly UI Improvement


thabaum:
The log details information is pushed to the sides with large amounts of space between labels and input

sbwalker:
A decision was made to not migrate the admin UI from tables to divs for V1. It requires too much regression testing at this stage. Plus it results in some of the forms being very long vertically which requires a lot of scrolling affecting the user experience.

PR #438, opened 2020-05-04
Enables boostrap frienldy forms by changing from Table to DIV adding …

thabaum:
...  Enables boostrap frienldy forms by changing from Table to DIV adding

Fixes issue #437 

image

Notes:

More pages can be updates this way for the UI.  This is something that should be done prior to letting it go down the road in some form.

I was thinking about maybe later putting the labels inside the borders.  This has the readonly fix included however you can merge the other PR first then merge this one things should work just fine.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Logs/Detail.razor
+++ b/Oqtane.Client/Modules/Admin/Logs/Detail.razor
@@ -7,159 +7,103 @@
 @inject IPageModuleService PageModuleService
 @inject IUserService UserService
 
-    <table class="table table-borderless">
-        <tr>
-            <td>
-                <label class="control-label">Date/Time: </label>
-            </td>
-            <td>
-                <input class="form-control" @bind="@_logDate" disabled />
-            </td>
-        </tr>
-        <tr>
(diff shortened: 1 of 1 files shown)


thabaum:
I have one PR I am creating that does the same thing to the rest of Admin pages and some CSS tune up to go with here as soon as I can figure out how to make them all show up right in a PR from Visual Studio... or copy and paste all the files in github manually.  This PR is an example of what I did for all the pages.

sbwalker:
@thabaum changing the admin UI from tables to divs was deferred until after the MVP release

thabaum:
OK I just wanted to give it a more formal approach.  I have all of the files redone if you want to review.  Nothing special but I think makes more device friendly for now.

The class's I didn't feel I should create for anything without a deeper discussion.  I will add my changed files to this PR in the event you would like to review it as the work is done for a simple tune-up.  Anytime this type of change gets discussed it is a "Breaking Change" so I wanted to make it now while the hood is still up.

sbwalker:
yes please submit for review

PR #446, opened 2020-05-07
dashboard table tags replaced with form and div tags

thabaum:
table tags replaced with form and div tags


thabaum:
well i made one from visual studio... but i see no files haha awesome! maybe next time...

PR #448, opened 2020-05-07
Dashboard update

thabaum:
Updates the admin dashboard to use more DIVs instead of Tables


sbwalker:
It is possible to preserve the current UX appearance but still transition from tables to divs. Bootstrap has a horizontal form rendering - which displays the fields in a horizontal form layout on desktop devices, and in a vertical layout on mobile devices.

https://getbootstrap.com/docs/4.0/components/forms/#horizontal-form

My concern is that using the standard bootstrap vertical forms for everything is that it is not a great user experience in desktop as it wastes screen real estate and results in really long forms which require excessive vertical scrolling.

thabaum:
I was doing a fairly simple grouping just to put things in the right direction.   

The only thing really missing from this PR to accomplish the end result you desire is adding classes like **col-md-6** to the **form-group** nesting them inside div tags with **form-row** class

I can work those into this PR as well if ok.

thabaum:
I believe all files changes currently wont conflict with any current changes that have been merged.  If you can merge this PR I can then spend a little time adjusting things for the next layer of primer.

thabaum:
I have worked with things and found that the documentation does not work 100%... unfortunately however I did find the work around to make the examples work to get the desired results.

I will apply the results along with some other tweaks to this PR when I feel confident they are ready possibly tomorrow.

sbwalker:
One of the challenges of any large PR at this stage is that 1.0 is scheduled for May 19 ( the date is fixed because that is the date when Blazor WebAssembly will be released ).  So there is just over 1 week to have a completely stabilized release. This is the reason I had made the decision to defer the transition from tables to divs until after 1.0.

thabaum:
I will submit something to at least talk about, however I am thinking of creating one PR, and if we get even closer to 1.0 adding things up to version 1 as requested if I can get it done it can get included possibly by then, if not it get be something worked on for the next version or version 1.1.  

I ran out of time this weekend however I will be working on it.  If I can get something out by tomorrow take a look.  I am trying to do it without adding classes and CSS, purely bootstrap with some styles built into the divs, however if classes are desired let me know now how we should name them I can add some to work with.

If nothing else I will learn 1 more way not to build a dashboard :)

thabaum:
looking back to where i left off this is what I have

image
current problem
image

thabaum:
Now two issues I see can be resolved.  The help icon explains what each feature does.  Maybe we can get these all down to 1 word with a few CamelCase instances?  This would clean up the form a lot.

thabaum:
col-xl-9 and col-xl-3 for splitting divs works as desired with current sized labels with longest I can find I believe so far.  if not it will be xl-4 xl-8 size

sbwalker:
I think the best way to move forward with this item is to reduce the scope to a single module, submit a PR, allow people to review and provide feedback, and once an approach is agreed upon make the changes to the remaining modules.

thabaum:
I have a fully tuned up version I would like for you to review once I can upload changes.  I worked with the entire framework to get it to ALL layout nicely.

Only thing missing is a tuneup on the Oqtane theme I need to do for a PR soon I have noticed.

I have this looking sharp.... all pages and modules fixed a few bugs found.  It is worth a test drive.

thabaum:
Do I need to put all my changes on the latest Repo possibly a new PR?  Or can this github system figure it out?

I can upload the files here on github quickly I just made changes to latest repo 5 days ago is all...

thabaum:
this has an update to Pager and Permissions components now as well.

Issue #475, opened 2020-05-13
BUG:  Role can not be saved


svreic:
Exception is thrown if a role is saved if a long description is entered.

## Expected Behavior
The role can be saved without an exception 

## Current Behavior
Exception is thrown if the description of the role exceeds 50 character. 

## Possible Solution
The description column type should be extended from varchar(50) to varchar(256). Also the maxlength of the textarea of the description should be limited to 256 characters. 

## Steps to Reproduce
1. Add or Edit a Role
2. Write a long description
3. Click the Save button

sbwalker:
A few comments:

- currently none of the forms in any of the admin UIs validate or restrict field lengths - this is not limited to Role Management
- if we want to implement field length validation it needs to be implemented on both the client UI elements as well as the server API controllers
- role descriptions are not surfaced anywhere in the application currently and actually serve no purpose so I am not sure if the length needs to be expanded. 
- in order to modify the database schema a new tenant upgrade script needs to be added for the next releasable version

svreic:
> if we want to implement field length validation...

I suggest to do it because of stability. The current implementation leads to an exception in the db-update/insert process.

> it needs to be implemented on both the client UI elements as well as the server API controllers

I haven't thought about the API controller in that context and of course, it needs to be implemented on both sides. 

> this is not limited to Role Management

If we decide to implement field length validation than this should be implemented in the whole framework. For that reason i suggest that we create an issue with a checklist for all admin UIs that needs field length validation. I can do it if we decide to do it. Than we can do this step-by-step.

> role descriptions are not surfaced anywhere in the application currently and actually serve no purpose so I am not sure if the length needs to be expanded.

The description field currently only supports 50 characters whereas the name field supports 256 character. Mostly the description of something will be more text than the name of something. In context of UX a user mostly expect that the description can be longer than the name. Currently the UI of the Role Management looks like you can have a longer description. The filed for the name is a single line input and the description field is a multi line textarea. Just my 2 cents.

sbwalker:
I made a quick fix in #616 for this item. There is still a broader topic of form validation in general, but this addresses this specific bug which was reported.

PR #476, opened 2020-05-13
Fixing role description 

svreic:
The pull reqest is a possible solution for issue #475 

It alters the description column of the role table from varchar(50) to varchar(256). varchar(256) seem to be the standard size for all other text based columns.

It also sets the maxlength attribute for the textarea elements where the user can insert the description.
So now the user can't insert more than the supported amount of characters.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Roles/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Roles/Add.razor
@@ -1,68 +1,69 @@
-﻿@namespace Oqtane.Modules.Admin.Roles
+﻿@using Microsoft.AspNetCore.Components.Forms
+@namespace Oqtane.Modules.Admin.Roles
 @inherits ModuleBase
 @inject NavigationManager NavigationManager
 @inject IRoleService RoleService
 
-<table class="table table-borderless">
-    <tr>
-        <td>
-            <Label For="name" HelpText="Name Of The Role">Name:</Label>
-        </td>
-        <td>
-            <input id="name" class="form-control" @bind="@_name" />
-        </td>
-    </tr>
-    <tr>
-        <td>
-            <Label For="description" HelpText="A Short Description Of The Role Which Describes Its Purpose">Description:</Label>
-        </td>
-        <td>
-            <textarea id="description" class="form-control" @bind="@_description" rows="5"></textarea>
-        </td>
-    </tr>
-    <tr>
-        <td>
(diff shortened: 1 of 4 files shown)


svreic:
resolves  #475

sbwalker:
To add a change to the database it needs to be in an upgrade script associated to a new version ( in this case Tenant.0.9.2 )

svreic:
Thank you for the hint with the upgrade script.
I wasn't sure where to put the database alternation.
I thought that Tenant.0.9.1 is already the script for the next version because officially no 0.9.1 is released.

sbwalker:
I added the role description expansion to 0.9.2 - are you able to alter this PR or submit a new PR for the remainder of the changes?

svreic:
I updated and changed the PR to take comments from issue #476 into account.

Now it is a possible solution how validation could be implemented on both sides (Controller/Server and Client).

sbwalker:
I am going to defer merging this PR until after the 1.0 release. This is because the changes do not follow any of the current architectural patterns. One of the philosophical goals in Oqtane is maintaining Consistency ( https://www.oqtane.org/Resources/Blog/PostId/538/oqtane-philosophy ). This being said, I think there is the opportunity post.-1.0 to discuss new architectural patterns - and if we make a decision to change the approach then it needs to be done throughout the framework - not just in a single location.

sbwalker:
The specific areas where this PR is inconsistent with the rest of the core framework are:

- use of IActionResult rather than plain objects
- use of DataAnnotations
- use of Razor Component forms

svreic:
Thank you for your feedback.

I agree that this PR is not vital for the 1.0 release. And I also think that this PR should be seen more as a POC.

Imho: 
Currently there is no architectural pattern for value validation. So I think for robustness and stability there should be one established after the 1.0 release.  And of course, if there is architectural pattern defined how value validation should be done than it needs to be done throughout the framework - not just in a single location. But I think this shouldn't be done in one big PR because it will be hard to review such a PR. 
These fundamental changes should be done in several PRs in a module-by-module way tracked in a single issue with a checklist or multiple issues - one for each module.

Comments to my suggested solution are very welcome. Maybe there are better solutions to implement value validation without breaking the mentioned architectural patterns.

hishamco:
@svreic could please update this and show some screenshots? I think this could be merged before `2.0.0`

sbwalker:
@hishamco the main issue was already resolved - expanding role description. I believe validation was already improved using existing methods. The reason this was not merged previously was because it used a variety of techniques that are inconsistent with the rest of the core framework. I am inclined to close this specific PR.

hishamco:
If the issue is resolved we need to close this

PR #480, opened 2020-05-14
FileManager Tune-Up

chlupac:
File Manager
* Added possibility to hide folder selection and upload part to allow user only select from uploaded images or only to specified folder
* Change parameters type from string to corresponding type

~~Control Panel~~
~~* Values are not cleared when add module is pressed~~
~~* last used category is stored to options~~

~~Settings~~
~~* `public Setting Get(string entityname, int entityid, string settingName)` added possibility to query only one value by name~~

### DIFF
--- a/Oqtane.Client/Modules/Admin/Files/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Files/Add.razor
@@ -12,7 +12,7 @@
                     <Label For="upload" HelpText="Upload the file you want">Upload: </Label>
                 </td>
                 <td>
-                    <FileManager UploadMultiple="true" ShowFiles="false" FolderId="@_folderId.ToString()" />
+                    <FileManager UploadMultiple="true" ShowFiles="false" FolderId="@_folderId" />
                 </td>
             </tr>
         </table>
--- a/Oqtane.Client/Modules/Admin/ModuleDefinitions/Add.razor
+++ b/Oqtane.Client/Modules/Admin/ModuleDefinitions/Add.razor
@@ -35,7 +35,7 @@
                         <Label HelpText="Upload one or more module packages. Once they are uploaded click Install to complete the installation.">Module: </Label>
                     </td>
                     <td>
-                        <FileManager Filter="nupkg" ShowFiles="false" Folder="Modules" UploadMultiple="True" />
+                        <FileManager Filter="nupkg" ShowFiles="false" Folder="Modules" UploadMultiple="true" />
                     </td>
                 </tr>
             </table>
--- a/Oqtane.Client/Modules/Admin/Site/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Site/Index.razor
@@ -39,7 +39,7 @@
                 <Label For="logo" HelpText="Upload a logo for the site">Logo: </Label>
             </td>
             <td>
-                <FileManager FileId="@_logofileid.ToString()" Filter="@Constants.ImageFiles" @ref="_logofilemanager" />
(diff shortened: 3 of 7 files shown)


sbwalker:
Larger PRs are more difficult to review and also more difficult to accept. Case in point, some aspects of this PR could easily be merged but others need further discussion - so I cannot accept any of this PR until all aspects are resolved.

In regards to Settings, the focus of Oqtane is performance - so we are purposely trying to prevent developers from creating "chatty" modules ( i.e modules which make a lot of calls from the client to the server to deal with tiny amounts of data ). Therefore the existing Settings pattern provides 2 methods: 1 for getting all setting values for an entity, and 1 for updating the collection of setting values for an entity:

            Dictionary<string, string> settings = await SettingService.GetModuleSettingsAsync(ModuleState.ModuleId);
            SettingService.SetSetting(settings, "SettingName", _value);
            await SettingService.UpdateModuleSettingsAsync(settings, ModuleState.ModuleId);

This is an efficient pattern, does not restrict the developer, and reduces the chance of creating modules with performance issues. So I would prefer to not add the additional API method.

chlupac:
PR is cracked to smaller logical parts. File Manager part.

sbwalker:
I would like to better understand the purpose behind these enhancements. Oqtane uses an abstraction over the file system so that it can manage security, etc... Currently the FileManager control exposes management options based on the security specified for a folder - Browse, View, or Edit. I would like to make sure that the new parameters that were introduced in this enhancement are not intended to override security behaviors - as this would be problem. Also note that all registered users in Oqtane are granted a personal folder where they can manage their own files.

Ability to hide folder selection - if you are following the security model, a user can only select from folders to which they were granted Browse or View permissions. If a developer has the ability to hide the Folder Selector via a parameter then a user would not have the ability to choose files from their folders - they would only be able to choose files from a specific folder which the developer has hardcoded in their module - and how does the developer know the Folder ID which the specific user has access to? 

Ability to upload files - if you are following the security model, a user can only upload files to folders to which they were granted Edit permissions. If a developer has the ability to hide the Upload Button via a parameter then a user would not have the ability to upload files to their folders. This means that the files would have needed to have been uploaded to the folder some other way previously so that they can be selected. But again, how does the developer know the Folder ID which the specific user has access to? 

I am concerned these enhancement were added due to a lack of understanding of the folder security model. This may be because most developers who are interacting with Oqtane tend to use the host account - which means they have full access to all folders and they may feel like they need to restrict access somehow. However the proper way to do it is via folder permissions.

chlupac:
File manager has 3 main functions
1. Select folder 
2. Select for file
3. File upload

My update deals with cases when you want on particular place only one of this functions.

1. Hide folder selection. For particular module I want to direct users to store files (images) in one common folder which is prepared for them by site admin 
Developer knows folder name (Images\Avatars\) and get folder id by 
 public async Task<Folder> GetFolderAsync(int siteId, [NotNull] string folderPath)

2. Hide Upload ability. Together with 1. I want to allow select only from preloaded set of images. 
3. I fully understand security model of folders. Which view on particular edit control developer uses is on him. 
4. Hide upload section is already here, based on permissions. I only extend this model with                 `@if (ShowUpload && _haseditpermission) ` 
4. Pass folder name ability is unaffected and unused.
5.  security model of control is unchanged

sbwalker:
Thank you for the clarification. It is not always easy to understand the intent of changes by reviewing the code - which is why I sometimes ask for the use case. This enhancement makes sense based on what you described and it is aligned with the security model.

chlupac:
image
Use case for hide folder selection. Site banners are stored in banners folder

PR #481, opened 2020-05-14
container class added to pane

thabaum:
resolves issue with title border DIV not utilizing 100% pane size


thabaum:
removing `class="row px-4"`

Resolves issues with 100% widths in 1.3  I think the container can be either way it really does not make much difference but we can put a switch in for choosing one or the other.  I agree this can be thought out better for v2.

PR #516, opened 2020-05-20
Reduce FOUC between pages

jimspillane:
Convert a relative URL to an absolute URL before testing against the href value.  This will reduce the number of link refreshes if the same URL is used.


sbwalker:
FOUC = "Flash Of Unstyled Content" - I was not familiar with that term ;) Just so I understand, you are saying that the JavaScript comparison of (link.href !== url) is always true currently because link.href is an absolute url and url is a relative url?

jimspillane:
Yes.  Changing the _url_ parameter to absolute will help stop the reloading the same resource if it is relative.

sbwalker:
would this also be relevant in the includeScript() function?

jimspillane:
Yes. I will change it to run in the includeScript() function too.  Thanks!

jimspillane:
I added a modification that addresses window.interop getting clobbered by the quill-interop when the editor is opened.

jimspillane:
I am going to close this PR because the CSS/JavaScript dependency issue is not resolved with this change

I will submit the namespace modifications in a separate request.

PR #521, opened 2020-05-21
Simple fix for a null page.

mikecasas:
It seems when you add a new page for a site, let's say 'Test', and then delete the page 'Test', and then add 'Test' back again, the app crashes. This is partly because the row/record isn't really deleted in the database, but the IsDeleted field is set to 1.

There can be further discussion on this issue, but for now, this PR simply handles the null page and dispalys an error.


sbwalker:
I am not sure if this is the right fix. It seems to deal with the symptom rather than the cause.

mikecasas:
I guess this would be just a temporary fix until discussion could be had on the page model. I am not sure why the IsDeleted field is there, but I am sure you had a valid reason in your original design. Instead of taking a lot of your time to discuss, should we not just triage these items, and create a proper backlog?  

Some of the "quick fixes" go to your point of creating a better experience for developers just starting out with the framework. I would rather have an error message displayed, then the app to crash or become unresponsive.

sbwalker:
Soft delete is supported for Pages and Modules - this is the purpose of the Recycle Bin in the Admin Dashboard, it allows to restore them or hard delete them.

sbwalker:
@mikecasas the "add default parameters" fix which was attached to this PR cannot be merged for the following reasons:

- we want the initial installation of Oqtane to be configured for local development - which is why LocalDB is the default option. We do not want to change this as many people do not have a full SQL Server installation.
- it is never a good practice to hard code a default password in any application as it greatly increases the risk of the application being compromised.

mikecasas:
I am so sorry, obviously the changes you mention above "sneaked" in from my master branch today. I have those default parameters set as a time saver when I want to do a fresh install. They were not intended for the framework. I will close this PR.

PR #524, opened 2020-05-21
Fix site router.

stephanlotter:
When page name ends in the same value as the alias the path becomes invalid.

The scenario I had was that the alias was the word **fusion** and the page name was **homefusion**.

This lead to **fusion/homefusion/** becoming **home**. 

The consequence was that I kept getting "Page Does Not Exist Or User Is Not Authorized To View Page home" when I tried to edit the page. After applying the fix described in the pull request the page editing feature worked as expected.

### DIFF
--- a/Oqtane.Client/UI/SiteRouter.razor
+++ b/Oqtane.Client/UI/SiteRouter.razor
@@ -173,7 +173,7 @@
 
             if (alias.Path != "")
             {
-                path = path.Replace(alias.Path + "/", "");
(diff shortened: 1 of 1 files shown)


sbwalker:
this is a good catch. since the alias is always located at the beginning of the Url I think the better solution would be:

                path = path.Substring(alias.Path.Length + 1);

Do you want to make the change to your PR so I can merge it?

sbwalker:
I fixed this using the change above in #534. Thank you very much for reporting this issue.

PR #538, opened 2020-05-25
Site Settings: removes ONLY " <Select Container>"Theme and Layout drop down options

### LINKED ISSUE #537: <Select Theme> option gets saved to db error loading page theme
image

To Reproduce this issue:
- Installed Oqtane Current Repo
- Clicked on Gear
- Clicked on Site Settings
- Changed site theme to `<select theme>`
- clicked Save

thabaum:
Fixes #537 

_Edited:_

**Select Theme, Select Layout** and **Select Container** options removed from the dropdown lists for selecting site default theme, layout and container.

    '<Select Theme>'  *** REMOVED ***
    Blazor
    Oqtane

**Reason:**
1.  This option will brick Oqtane Dashboard
2.  Oqtane is default option, why include "Select Theme" when you have label and help icon, too redundant and makes something more for your eyes to view that is irrelevant in a way self explanatory.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Site/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Site/Index.razor
@@ -56,7 +56,6 @@
             </td>
             <td>
                 <select id="defaultTheme" class="form-control" @onchange="(e => ThemeChanged(e))">
-                    <option value="">&lt;Select Theme&gt;</option>
                     @foreach (KeyValuePair<string, string> item in _themes)
                     {
                         if (item.Key == _themetype)
@@ -77,7 +76,6 @@
             </td>
             <td>
                 <select id="defaultLayout" class="form-control" @bind="@_layouttype">
-                    <option value="">&lt;Select Layout&gt;</option>
                     @foreach (KeyValuePair<string, string> panelayout in _panelayouts)
                     {
                         <option value="@panelayout.Key">@panelayout.Value</option>
@@ -91,7 +89,6 @@
             </td>
             <td>
                 <select id="defaultContainer" class="form-control" @bind="@_containertype">
-                    <option value="">&lt;Select Container&gt;</option>
                     @foreach (KeyValuePair<string, string> container in _containers)
                     {
                         <option value="@container.Key">@container.Value</option>


sbwalker:
This is not the correct solution - the ability to set default theme, layout, and container at the Site level is an essential functionality that cannot be removed. The real technical problem in #537 is related to the fact that a dropdown option with a value of "" is not handled correctly in Blazor. Rather than returning "" it will return the text instead. You will notice in other components that when a default dropdown element is needed it typically sets the value to "-" deal with this issue ( see the Page/Edit component as an example ).

thabaum:
This does not remove functionality... it removes options

thabaum:
I think those options can be disabled so they cannot be selected.  

I am thinking in this case there is always a theme so whichever one is used will be selected.  The <select theme> option I removed because why does it need to be there?  This cleans up the list of options.

Do these options really need to be here?  If so I will see if we can disable selecting it as an option.

sbwalker:
the ability to be able to set and modify a default theme, layout, and container at the Site is an essential feature - it should not be removed or disabled.

thabaum:
again this is not removing the feature...

- <Select Theme>   *** REMOVED ***
-  Blazor
-  Oqtane

thabaum:
I edited the first post here to more clearly describe what this is fixing in my opinion.

thabaum:
This is a good example of why making a clear description of what you have done is important.

I will take a look at the if statement sounds simple enough.  I will play with it for a bit get to know that part of the UX

thabaum:
GitHub does not make it very easy to post here things with code it is a trick of it's own to get things to view right as well..   Things rarely look how you think they will after you comment.

You have to code your comment in :)

sbwalker:
There is a default theme/layout/container which exists for all Oqtane installations. However an installation can have many different Sites. Each Site needs to be able to specify its own default theme/layout/container which will be used when rendering its Pages. A Site can choose from any of the themes which are available - not just the ones which are part of the default installation. It is also possible to override the theme/layout/container at a Page level. And it is possible to override a container at a Module level.

sbwalker:
sorry I finally understand - you are saying that since it is mandatory for the user to specify a Default Theme, that the option for < Select Theme > is not relevant. Correct?

thabaum:
Bingo

thabaum:
It does raise the error handling concern here however this PR will not handle it.  Quick fix ;)

thabaum:
image

Blazor theme does not have a layout... this needs to change....  I can leave the options and just remove the < select theme > option for now leaving the others or we can deal with the error.

thabaum:
Create a basic layout page for the blazor theme is another option.

sbwalker:
Themes don’t need to include a Layout - it’s an optional feature that a theme can choose to implement.

thabaum:
But does the option for the layout need to be there?  It would be blank or the first one default selected unless set in this PR.  (Not say <select layout> since there is no layout to select.

I was going to try just disabling the option so that they cannot be selected.  Right now it is not an option in this PR.  I will play with it again soon.

sbwalker:
I will merge this PR but will likely make some additional modifications to it to address some of the other feedback.

PR #542, opened 2020-05-25
Oqtane Theme:  Adjustment to nav items padding and size

### LINKED ISSUE #467: Logo is not straight
Is there a reason the logo is not straight and skewed a bit? It's not skewed enough to make it seems purposeful and almost seems like it's a mistake.

thabaum:
I believe this may help the following issue with the Oqtane Theme.
Does not by may... fixes #467 (discovered while reviewing this issue)

This PR will:

- Adjust App-Menu Nav item CSS for Oqtane Theme
- Nav Items will be bigger and the bottom of the buttons will align with the bottom of the 'oqtane' characters
 
Screenshots:  

image

image

Notes:
I think the offsetting of the logo should be done with CSS instead of adjusting the logo position in the palette.  This way any logo uploaded will look good and this CSS will need adjusted to allow that to happen along with the logo.

A fixed up version of the logo (Centered) is below (White) I can adjust this PR for this logo which I recommend.  Or you can change it using your app so the images stay a same or I can create another PR submitting the logo fix for the Oqtane theme with the CSS adjustments later.

logo-white

Or we can close this if not an issue or desired solution for Oqtane Theme style.

### DIFF
--- a/Oqtane.Server/wwwroot/Themes/Oqtane.Themes.OqtaneTheme/Theme.css
+++ b/Oqtane.Server/wwwroot/Themes/Oqtane.Themes.OqtaneTheme/Theme.css
@@ -12,8 +12,8 @@ body {
 }
 
 .app-menu .nav-item {
-    font-size: 0.9rem;
-    padding-bottom: 0.5rem;
+    font-size: 1rem;
+    padding-bottom: -0.5rem;
 }
 
 .app-menu .nav-item a {


sbwalker:
Was this tested in a mobile rendering?

thabaum:
I will take a closer look at this, maybe just close it.

I see one issue in mobile responsive that I want to look closer at. 

image

Top of header title is cut off by the navigation maybe due to the login shifting the content pane.  Just looks like if you took that login button space and added that to the pane top things would be fine...

PR #544, opened 2020-05-26
Generate nugets in correct format

chlupac:


sbwalker:
Thank you for this PR... I have a few questions....

1. Currently the Module Creator templates as well as the sample module and theme on Github ( ie. https://github.com/oqtane/oqtane.blogs ) use the specific Oqtane.Client, Oqtane.Server, and Oqtane.Shared package references. This PR removes the creation of these packages from release.cmd. Is this because you feel there should be an individual cmd file for each package? Or are you indicating that the individual packages are not needed ( which is not true because there are already applications depending on them ). Or was this just a mistake?
2. Currently the Oqtane.Framework package includes its static resources in a specific structure. The Update Service relies on this package structure to extract the resources to the appropriate location. This change breaks the Update Service functionality. How would this new package structure be integrated so that it still preserves the required functionality?

chlupac:
1. `dotnet pack  -o .\ -c Release ..\Oqtane.sln` makes individual packages for all projects in solution which has    `<IsPackable>true</IsPackable>` set. Just try it from commandline.
2. Oqtane.framework package is untouched, can be used for updating framework.

PR #551, opened 2020-05-28
Uri extensions for read query values

chlupac:
- safe way to check if key is presented
- safe parsing int values in query
- should replace int.Parse() when reading query string

for example
  `_jobId = Int32.Parse(PageState.QueryString["id"]);`
  can throw unhanded exception when "id" is not int

  correct way
`  if (PageState.Uri.TryGetQueryValueInt("id",out _jobId)) ....`


sbwalker:
yes you correct that the current code is making assumptions that the querystring parameter exists and that it is in the proper form to convert to an integer. Both wrong assumptions.

PR #556, opened 2020-05-28
Patches/fix544

### LINKED ISSUE #554: ThemeRepository does improper Namespace-Prefix-Matching
I have two themes in the Oqtane.Client, one called `BootstrapInstant` and one called `BootstrapInstant4`. Both have a container called `Invisible`. Now the UI always crashes claiming that the same Container is being added multiple times to a dictionary. 

image

Turns out the code in `ThemeRepository.cs` is incorrectly checking the namespaces, by just doing a trivial string-prefix check, so that all the containers in `BootstrapInstant4` also get assigned to `BootstrapInstant` since the start the same way. 

image

I'll work out a fix and push asap.

iJungleboy:
Fix #554

Basically checks if namespace is == or begins with Namespace + ".".

Also centralized some of the LINQ code to only run once (faster)


sbwalker:
@iJungleboy I need you to sign the CLA so that I can merge this PR

PR #558, opened 2020-05-28
Menu @ Breadcrumb fix  Use Page.Title when presented

chlupac:


sbwalker:
the page title was only expected to be used as the browser dcument.title ( ie. for the browser header or tab title ):

       // set page title
        if (!string.IsNullOrEmpty(PageState.Page.Title))
        {
            await interop.UpdateTitle(PageState.Page.Title);
        }
        else
        {
            await interop.UpdateTitle(PageState.Site.Name + " - " + PageState.Page.Name);
        }

The value provided for this field is expected to be more than just a few words - it could be an entire sentence. Therefore I don't think it will work very well for a breadcrumb navigation menu?

chlupac:
Ok, i'ts my misunderstanding what this field mean.

PR #574, opened 2020-06-01
OqtaneIgnore implementation fix

chlupac:
To avoid false detection of base types


sbwalker:
Is a change from a public class to an abstract class a breaking change? This will need further testing before it can be merged.

chlupac:
no it's not breaking change. It only prevents base class from instantiating.

PR #576, opened 2020-06-01
Added support for friendly names and thumbnails in theme, layout, and container components. Added fallback support during loading for themes, layout, and containers.

sbwalker:


sbwalker:
Note that I wanted to move ITheme, ILayoutControl, and IContainerControl to the Oqtane.Shared project; however, this broke backward compatibility as themes built against earlier Oqtane versions are expecting those interfaces to reside in the Oqtane.Client project.

PR #581, opened 2020-06-03
restrict user data leakage

sbwalker:


Oqtane.Server/Scripts/Tenant.1.0.1.sql
```diff
@@ -31,3 +31,9 @@ CREATE UNIQUE NONCLUSTERED INDEX IX_File ON [dbo].[File]
 	[Name]
 	) ON [PRIMARY]
 GO
+
```
chlupac, line 34: You have to create NEW script, not add to existing one. name it Tenant.1.0.1.NotificationFix.sql. this will not run on existing install

sbwalker:
My assumption for database migrations was that we needed to be able to upgrade users from one "release" version to the next. So if the previous release version was 1.0.0 then then migration scripts for the next release would all be placed in 1.0.1. And during an upgrade from one release to the next it would execute the script.

What you are referring to is allowing a developer to upgrade their local installation based on the latest code in the repo. I am assuming you would only want to do this for development purposes - ie. to keep your development environment in sync with the latest code ( you would definitely never ever want to deploy the latest repo code into a production/release environment ). 

During the course of development on a release there may be a variety of database migrations required - so the question is if/how to handle them. In DNN we decided that it was up to the developer to monitor the check-ins and sync their local development environment based on changes to the migration script. We did this because during development there are often cases where a migration needs to be rolled back or modified based on additional feedback - and it can be very challenging to expect developers to create an upgradeable migration script for every single database change during development - and you certainly would not want all those rollbacks, etc... to make their way into the actual release migration script as it introduces too much risk of additional problems.. We decided it is more practical to take a stance that during development of a release, developers who are pulling the latest code will need to sync their own environment. 

I realize that DBUp supports the concept of multiple scripts per version but in the long run I do not want Oqtae to be tied to DBUp - I would like for it to use EF migrations. So I do not want to base our development approach/process on the capabilities of DBUp.

chlupac:
I disagree. Even for testing purposes you may want to have fully developed site and incrementally implement changes. Regenerate this status after every database update is time consuming. Even with database migrations you will end with atomic migration for every change batch. 
I understand, you want to have code clean. But doing this is unpractical and against principles.

sbwalker:
My point was not really about clean code - it was about the process flow. To be more clear, I would like to get your feedback on how you would deal with the following scenario:

A granular script is merged during development which ends up having a bug in it or it is the wrong solution to a problem. However based on your preference, the script has already been automatically integrated with the local databases of other people who are working against the latest development code. So would a new granular script be created to undo the previous script? And if so, would you include all of these scripts as part of the actual professional release package? Or would you modify the script to remove the bug? Or is there another approach I am missing?

I experienced these scenarios many times when developing DNN which is why we established the process I described above. I am open to other options but I need to understand the details of how it would handle these common issues.

chlupac:
https://paulstovell.com/database-deployment/

Yes I have to make script to restore bug

Yes, I will include all these scripts. 

Scripts should have good naming convention and not to be tied to version. When you reach version 10,  current naming convention collapses and scripts will be executed in wrong order.

Here is no other way how to do it, because your concept work only in scenario when you install clean database all the time. It's about incremental maintaining of state. No matter if it is nightly builds, fresh hotfix code from github or stable release. You should be able to change state from any previous state to last one. If you change already imported script, you broke this algorithm.

sbwalker:
Perhaps I did not do a good job of explaining myself previously...

Based on past experience the vast majority of developers ( 90% ) who use a framework like Oqtane will choose to use the release versions. This is because they need to rely on the stability, reliability, etc.. which official releases provide for their production applications. In this scenario, Oqtane absolutely needs to be able to provide a seamless upgrade path from one release version to another, and upgrade paths from various releases should be regression tested to ensure they work flawlessly.

There is another scenario for developers ( <10% ) who are willing to accept the non-stable nature of the development code in order to experiment with the latest features. It would not be recommended they ever run this code in a production environment because it is unstable. In this specific scenario I was saying that at DNN it was up to the developer to monitor the development migration script to determine which commands had already been executed on their local database and which commands needed to be executed - it was not automated. However it was not required to install a clean database every time. If you are a developer following the latest code then it is not difficult to determine which script commands need to be run on your existing database. 

I do agree with the statement that a script for a prior release should never be modified - this was a rule for DNN as well.

As far as naming conventions for scripts I disagree with you. I believe that it is natural and intuitive for migration scripts to be aligned with the version of the product they belong to. However because of the internal behavior of DBUp, the filename format of the current scripts needs to change. The filenames should be in the form Tenant.##.##.## ( this was the format used in DNN ). And if we want to support "patch" scripts like you are suggesting above then we should use the semantic version 4th element which is reserved for specifically this purpose ie. Tenant.##.##.##.## - then the first 3 elements would correspond to the app version and the 4th element would allow for "patch" scripts.

chlupac:
but current script is 
tenant.1.1.1
tenant.1.1.11
tenant.1.1.12
tenant.1.1.2

sbwalker:
Yes, I am saying that the current naming convention for scripts is wrong - it is a bug that was not caught during the various refactorings of the installation logic. And it is because of DBUp behavior. It should be changed.

sbwalker:
unfortunately it is not easy to fix because of DBUp ( this is a good example of why we want to avoid third party dependencies as much as possible )

chlupac:
Personally I like timestamped migration files. Each file starts off with a database-style timestamp followed by a short description: tenant.202006042200.your_migration_name.sql. This ensures they show up in order and give you information about what the migration actually does. It's just one approach, but I think it works well.

chlupac:
And no naming conflict between team

PR #540, opened 2020-05-25
UserProfile View (Notifications View): Resolves issues with enabled/disabled input along with viewing or replying to notifications

thabaum:
fixes #539 

DO NOT MERGE THIS PR AS IT IS DIVIDED IN TWO.

https://github.com/oqtane/oqtane.framework/pull/585  UserProfile\Index
https://github.com/oqtane/oqtane.framework/pull/586  UserProfile\View

## **The ABOVE two PR's Issues fixed**

Index:

- Infinite RE: in reply messages
- Adds "..." to messages longer than 100 characters in the profile notifications view
- Removes carriage returns from message
- Removed reply message from user profile notifications tab messages

View:
- Reply Message Page Break using underscores
- Ability to enter cursor into reply area without having to hit enter to type message
- Enabled/Disabled input depending on view or reply
- Original Message shows up in text box that allows scrolling and line breaks.

## Screenshots

_Edited_
- Easier to put cursor at top of message text area with previous message adding line breaks to the body
image

- Fixes infinite subject RE: RE: RE:....
image

- Original message displayed in same format to display properly (although something else is bugging out the layout of the table possibly in the buttons)
image

- Removes fields you do not need to see when sending a reply
image
(date is not showing, but is added to the reply message.)

- Enables/Disables fields while viewing notifications and replying as necessary
image
image

image

This PR also now detects and removes the reply message from the User Profile Notifications message and adds a "..." to the end of the message if > 100

image

PR #585, opened 2020-06-05
'...' added, Reply Removed

### LINKED ISSUE #539: User Profile: Viewing Notification Messages Allows Editing
The fields in View Notification (messages) allow changing and editing,

thabaum:
fixes #539
Part 1 of 2 #586,  

## Summary
This PR is similar to #540 which should not be merged, instead use these two as they have broken down the changes into two PR.

## Issues Resolved
- Issue:  Removes Infinite RE: in reply messages
- Issue:  Removes reply from message

## Enhancements
- Enhancement:  Adds "..." to messages longer than 100
- Enhancement:  Removes carriage returns from messages

This PR is similar to #540 which should not be merged, instead use these two as they have broken down the changes into two PR.

## Concerns

-  Grouping messages that are replies with a sort feature "by conversation" so they don't all show up from a user .  A better UX to keep up with other users and things going on can help keep the amount of redundant information from being display using a way people are familiar with today.

## Screenshots
image

### DIFF
--- a/Oqtane.Client/Modules/Admin/UserProfile/Index.razor
+++ b/Oqtane.Client/Modules/Admin/UserProfile/Index.razor
@@ -126,7 +126,14 @@ else
                     </Row>
                     <Detail>
                         <td colspan="2"></td>
-                        <td colspan="3">@(context.Body.Length > 100 ? context.Body.Substring(0, 100) : context.Body)</td>
+                        <td colspan="3">
+                            @{ 
+                            string input = "---";
+                            if (context.Body.Contains(input)){ 
+                            context.Body = context.Body.Split(input)[0];
+                            } }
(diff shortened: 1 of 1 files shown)


thabaum:
Looks like carriage returns are showing up in the message.

Should I remove the carriage returns.  We want to keep messages to a single line while viewing all the index of them am I correct?

thabaum:
> Should I remove the carriage returns.

Great Idea! I think I will put out a commit in 10 seconds as issue resolved.

thabaum:
@sbwalker I will have this taking on the big dogs here pretty soon what you think for a start?

I will keep adding to it as time flows by if I can think of anything more while testing it further.  I just can go so many directions... I dont want to get carried away and just stick more to functionality and usability than redoing the UI.

I want to start making each dashboard component device friendly then revisit things.  I know you are hot on some other more major issues and putting out the next release so no pressure just a heads up...

PR #586, opened 2020-06-05
To/From Fields enabled/disabled, reply message fix

### LINKED ISSUE #539: User Profile: Viewing Notification Messages Allows Editing

thabaum:
fixes #539  (references PR #540 closed)

Part 2 of Part 1 #585

## Summary
Resolves issues with the user profile notifications area.

## Issues Resolved
- Issue:  Enabled/Disabled input depending on view or reply (To/From)  

This was my original fix... all others are fall out from working in this area which is why this has been broken into two

## Enhancements
- Enhancement:  Reply message page break line above '___' then date + subject + original conversation or message similar to email.
- Enhancement:  Message reply has empty line for cursor
- Enhancement:  Original Message shows up in text box that allows scrolling and line breaks.

## Concerns
-  To/From Form Table could be two if statements with two tables instead of many if statements inside columns, works as is, just not sure if I should approach that differently.  I can rewrite if desired.  I just followed what was already written.
- Original Message we may want a larger number of rows or not displayed in a text area however this put carriage returns into the message while viewing.  There are a number of ways this can be handled... I just put this as a solution to keep the screen from growing too far past where you are typing.

I tried everything I would use it for

## Screenshots
**View Notification**
image

image

Updated Original Message
image

unless you or others want me to go ahead try and make this more responsive along with adding CSS selectors...

calling this good for now unless changes are suggested.

Thank you @sbwalker

### DIFF
--- a/Oqtane.Client/Modules/Admin/UserProfile/View.razor
+++ b/Oqtane.Client/Modules/Admin/UserProfile/View.razor
@@ -173,7 +173,7 @@
             subject = "RE: " + subject;
         }
         reply = body;
-        body = "\n\n--------------------------------------------\nSent: " + createdon + "\nSubject: " + subject + "\n\n" + body;
+        body = "\n\n____________________________________________\nSent: " + createdon + "\nSubject: " + subject + "\n\n" + body;
         StateHasChanged();
     }
 
--- a/Oqtane.Client/Modules/Admin/UserProfile/View.razor
+++ b/Oqtane.Client/Modules/Admin/UserProfile/View.razor
@@ -91,16 +91,10 @@
     <br />
     @if (title == "To")
     {
-        <table class="table table-borderless">
-            <tr>
-                <td>
-                    <label class="control-label">Original Message: </label>
-                </td>
-                <td>
-                    <textarea class="form-control" @bind="@reply" rows="5" readonly />
-                </td>
-            </tr>
-        </table>
+        <div class="control-group">
+            <label class="control-label">Original Message </label>
+            <textarea class="form-control" @bind="@reply" rows="5" readonly />
+        </div>
         }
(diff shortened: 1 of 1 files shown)


thabaum:
I had this working great... now I just need more time to figure out why things changed so much...

It does not make sense hoping I can find my working branch, trying to resolve conflicts bugged it all up...  I need to find my working copy hopefully.

Cracks me up.  I will take a look at this again tomorrow sorry for the sloppy PR's...

thabaum:
https://github.com/oqtane/oqtane.framework/pull/585
combine these two PR's for the freshest new features ready for Oqtane User Profile Notifications @sbwalker they are ready.

The userid for selecting who to send a notification to changed causing the bugs... That and VS formats files differently now all the sudden if your razor blazor file does not follow good form it goes all over so I dont want to attempt these again.  Two files for two PR's that can have adjustments made if needed or go ahead and merge.

Enjoy.

thabaum:
These file changes compiled on latest repo, tested and approved on my end.

thabaum:
Only thing I have question about maybe someone can influence this commit.

I used --- dashes for the reply line break

I probably should use ___________

So I can fix this now real quick since it is not merged take the dash version if you like it as it will be one commit behind in a few minutes.

thabaum:
ok both PR's are now available in two flavors with dashes or underscores, latest is with underscores for the reply line breaks

sbwalker:
After reviewing this fix more over the weekend I have a question: Based on the fact that the Reply feature includes the entire thread of previous messages, I am wondering why there is a need for the Original Message field at the bottom of the form? It seems redundant since all of the same information is in the Message field already?

PR #603, opened 2020-06-09
Profile Managment: Index Username Class="Col"

thabaum:
uses 100% of screen pushing buttons together.

Resolves large gaps buttons...

image

Another way that works is adding the "table-responsive" class.

This does not make it mobile friendly it just moves cells over for now.

It looks like a few other areas could use this quick fix as well if ok such as 
image

If last two are accepted I will go ahead with the rest dashboard index files to help push the buttons and usernames together.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Profiles/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Profiles/Index.razor
@@ -19,7 +19,7 @@ else
         <Row>
             <td><ActionLink Action="Edit" Parameters="@($"id=" + context.ProfileId.ToString())" /></td>
             <td><ActionDialog Header="Delete Profile" Message="@("Are You Sure You Wish To Delete " + context.Name + "?")" Action="Delete" Class="btn btn-danger" OnClick="@(async () => await DeleteProfile(context.ProfileId))" /></td>
-            <td>@context.Name</td>
+            <td class="col">@context.Name</td>
         </Row>
     </Pager>
 }


sbwalker:
The buttons in the UI originally were left aligned and did not have a lot of padding. I believe PR #481 broke this behavior. That PR seemed to indicate it was attempting to "resolve issue with title border DIV not utilizing 100% pane size" but it had much broader consequences on the UI. Is it possible that #481 was not the appropriate fix?

sbwalker:
I verified that your comment above related to #481 is correct. However, the change did cause a visual effect on other areas of the UI - as the tables which were squashed expanded to 100% which caused the table columns to expand. The impact was mostly on the module UIs which were using the Pager component. The preferred behavior would be for the columns with buttons to be left aligned and not expand - whereas the columns with content should expand. If possible it should be accomplished using a standard Bootstrap approach so that no custom styles need to be introduced beyond what default Bootstrap utilizes itself.

thabaum:
Bootstrap is 100% what I am using

thabaum:
A fix is needed in the pager, and other areas.

In fact in the logger the pager is broken only works if you select other sizes.

I am not sure when these were created if they had Blazor in mind as it was pretty fresh still.

thabaum:
I did look at #481 and also looked at the one @chlupac had when he said he set admin container to false to get it to work is that related to this at all?  

`<div class="@_paneadminborder">`

to 

`<div class="container @_paneadminborder">`

Could the other class also need container ?

The way I thought in bootstrap is containers are holding things and you class them to select what maybe a form or table or whatever inside that.

I am not sure I would have to go back to testing this area again and get my mind into what was going on here.  I know I needed a container class if false to allow the header to use 100%

sbwalker:
I do not believe adding a class="col" to one of the table columns is the correct solution. The grid is a common pattern throughout the Admin UI and in most cases there are buttons on the left side of the grid and multiple other columns which contain content. If you add class="col" to one or all of the content columns it does not produce the desired result in most cases. Based on what I have seen in the past, if you want to ensure that the button columns are sized according to the button width then you would simply add a style="width: 1%;" or style="width: 1px;" to the TH tag for those columns. And you would not a style or class to the content columns so that they retain the default Bootstrap sizing behavior. Obviously you could define a CSS class rather than embedding a style, but since the plan is to replace these tables with divs in the future then the quick fix would be to just add the styles.

thabaum:
Would you like for me to attempt a PR for a component first using ```<th style="width: 1px;">&nbsp;</th>``` then maybe a larger PR with the rest of the components as needed for v1 dashboard took right for now or would you like to handle this?

sbwalker:
If you have time, please go ahead and replace #602 and #603 with a single PR which fixes the widths of the button columns in all of the grids in the Admin modules.

thabaum:
I will be able to spend some time here next week.

thabaum:
I will close these for now create a new PR for a continued conversation next week.  Possibly Monday.

sbwalker:
@thabaum should I log an issue for this item so that it is registered as an open item for the PR you are working on?

thabaum:
You can, I will put one up when I get the PR ready otherwise.

Issue #593, opened 2020-06-07
Question : Image caching.


chlupac:
I have page with images stored in Oqtane filestore. These images are downloaded again and again on every request, because of dynamic content. 
I thing we have to implement middleware or modify headers to convince browser than image is static and store to cache.

thabaum:
A cache on/off switch should be created for this so you can make the framework act either way you choose.  Potentially even a module option?  So when you download a component if it is info that will not change anytime soon it can be "static" although created dynamically in the back-end of Oqtane.

thabaum:
I always thought HTTPS/2 was suppose to take care of this type of stuff in the browser, we really need to dive into that part of Oqtane and test how we can achieve desired results for different situations potentially?

chlupac:
best place imho will be DownloadControler on request level with url query or path, if it is possible. 
i have no experiences with this, maybe i'm wrong.
now it is 
```
        // GET api/<controller>/download/5
        [HttpGet("download/{id}")]
        public IActionResult Download(int id)
        {
```
maybe we can add 
```
        // GET api/<controller>/static/5
        [HttpGet("static/{id}")]
        public IActionResult DownloadStatic(int id)
        {
```
and developer will choose which variant suits his needs

jimspillane:
Should files that require authorization be cached?

thabaum:
potentially not.  Maybe only content that is declared "public" ?

sbwalker:
This needs to be researched further as I am fairly sure that the default response caching middleware will not cache requests that contain cookies - and Oqtane is using cookies for auth currently.

https://docs.microsoft.com/en-us/aspnet/core/performance/caching/middleware?view=aspnetcore-3.1

chlupac:
I have caching implemented now. Together with image sizing/thumb-nailing engine in custom controller. It works wit oqtane filesystem.
```c#

        [ResponseCache(Duration = 1200)]
        [HttpGet("{id}/{size?}")]
        public async Task<IActionResult> Get(int id, string size)
        {
            var errorpath = Path.Combine(GetFolderPath("images"), "error.png");
            var file = _files.GetFile(id);
            if (file != null)
            {
                if (_userPermissions.IsAuthorized(User, PermissionNames.View, file.Folder.Permissions))
                {
                    var filepath = Path.Combine(GetFolderPath(file.Folder), file.Name);
                    if (System.IO.File.Exists(filepath))
                    {
[...]

PR #617, opened 2020-06-15
Image caching & thumbnails, file/download optimalization

chlupac:
/file/image
+ Ability to create thumbnails from images
+ Caching enabled 

/file/download
+ add mime types to downloaded files
+ download controller optimization
+ use PhysicalFile() method instead of reading whole file to memory.

TODO: because of recent changes in base classes I did not add support methods to them. 
We have to decide, where and what to add.

### DIFF
--- a/Oqtane.Server/Controllers/FileController.cs
+++ b/Oqtane.Server/Controllers/FileController.cs
@@ -11,12 +11,12 @@ using System.Threading;
 using System.Threading.Tasks;
 using Oqtane.Security;
 using System.Linq;
-using System.Drawing;
 using System.Net;
 using Oqtane.Enums;
 using Oqtane.Infrastructure;
 using Oqtane.Repository;
-using Microsoft.AspNetCore.Routing.Constraints;
+using Oqtane.Extensions;
+using static Oqtane.Infrastructure.ImageProcessing.ThumbnailImage;
 
 // ReSharper disable StringIndexOfIsCultureSpecific.1
 
@@ -65,7 +65,7 @@ namespace Oqtane.Controllers
                     {
                         foreach (string file in Directory.GetFiles(folder))
                         {
-                            files.Add(new Models.File { Name = Path.GetFileName(file), Extension = Path.GetExtension(file)?.Replace(".", "") });
+                            files.Add(new Models.File {Name = Path.GetFileName(file), Extension = Path.GetExtension(file)?.Replace(".", "")});
                         }
                     }
                 }
@@ -208,7 +208,7 @@ namespace Oqtane.Controllers
             {
                 _logger.Log(LogLevel.Error, this, LogFunction.Create,
                     "File Could Not Be Downloaded From Url Due To Its File Extension {Url}", url);
-                HttpContext.Response.StatusCode = (int)HttpStatusCode.Conflict;
+                HttpContext.Response.StatusCode = (int) HttpStatusCode.Conflict;
                 return file;
             }
 
@@ -216,7 +216,7 @@ namespace Oqtane.Controllers
             {
                 _logger.Log(LogLevel.Error, this, LogFunction.Create,
                     $"File Could Not Be Downloaded From Url Due To Its File Name Not Allowed {url}");
-                HttpContext.Response.StatusCode = (int)HttpStatusCode.Conflict;
+                HttpContext.Response.StatusCode = (int) HttpStatusCode.Conflict;
                 return file;
             }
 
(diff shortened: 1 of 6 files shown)


sbwalker:
ContentUrl() overloads should be added to ModuleBase and ThemeBase ( LayoutBase, ThemeControlBase, ContainerBase all inherit from ThemeBase now so that methods are not duplicated )

sbwalker:
I have some concern about this new method:

        [ResponseCache(Duration = 1200)]
        [HttpGet("image/{id}/{size?}/{extension?}/{quality?}")]
        public IActionResult Image(int id, string size = "full", string extension = "jpg", long quality = 90L)

It is a GET method which is generally only used for retrieval and should not change the state of a resource. However in this case it actually results in the creation of a file. Internally it does have some security to ensure the caller has appropriate permissions for the folder - however it is checking if the user has View permissions rather than Edit permissions. It is common to have many folders in an Oqtane installation which allow All Users to have View permissions - so this does not provide much security. And when the thumbnail is created it is not registered in the File table in the database - so there is no way for the site administrator to be aware of or manage these files through the UI. Lastly, it would be trivial to create a client application which sends automated GET requests to the server for a specific fileid ( obtained by viewing the source on any page ) and passing a Size parameter of 1 to 1000000 and also alternating the Extension. Essentially this would quickly fill the file system with image files which are invisible to the site administrator ( until the site crashed due to disk space constraints ).

sbwalker:
I apologize that I did not noticee these vulnerabilities in the code you posted in the original issue #593 - I was more focused on the image caching aspect.

chlupac:
You are right. It can be a problem. Thumbs can be easily deleted, but system can be flooded  in seconds. Do not merge it. I have to rethink thumb generation strategy. Maybe create thumb with some permissions only not for public. Edit rights will be ok. Or during upload. Or save them only to cache.  /content/temp ?? folder can be safely deleted. 
But speed-up and bandwidth save is great. 
Too many possibilities.

sbwalker:
I think at the very least, thumbnail generation should only be available to users who have Edit permissions for the folder where the image resides. And perhaps if the user does not have Edit permissions, the API will simply return the full size image ( cached )? 

However relying on Edit permissions is not enough... in Oqtane all registered users get their own "personal" folder for managing their own files ( ie. profile picture, etc... ). So registered users have Edit rights to at least one folder in the system. If they uploaded a photo and then got the fileid for that photo they could call the API and flood the server. So I think the API should strict the size to a whitelist of acceptable dimensions. This could still be abused but at least it is much more restricted.

And any files that are created should be stored in the File table and and an event log entry should be created.

chlupac:
Thumbnails can be created on demand, but it is cpu-intensive action. 
I'm thinking about implementing caching mechanism with limited cache items per file. Cache will be stored independently of content folder and will have limited size. 

>And perhaps if the user does not have Edit permissions, the API will simply return the full size image ( cached )?

If thumb  is generated, will be returned. if not, the API will simply return the full size image ( cached ).

> So I think the API should strict the size to a whitelist of acceptable dimensions. 

We can define small, medium, large, full sizes and limit size of thumbnail to them. `/file/image/55/small`
It is about images for blogs, photo galeries and similar modules. Due to responsible design we do not need to be pixel-perfect.    

>And any files that are created should be stored in the File table and and an event log entry should be created.

It is cache, I do not like to flood log and file table.  It is my design mistake to put them to main file storage. They should be stored separately

sbwalker:
The original issue #593 was related to caching and not related to thumbnail generation - these are totally different concerns which have been combined together in this PR. Since there are no security concerns related to caching, does it make sense to split this PR?

In regards to thumbnails. I am not in favor of the creation of unmanaged caches. Resources that are "invisible" to the site administrator and have no form of access control almost always result in maintenance/support/vulnerability issues in the long term.

sbwalker:
"Thumbnails can be created on demand, but it is cpu-intensive action."

Whether the system is consuming disk resources by creating a physical file, or consuming CPU resources by generating a thumbnail on demand, or consuming RAM by storing items in memory, the problem is the same - we need to ensure the system is protected from excessive resource consumption.

sbwalker:
One feature which DNN contained - which I expect Oqtane will need as well - is the ability to assign a maximum size to a Folder. Then the system is able to restrict the the creation of files based on size and mitigate the risk of excessive consumption.

chlupac:
It needs more research, closing this topic for now.
But .. "I'll Be Back" (Terminator)

PR #625, opened 2020-06-18
Send notification based on a future datetime

mikecasas:
This PR adds the ability to send a notification based on a future datetime. The logic uses a new field in the Notification table named SendOn. To keep the current functionality with no breaking changes, the SendOn date is defaulted to ```UtcNow``` in the framework. 

In the Notificaiton repository class, an additional WHERE cause is added to ```GetNotifications``` so the List/IEnumerable of Notifications only includes notifications with a SendOn date before now.

A developer would use this by creating their own Notification object and adding a send on date that meets their needs.

### DIFF
--- a/Oqtane.Client/Modules/Admin/UserProfile/Add.razor
+++ b/Oqtane.Client/Modules/Admin/UserProfile/Add.razor
@@ -66,6 +66,7 @@
                 notification.CreatedOn = DateTime.UtcNow;
                 notification.IsDelivered = false;
                 notification.DeliveredOn = null;
+                notification.SendOn = DateTime.UtcNow;
                 notification = await NotificationService.AddNotificationAsync(notification);
                 await logger.LogInformation("Notification Created {Notification}", notification);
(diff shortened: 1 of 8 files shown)


sbwalker:
Framework upgrade scripts need to follow a specific naming convention - Master.##.##.##.##.sql for Master DB scripts and Tenant.##.##.##.##.sql for Tenant DB scripts. The initial ##.##.## correspond to the framework version they belong to, and the last ## is a sequential patch to maintain ordering. The upgrade script for the Notification table change should be Tenant.01.00.01.01.sql.

sbwalker:
should the update script also contain an initialization command to set the SendOn field to the CreatedOn date ( so that any existing records have valid values ) ?

mikecasas:
I thought about the update, but I figured any existing notifications would have already been sent. I can add it though. I will update the script name too.

thabaum:
@mikecasas thank you for holding off on this one till my  PR went through.  I appreciate it.

PR #628, opened 2020-06-19
Fix Dashboard button/input table col size

### LINKED ISSUE #601: Bug:  Unable to update user profile "Passwords Entered Do Not Match"
## ISSUE

Password fields if you have saved for site will auto fill form in Identity tab causing passwords do not match.  This forces a user to first go back to Identity tab, clear the fields to save profile changes.

## STEPS TO REPRODUCE

0.  Have a user password saved in browser for the Oqtane website.
1.  Click on gear
2.  Click on Admin Dashboard
3.  Click on User Management
4.  Add User and save
5.  Click Edit User Button
6.  Notice Password field is autocompleted  (Confirm Password is not)
7.  Switch to Profile and make a change
8.  Click save and get error that passwords do not match.

## SCREENSHOTS

image

image

### LINKED ISSUE #627: Issue:  dasbhoard buttons are not close to table fields
## Issue

The dashboard buttons are currently stretched across the screen and should be closer together

Related discussion:
#603

thabaum:
## Summary
Fixes #627 
and Fixes #601 

- Moves buttons closer together as shown in screenshots below.
- autocomplete removed from profile password while viewing profile to prevent it from being filled while saving

Note:  I also edited permissions table to put the inputs closer together.   Putting a `table-responsive` class on the table will bring all of them close to the names/roles if desired.  It just does not stretch out with the user input.

I also edited the user input and grouped the items together so they are one as shown in the screenshot below for permissions.

## Screenshots
image

Permission Grid Current Results in this PR:
image
## ALTERNATIVES

Adding Table-Responsive Class to Permission Grid Table
image

Personally I like adding the table responsive class to it however it does not use 100% of space with the user input area so not sure which to go with.  Having them floating more to the left when a user is added the grids for user and roles are not lined up.  I felt possibly adding some width to that table column header the same size as the Registered Users label field takes up plus a little extra which would resolve this look.  And possibly adding a page line break to make UX clearer so that the User permission grid is more clearly visible.

Third option of coarse is to keep it as is.

I would like to at least keep the user input area grouping with an input-group class added to the row that includes the user input.

One last note on the spacing of the permission grid input check boxes is potentially making a class for it and wrap in a div `app-permission-grid` along with the `app-table-colfit` to address these.  This can and I think should be done more when we focus on the responsiveness of the dashboard as well.  If you like the PR as is you can merge it now.  I am sorry the PR had a few mistakes in handling it's issues from things like uploading the configuration file... I can redo as desired.


thabaum:
I went ahead put two issues together since they relate to the same files.

I am on the fence with the permissionsgrid component.  I removed the space between the inputs so they are closer together which I like however I also like them closer to the name or role name.  Something that can get discussed more later I can remove it and leave it as it was.  

If you like the Alternate way with the input closer to the name I can put that table-responsive class into the table to give it that effect.  I just feel there needs to be a minimum width so the first row when users are selected line up the first View Edit... permissions together like when they are aligned to the right.

PR #629, opened 2020-06-19
ContainerSettings

chlupac:
Container settings
* Add possibility to container have own settings (background, frame, shadow, show header .....), allows one universal container, do not need extra container type for combination of values
* Add ISettingPage interface for content of Tabpage with settings
* Container must implement  ISettingProvider interface (defines type of setting page)
Settings generally
* Added extensions to store complex model to settings in serialized JSON format.

Existing code without changes, only additions

### DIFF
--- a/Oqtane.Client/Modules/Admin/Modules/Settings.razor
+++ b/Oqtane.Client/Modules/Admin/Modules/Settings.razor
@@ -80,6 +80,14 @@
             @DynamicComponent
         </TabPanel>
     }
+
+    @if (ContainerSettingsComponent != null)
+    {
+        <TabPanel Name="ContainerSettings" Heading="Container Settings">
+            @ContainerSettingsComponent
+        </TabPanel>
+    }
+
 </TabStrip>
 <button type="button" class="btn btn-success" @onclick="SaveModule">Save</button>
 <NavLink class="btn btn-secondary" href="@NavigateUrl()">Cancel</NavLink>
@@ -95,12 +103,15 @@
     private PermissionGrid _permissionGrid;
     private Type _settingsModuleType;
     private string _settingstitle = "Other Settings";
+    
     private object _settings;
+    private ISettingPage _containerSettings;
 
     public override SecurityAccessLevel SecurityAccessLevel => SecurityAccessLevel.Edit;
     public override string Title => "Module Settings";
 
     private RenderFragment DynamicComponent { get; set; }
+    private RenderFragment ContainerSettingsComponent { get; set; }
 
     protected override async Task OnInitializedAsync()
     {
@@ -137,6 +148,22 @@
                 builder.CloseComponent();
             };
         }
+
+        var containerType = Type.GetType(ModuleState.ContainerType);
+        if (containerType != null && typeof(ISettingPageProvider).IsAssignableFrom(containerType))
+        {
+            var moduleobject = Activator.CreateInstance(containerType) as ISettingPageProvider;
+            var type = moduleobject?.SettingsPage;
+            if (type != null)
+            {
+                ContainerSettingsComponent = builder =>
+                {
+                    builder.OpenComponent(0, type);
+                    builder.AddComponentReferenceCapture(1, inst => { _containerSettings = Convert.ChangeType(inst, type) as ISettingPage; });
+                    builder.CloseComponent();
+                };
+            }
+        }
     }
 
     private async Task SaveModule()
@@ -161,14 +188,24 @@
         module.Permissions = _permissionGrid.GetPermissions();
         await ModuleService.UpdateModuleAsync(module);
 
-        if (_settingsModuleType != null)
+        if (_settings is ISettingPage page)
         {
-            var moduleType = Type.GetType(ModuleState.ModuleType);
-            if (moduleType != null)
+            await page.UpdateSettings();
+        }
+        else
+        {
+            // For compatibility with settings where ISettingPage is not implemented
+            if (_settingsModuleType != null)
             {
-                moduleType.GetMethod("UpdateSettings")?.Invoke(_settings, null); // method must be public in settings component
+                var moduleType = Type.GetType(ModuleState.ModuleType);
+                if (moduleType != null)
+                {
+                    moduleType.GetMethod("UpdateSettings")?.Invoke(_settings, null); // method must be public in settings component
+                }
             }
         }
+
+        _containerSettings?.UpdateSettings();
 
         NavigationManager.NavigateTo(NavigateUrl());
     }
--- a/Oqtane.Client/Modules/ISettingPage.cs
+++ b/Oqtane.Client/Modules/ISettingPage.cs
@@ -1,0 +1,9 @@
+﻿using System.Threading.Tasks;
+
+namespace Oqtane.Modules
+{
+    public interface ISettingPage
+    {
+        Task UpdateSettings();
+    }
+}
--- a/Oqtane.Client/Modules/ISettingPageProvider.cs
+++ b/Oqtane.Client/Modules/ISettingPageProvider.cs
@@ -1,0 +1,9 @@
+﻿using System;
+
+namespace Oqtane.Modules
+{
+    public interface ISettingPageProvider
+    {
+        Type SettingsPage { get; set; }
+    }
+}
--- a/Oqtane.Shared/Shared/Utilities.cs
+++ b/Oqtane.Shared/Shared/Utilities.cs
@@ -5,6 +5,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Text.Json;
 using System.Text.RegularExpressions;
 using File = Oqtane.Models.File;
 
@@ -331,6 +332,41 @@ namespace Oqtane.Shared
             }
 
             return dictionary;
+        }
+        
+        public static T GetSetting<T>(this Dictionary<string, string> settings, string settingName, Func<T> defaultValue)
+        {
+            try
+            {
+                string value = null;
+                if (settings != null && settings.ContainsKey(settingName))
+                {
+                    value = settings[settingName];
+                }
+                return string.IsNullOrEmpty(value)  ? defaultValue() : JsonSerializer.Deserialize<T>(value);
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine(e);
+            }
+            return defaultValue();
+        }
+
+        public static Dictionary<string, string> SetSetting<T>(this Dictionary<string, string> settings, string settingName, T value)
+        {
+            if (settings == null)
+            {
+                settings = new Dictionary<string, string>();
+            }
+            if (settings.ContainsKey(settingName))
+            { 
+                settings[settingName] =  JsonSerializer.Serialize(value);
+            }
+            else
+            {
+                settings.Add(settingName, JsonSerializer.Serialize(value));
+            }
+            return settings;
         }
     }
 }


sbwalker:
I am having a difficult time processing this PR as although I think there is a practical purpose behind it, I don't think the way it was implemented aligns with the architecture. I will do my best to try and explain...

Themes and Containers are intended to be for design purposes only. They are global in nature and are purposely not tied to any stateful concept such as Site, Page, or Module. A Theme can be associated to a Site or a Page to provide a visual design, and a Container can be associated to a Module to provide it with a visual design - but any of the metadata which relates to the behavior of Sites, Pages, and Modules must be associated to those stateful entities - not to the Themes or Containers themselves. For example, if you want a setting to be able to control whether a module title is displayed in a Container - it should be implemented in the Module Settings. 

Custom settings are a common use case that most modules require. The way this is currently implemented in Oqtane is that a module simply needs to include a razor component named Settings.razor. You can see an example for the example blog module here: https://github.com/oqtane/oqtane.blogs/blob/master/Client/Settings.razor . If a module provides this component, when a user uses the action menu and chooses Manage Settings for the module, it will display a tab for Other Settings and it will render the UI defined in the component. 

image

chlupac:
Yes, and my solution adds another tab with ContainerSettings. which belongs to container. I know than container is theme-related but in custom solution with custom theme is theme integral part of solution. 
I have three options.
1. Either have a pile of containers that differ in little things. I now have 6 types of containers and need other combinations of the same properties. And to explain when to use which container is more difficult with the growing number of containers.
2. One empty container and module-controlled design. However, this solution cannot be applied to existing modules.
3. Ability to specify container parameters for rendering (this PR)
image

chlupac:
Usage: 
1. in container implement ISettingPageProvider
```c#
    public abstract class MyContainerBase : ContainerBase, ISettingPageProvider
    {
        [Inject] public ScopeState ScopeState { get; set; }
        public virtual Type SettingsPage { get; set; }
    }
    public partial class Container : MyContainerBase
    {
        public override Type SettingsPage => typeof(ContainerSettings);
    }
```
2. implement `ContainerSettings` which uses `ISettingPage` interface. This interface should be also used on all settings pages for     `public async Task UpdateSettings()` declaration
```c#
@inherits ModuleBase
@implements ISettingPage
@inject ISettingService SettingService
                
<h1>Hello</h1>
@if(Model!=null){

    <div class="row">
        <div class="col">
            <div class="form-group">
                <SfCheckBox Label="Show Header:" @bind-Checked="@Model.ShowHeader"></SfCheckBox>
            </div>
            <div class="form-group">
                <SfCheckBox Label="Show Shadow:" @bind-Checked="@Model.Shadow"></SfCheckBox>
            </div>
            <div class="form-group">
                <SfCheckBox Label="Public only:" @bind-Checked="@Model.PublicOnly"></SfCheckBox>
            </div>

        </div>
    </div>
}
@code {

    public ContainerSettingsModel Model { get; set; }

    {
        try
        {
            var settings = await SettingService.GetModuleSettingsAsync(ModuleState.ModuleId);
            Model = settings.GetSetting("Container", () => new ContainerSettingsModel());
        }
        catch (Exception ex)
        {
            ModuleInstance.AddModuleMessage(ex.Message, MessageType.Error);
        }
    }

    public async Task UpdateSettings()
    {
        try
        {
            var settings = await SettingService.GetModuleSettingsAsync(ModuleState.ModuleId);
            settings.SetSetting("Container", Model);
            await SettingService.UpdateModuleSettingsAsync(settings, ModuleState.ModuleId);
        }
        catch (Exception ex)
        {
            ModuleInstance.AddModuleMessage(ex.Message, MessageType.Error);
        }
    }
}
```
3. settings of container belongs to Module:
image

chlupac:
In Admin/Modules/Settings.razor I did these changes:
1. Add page Container settings to interface
image
2. Add component to page
image
3. Handling save (It's easier, thanks to ISettingPage interface)
image
4. Use ISettingPage instead of reflection on "Other Settings" if interface is present. All settings pages can benefit from this interface.
image

sbwalker:
Thank you for the explanation... it helps me understand what you are trying to accomplish. I had to think about this use case quite a bit, as it was not something that was ever requested in the 15-20 history of DNN ( and Oqtane follows a very similar conceptual approach to DNN ).

Basically you would like to have a number of different variations of a container, but you do not want to create a separate component for each variation - partly because you don't want to maintain multiple components and partly because the user experience for selecting the appropriate container can be a bit challenging for end users. This seems like a legitimate use case, although it is not the only use case - the system still needs to provide the freedom to support multiple containers as well as this is a proven pattern which many developers/designers will embrace.

My first approach was to validate that it was not possible to achieve your goals with the current architecture. You could certainly create a custom container where its behavior is controlled by configuration. And you could create a Settings component and include your configuration options in that component. However this would only work if there was a tight coupling between a module and a container - and that is not a good idea, you want them to be decoupled. Another option would be to create a global settings UI where the configuration for a container could be managed - but that is not viable either as you need the ability to customize the configuration on a module by module basis. So I determined that it is not possible to achieve your goals with the current architecture.

So then I focused on the technical solution. I understand how the solution in the PR works but I have a number of comments/questions...

1. One of the central concepts in Oqtane is simplicity - keeping things very simple for developers. The idea with Modules and Themes is that a developer simply needs to inherit from a base class and it provides all of the fundamentals that a developer needs. When it comes to Containers there is a ContainerBase class which all containers must inherit from and it implements IContainerControl to ensure backward compatibility. It would seem to make sense that rather than introducing another interface which seems redundant and increases the overall complexity to simply add another property to the IContainerControl and include a default initializer in ContainerBase. 

2. The term "Page" has a very distinct meaning in Oqtane - it has a route and represents a grouping of modules. Therefore using the term "Page" as part of SettingPage is not appropriate as it creates confusion within the framework. The SettingPage is not a "page" - it is a type which refers to a component. Oqtane refers to components as "Controls" so most likely the name which is most consistent with the architecture is SettingControlType.

3. The model is not really a technical requirement at all - its an implementation preference. It could be accomplished just as easily with a Settings dictionary. The ultimate solution should support both approaches rather than forcing a developer to have to create a model.

4. In general I prefer to avoid storing serialized objects in table fields. There are many reasons for this ( which I unfortunately learned the hard way in the past ): 1) they are difficult to query to locate a specific property value, 2) they are difficult to manage when models change ( ie. when properties are added or removed from models ), 3) they are difficult to upgrade via scripts or migrations. Its fine if third party developers want to store serialized objects in their own tables and take responsibility for them, but I do not want the core framework to use this approach.

sbwalker:
5. In the Module Settings UI when a user selects a container from the dropdown it would make sense that the container settings would become visible ( if the container had settings ) or became invisible ( if the container did not have settings ). Perhaps it would be more intuitive if the container settings were displayed directly beneath the Container dropdown rather than in their own tab?

6. The ability to specify a default container at the site or page level is a very common scenario. Would container settings be applicable at these levels as well - or only at the module level? If a default container was specified at the site level but the administrator did not like the default value set for a specific container setting, they would need to visit every module on the site in order to override that container setting. This would be very cumbersome. 

7. If we were going to support container settings wouldn’t it also make sense to support theme settings in a consistent manner? Then you could have a single theme with behavior controlled by configuration.

sbwalker:
8. If the hierarchy of settings at the Site, Page, and Module need to be supported, it gets tricky due to inheritance. If a container has a setting that is set at the Site level it should be inherited at the Page and Module level. However it could be overridden at those levels as well. Resolving the setting value based on the hierarchical entity is challenging. 

9. The current implementation of containers is light-weight - containers do not require any other state. If someone chose to use this new approach, a container would need to load it’s settings from the database which will create a lot more network/DB calls - which will have a negative impact on system performance.

chlupac:
Great analysis. It made me look at the problem more globally. We have a number of entities in the system. Some are represented by the model in the database, some only in the code. Some currently allow you to edit and save user metadata in settings, but most do not. It is therefore a matter of deciding whether to implement this in a uniform procedure for entities where the possibility of entering user settings is not implemented.
At the Site and Page level they would be very useful, especially for setting different constant values. Similarly at the Theme and Container level, especially for simple appearance modifications.
But you are right that there is power in simplicity and a balance needs to be struck between usefulness and complexity.

For individual points:
1. If we implement settings only for Container, it is definitely better to extend IContainerConrol. But if a similar mechanism is used in general, an extra interface is needed.
2. Yes, the names I used were not chosen very well, they need to be renamed
3. The solution in no way forces the user to use the model and allows both approaches to be used.
4. Serialized objects. The solution does not force them to use it, it implements two procedures that allow it. They are not needed to solve the problem. SQL2016 + allows work with JSON data serialized in nvarchar column.
If you don't want these two procedures, I'll drop them.

chlupac:
5. It's a matter of design, for the simplicity of the thing I chose an extra tab page and didn't come up with an extra solution.
6. A simple thing brought to perfection becomes very complex. But I'll think about how to easily implement it. It will probably be best that there will be a "save as default" button on the settings form.
7. In the preface.
8. The implementation of this is very complex and would require a change of work with the setting values.
9. I agree. Because the setting value is tied to the module, there is an option to load the complete module and container settings at the time of loading PageModule and have the settings already stored in ModuleState and not reload them from the database when the module is rendered.
10. It is based on the fact that a module is a group of classes that associate a common namespace. This is not the case with the container. That's why I introduced the ISettingPage interface.

sbwalker:
Considering all the impacts of this enhancement discussed above, the main question before going any further is if its truly worth it - especially considering that there is support for variations in containers in the framework already. Certainly the architecture could be adapted to support this specific use case - technology is rarely the blocker - the real question is if this use case is so important that it should force the framework architecture to adapt. This is where product management is challenging as there are all always tons of potential features, but every feature you add incurs complexity and long term support and maintenance. It is also important to consider if there are any other ways to solve the problem. 

For example, the user experience for choosing a container needs to be improved regardless of whether or not this enhancement is integrated. Themes and containers have thumbnail properties in their interfaces which are intended to allow the designer to include an example screen shot. This would allow the application to provide the end user with a visual aid when they choose a theme or container - it just has not been fully implemented yet. 

I have some some follow on questions related to your previous responses above, but before I ask them I need to sort out the question on whether or not the framework needs this feature.

chlupac:
a few lines of code and such a great discussion :)
But let's not look at it as a solution to my partial container problem. It is a matter of deciding whether and how to create a general mechanism that will allow the developer to easily configure the UI for metadata (settings) of individual Oqtane entities. Now this is possible for modules, but the question is whether to extend it with a general implementation.

sbwalker:
@chlupac I have not forgotten about this PR... I am still considering whether it is an important enhancement at this stage. It was a very useful discussion as it brought up a lot of good points. There are definitely a few items I would like to explore further - even if this PR is not integrated in its entirety. 

To provide a bit more high level background related to this thread, the general expectation is that if a specific module has settings it should implement them within its own module settings UI. This is true regardless of whether these settings are internally utilized by the module at the tenant, site, page, module, or pagemodule level. For example if you create a Blog module and it has a variety of settings that it utilizes, then the administrator who is maintaining the blog on their site should be able to go to a single location within the module to manage all of the Blog settings. This is important from a user experience perspective because you do not want to force the user to navigate to various settings screens within the framework to find Blog settings. In addition, the administrator of the Blog module may not even have permissions to access the Page settings or Site settings screens.

thabaum:
I would like to spend some time going over things here as well.  I am trying to process things as they relate to Oqtane and ability to add container parameters.

But my question here is shouldn't the "Container Settings" be something that is configured by the container and maybe even off to the right a gear for the container itself to make adjustments to it?  The module and container should be in separate areas am I correct?  Maybe above the first module in a container there could be a set of cogs or cog for container settings?

Issue #23, opened 2019-05-25
3rd Party Control Support


bbakermmc:
One of the problems with a lot of frameworks like these is they are built w/out any 3rd party controls like DevExpress, Telerik, SynFusion, etc.  So then when you start programming, some of the UI elements tend to look "off" because maybe some of the pages that are defied in the framework havent been updated to use the other control sets.

I havent looked at the framework fully yet, but it would be nice if you implemented helpers to render out controls that way they could be overridden easily if we wanted to use DevExpress editors and such.

Or is that like a "theme" instead?

sbwalker:
You have to walk a fine line when it comes to application frameworks. On the one hand you should not dictate a specific commercial control suite - you should allow developers to use their preferred control suite ( and in fact you generally cannot distribute an open source framework with a commercial control suite anyways due to licensing restrictions ). On the other side, you need to provide at least a set of basic styles and basic controls so that the framework has cohesive look and feel.

bbakermmc:
@sbwalker Correct, its always the issue :).  The base framework components usually look like bootstrap, in most cases they are tied to admin type pages so its not too bad, but then the other pages have these nice styled controls, it would be nice to just implement an override on a tag helper or something.  Its something to think about since a lot of use use the 3rd party grids and stuff, we just want things to be uniform if possible and not have to re code all the pages you have already done.

leedavi:
On a related, but slightly different angle.  This CSS framework is nice, light and easy.

https://www.w3schools.com/w3css/

It's basically a light weight Bootstrap, I've started to use it on most admin UI.

iJungleboy:
From my DNN experience, I would not include _any_ css by default. Better put the admin-UI stuff in Iframes. The pain on the designers is just too great otherwise.

bbakermmc:
@sbwalker Thanks for making Razor Components for things like the PermissionGrid, etc.  This will help with wanting to use controls like DevExpress.

PR #638, opened 2020-06-27
Module Router Enhancement

PoisnFang:
Allows for PageVariables through the URL. This should not change the current functionality of module routing, it just allows for you to get variables through the URL.

Example: 
```
public override string Actions { get { return "AddBlog,
EditBlog/[blogId],
EditBlog/[blogId]/EditChildBlog/[childBlogId]]"; } }

### DIFF
--- a/Oqtane.Server/appsettings.json
+++ b/Oqtane.Server/appsettings.json
@@ -1,6 +1,6 @@
 {
   "ConnectionStrings": {
-    "DefaultConnection": ""
+    "DefaultConnection": "Data Source=(LocalDb)\\MSSQLLocalDB;AttachDbFilename=|DataDirectory|\\Oqtane.mdf;Initial Catalog=Oqtane;Integrated Security=SSPI;"
   },
   "Runtime": "Server",
   "Installation": {
--- a/Oqtane.Server/appsettings.json
+++ b/Oqtane.Server/appsettings.json
@@ -1,6 +1,6 @@
 {
   "ConnectionStrings": {
-    "DefaultConnection": "Data Source=(LocalDb)\\MSSQLLocalDB;AttachDbFilename=|DataDirectory|\\Oqtane.mdf;Initial Catalog=Oqtane;Integrated Security=SSPI;"
+    "DefaultConnection": ""
   },
   "Runtime": "Server",
   "Installation": {
--- a/Oqtane.Client/Modules/ModuleBase.cs
+++ b/Oqtane.Client/Modules/ModuleBase.cs
@@ -132,10 +132,10 @@ namespace Oqtane.Modules
                         if (templateSegments[i] == parameters[i])
                         {
                         }
-                        else if (templateSegments[i].StartsWith("[") && templateSegments[i].EndsWith("]"))
+                        else if (templateSegments[i].StartsWith("{") && templateSegments[i].EndsWith("}"))
                         {
-                            var key = templateSegments[i].Replace("[", "");
-                            key = key.Replace("]", "");
+                            var key = templateSegments[i].Replace("{", "");
+                            key = key.Replace("}", "");
                             urlParameters.TryAdd(key, parameters[i]);
                         }
                         else
--- a/Oqtane.Client/Modules/ModuleBase.cs
+++ b/Oqtane.Client/Modules/ModuleBase.cs
@@ -17,7 +17,7 @@ namespace Oqtane.Modules
         private Logger _logger;
 
         protected Logger logger => _logger ?? (_logger = new Logger(this));
- 
+
         [Inject]
         protected ILogService LoggingService { get; set; }
 
@@ -30,7 +30,7 @@ namespace Oqtane.Modules
         [CascadingParameter]
         protected Module ModuleState { get; set; }
 
-        [CascadingParameter] 
+        [CascadingParameter]
         protected ModuleInstance ModuleInstance { get; set; }
 
         // optional interface properties
@@ -62,7 +62,7 @@ namespace Oqtane.Modules
                 }
             }
         }
-        
+
         // path method
 
         public string ModulePath()
@@ -116,6 +116,38 @@ namespace Oqtane.Modules
             return Utilities.ContentUrl(PageState.Alias, fileid);
         }
 
+        public Dictionary<string, string> GetUrlParameters(string parameterTemplate)
+        {
+            var urlParameters = new Dictionary<string, string>();
+
+            var templateSegments = parameterTemplate.Split('/', StringSplitOptions.RemoveEmptyEntries);
+            var parameters = PageState.UrlParameters.Split('/', StringSplitOptions.RemoveEmptyEntries);
+
+            if (parameters.Length == templateSegments.Length)
+            {
+                for (int i = 0; i < parameters.Length; i++)
+                {
+                    if (parameters.Length > i)
+                    {
+                        if (templateSegments[i] == parameters[i])
+                        {
+                        }
+                        else if (templateSegments[i].StartsWith("[") && templateSegments[i].EndsWith("]"))
+                        {
+                            var key = templateSegments[i].Replace("[", "");
+                            key = key.Replace("]", "");
+                            urlParameters.TryAdd(key, parameters[i]);
+                        }
+                        else
+                        {
+                            i = parameters.Length;
+                        }
+                    }
+                }
+            }
+            return urlParameters;
+        }
+
         // user feedback methods
         public void AddModuleMessage(string message, MessageType type)
         {
@@ -154,12 +186,15 @@ namespace Oqtane.Modules
                 case "add":
                     logFunction = LogFunction.Create;
                     break;
+
                 case "edit":
                     logFunction = LogFunction.Update;
                     break;
+
                 case "delete":
                     logFunction = LogFunction.Delete;
                     break;
+
                 default:
                     logFunction = LogFunction.Read;
                     break;
--- a/Oqtane.Client/UI/PageState.cs
+++ b/Oqtane.Client/UI/PageState.cs
@@ -14,6 +14,7 @@ namespace Oqtane.UI
         public List<Module> Modules { get; set; }
         public Uri Uri { get; set; }
         public Dictionary<string, string> QueryString { get; set; }
+        public Dictionary<string, string> PageVariables { get; set; }
         public int ModuleId { get; set; }
         public string Action { get; set; }
         public bool EditMode { get; set; }
--- a/Oqtane.Client/UI/PageState.cs
+++ b/Oqtane.Client/UI/PageState.cs
@@ -15,7 +15,6 @@ namespace Oqtane.UI
         public Uri Uri { get; set; }
         public Dictionary<string, string> QueryString { get; set; }
         public string UrlParameters { get; set; }
-        public string Anchor { get; set; }
         public int ModuleId { get; set; }
         public string Action { get; set; }
         public bool EditMode { get; set; }
--- a/Oqtane.Client/UI/PageState.cs
+++ b/Oqtane.Client/UI/PageState.cs
@@ -14,7 +14,8 @@ namespace Oqtane.UI
         public List<Module> Modules { get; set; }
         public Uri Uri { get; set; }
         public Dictionary<string, string> QueryString { get; set; }
-        public Dictionary<string, string> PageVariables { get; set; }
+        public string UrlParameters { get; set; }
+        public string Anchor { get; set; }
         public int ModuleId { get; set; }
         public string Action { get; set; }
         public bool EditMode { get; set; }
--- a/Oqtane.Client/UI/SiteRouter.razor
+++ b/Oqtane.Client/UI/SiteRouter.razor
@@ -78,7 +78,6 @@
         var moduleid = -1;
         var action = string.Empty;
         var urlparameters = string.Empty;
-        var anchor = string.Empty;
         var editmode = false;
         var reload = Reload.None;
         var lastsyncdate = DateTime.UtcNow;
@@ -88,12 +87,6 @@
 
         // get path
         var path = uri.LocalPath.Substring(1);
-
-        //set anchor
-        if (uri.Fragment.StartsWith('#'))
-        {
-            anchor = uri.Fragment.Remove(0, 1);
-        }
 
         // parse querystring
         var querystring = ParseQueryString(uri.Query);
@@ -190,17 +183,17 @@
 
             int modIdPos = 0;
             int actionPos = 0;
-            int ulrParametersPos = 0;
+            int urlParametersPos = 0;
 
             for (int i = 0; i < segments.Length; i++)
             {
 
                 if (segments[i] == Constants.UrlParametersDelimiter)
                 {
-                    ulrParametersPos = i + 1;
+                    urlParametersPos = i + 1;
                 }
 
-                if (i >= ulrParametersPos && ulrParametersPos != 0)
+                if (i >= urlParametersPos && urlParametersPos != 0)
                 {
                     urlparameters += "/" + segments[i];
                 }
@@ -238,9 +231,9 @@
 
             }
 
-            if (ulrParametersPos > 0)
+            if (urlParametersPos > 0)
             {
-                path = path.Replace("/" + segments[ulrParametersPos - 1] + urlparameters, "");
+                path = path.Replace("/" + segments[urlParametersPos - 1] + urlparameters, "");
             }
 
             // remove trailing slash so it can be used as a key for Pages
@@ -311,7 +304,6 @@
                         Uri = new Uri(_absoluteUri, UriKind.Absolute),
                         QueryString = querystring,
                         UrlParameters = urlparameters,
-                        Anchor = anchor,
                         ModuleId = moduleid,
                         Action = action,
                         EditMode = editmode,
--- a/Oqtane.Client/UI/SiteRouter.razor
+++ b/Oqtane.Client/UI/SiteRouter.razor
@@ -209,7 +209,15 @@
                 {
                     modIdPos = i + 1;
                     actionPos = modIdPos + 1;
-                    action = segments[actionPos];
+                    if (actionPos > segments.Length - 1)
+                    {
+                        action = Constants.DefaultAction;
+                    }
+                    else
+                    {
+                        action = segments[actionPos];
+
+                    }
                 }
 
             }
@@ -219,7 +227,14 @@
             {
                 int.TryParse(segments[modIdPos], out result);
                 moduleid = result;
-                path = path.Replace("/" + segments[modIdPos - 1] + "/" + segments[modIdPos] + "/" + segments[actionPos], "");
+                if (actionPos > segments.Length - 1)
+                {
+                    path = path.Replace("/" + segments[modIdPos - 1] + "/" + segments[modIdPos], "");
+                }
+                else
+                {
+                    path = path.Replace("/" + segments[modIdPos - 1] + "/" + segments[modIdPos] + "/" + segments[actionPos], "");
+                }
 
             }
 
--- a/Oqtane.Client/UI/SiteRouter.razor
+++ b/Oqtane.Client/UI/SiteRouter.razor
@@ -75,6 +75,7 @@
         Page page;
         User user = null;
         List<Module> modules;
+        Dictionary<string, string> pageVariables = new Dictionary<string, string>();
         var moduleid = -1;
         var action = "";
         var editmode = false;
@@ -179,22 +180,32 @@
             // extract admin route elements from path
             var segments = path.Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
             int result;
-            // check if path has moduleid and action specification ie. pagename/moduleid/action/
-            if (segments.Length >= 2 && int.TryParse(segments[segments.Length - 2], out result))
+
+            int modIdPos = 0;
+
+            for (int i = 0; i < segments.Length; i++)
             {
-                action = segments[segments.Length - 1];
-                moduleid = result;
-                path = path.Replace(moduleid.ToString() + "/" + action + "/", "");
-            }
-            else
-            {
-                // check if path has moduleid specification ie. pagename/moduleid/
-                if (segments.Length >= 1 && int.TryParse(segments[segments.Length - 1], out result))
+                if (segments[i] == Constants.ModuleSegment)
                 {
-                    moduleid = result;
-                    path = path.Replace(moduleid.ToString() + "/", "");
+                    modIdPos = i + 1;
+                }
+
+                if (i > modIdPos && modIdPos != 0)
+                {
+                    action += segments[i] + "/";
                 }
             }
+
+            // check if path has moduleid and action specification ie. pagename/moduleid/action/
+            if (modIdPos > 0)
+            {
+                int.TryParse(segments[modIdPos], out result);
+                moduleid = result;
+                path = path.Replace(segments[modIdPos - 1] + "/" + segments[modIdPos] + "/" + action, "");
+
+            }
+
+            if (action.EndsWith("/")) action = action.Substring(0, action.Length - 1);
 
             // remove trailing slash so it can be used as a key for Pages
             if (path.EndsWith("/")) path = path.Substring(0, path.Length - 1);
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

**Comment by dnfadmin** [comment]
[CLA assistant check](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=638) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=638) before we can accept your contribution.<br/><br/>:x: PoisnFang [sign now](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=638)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/oqtane/oqtane.framework?pullRequest=638) it.</sub>

File: Oqtane.Server/appsettings.json [hunk 1]
```diff
@@ -1,6 +1,6 @@
 {
   "ConnectionStrings": {
-    "DefaultConnection": ""
+    "DefaultConnection": "Data Source=(LocalDb)\\MSSQLLocalDB;AttachDbFilename=|DataDirectory|\\Oqtane.mdf;Initial Catalog=Oqtane;Integrated Security=SSPI;"
```
mikecasas, line 3: This should  be empty. You may have forgot to delete your connection string.

sbwalker:
Thank you for submitting this PR - the ability for developers to have more control over their Urls is a useful feature. The technical direction taken with this PR is consistent with the rest of the architecture. However there are a few use cases which are not covered as well as some general comments/questions...

If we take a step back and focus on the purpose of this enhancement its goal is to allow module developers to pass parameters between components via the Url. Currently this is already supported using the Querystring but it would also be useful if it was supported via the Url path as well ( ie. for cleaner Urls ).

There are two aspects to routing in Oqtane - the Router itself which interprets urls and the Url helper methods which allow developers to create properly formatted urls. 

There are two Url Methods ( with a variety of overloads ):

NavigateUrl() - creates urls for navigating between "pages" in a site or modifying parameters for the current "page"
EditUrl() - creates urls for instantiating various components which are part of a specific module

Both of the above methods support Parameters already. Parameters are currently passed as strings in the form of "x=1&y=2" ( ie. Querystring format ).  

If we now focus on the Router, there are a number of conventions which are currently utilized to interpret a Url:

http://www.domain.com/alias/path1/path2 - a url to a specific "page" in a site
http://www.domain.com/alias/path1/path2/3 - a url that displays only a specific module ( id=3 ) on a "page" in a site 
http://www.domain.com/alias/path1/path2/3/edit - a url that displays a specific component for a module on a "page" in a site ( displayed in a modal by default )

All of these url conventions currently support Parameters as well:

http://www.domain.com/alias/path1/path2?page=3
http://www.domain.com/alias/path1/path2/3?id=4 
http://www.domain.com/alias/path1/path2/3/edit?id=1

One of the things to keep in mind is that some urls ( ie. those produced by NavigateUrl() ) are at the page level - not the module level. And since a page can contain many modules there is the potential for parameter conflicts between them. For example if ModuleA supported a parameter for id and ModuleB supported a parameter for category and they were both used on the same page, you would not want to generate a url in the format of http://www.domain.com/alias/path1/path2/3 as it is impossible to determine if "3" is an id or a category - both ModuleA and ModuleB would see the existence of the value "3" and would try to utilize it. To mitigate this issue, parameters should always provide an identifier with a value. Note that this is still not 100% effective as it relies on module developers to use unique identifiers but it does reduce the potential for conflicts to a point which is effective.

Currently it is very simple and fast for the Router to be able to identify Parameters - it looks for the "?" delimiter which identifies a Querystring and it knows that it only needs to process the path prior to the "?" to lookup the page in the database. In order to support Url Parameters in a similar manner, the router would need some form of delimiter to be able to quickly identify the parameters versus the path. 

Alternative Suggestions to the approach taken in this PR:

This PR uses a segment "/module/" which is the right approach, but I am not sure if it is the optimal choice for a delimiter. We may want to consider using something which is a bit shorter such as a segment like "/+/" or "/*/" ( or another unreserved character ). Then the router can easily identify Url Parameters in exactly the same way as Querystring parameters.

Rather than introduce some other method or approach for supporting Url parameters it would be optimal if the framework utilized the same exact Url helper methods. A suggestion is that we allow the developer to pass "x/1/y/2" as a parameter to the Url methods and internally based on the presence of the "/" character it would know that this should be appended to the Url path rather than querystring. Note that the Url parameters would be expected to be in the format of "name/value" in order to deal with the module conflict issue I outlined above.

The router could interpret parameters regardless of whether they are passed via the querystring or in the url in exactly the same way and populate a Parameters dictionary as part of PageState ( the current Querystring dictionary would be deprecated in a backward compatible manner ). Then module developers have a single dictionary object to work with which covers parameters passed using either method.

Oqtane.Server/appsettings.json
PoisnFang: I know, I figured I would remove it after the code was reviewed, because I was pretty sure it wouldn't be accepted right off the bat.

PoisnFang:
First off, thank you for looking at the PR and responding to quickly! Also, I am not trying to be rude, but you said that you had "questions" but you did not ask any, you also did not leave me with any specific action items so it makes it a bit tough for me to know how to respond. As such, I will just try to touch on key points based on the suggestions and your comments.

1. The delimiter can be anything and that is very easy to change, you just tell me what it should be.

2. One of the things I do not like with the Querystring Parameters is exposing the variable names to the user. it also helps to use specific examples rather than generic ones. This helps gives us a use case scenario.  For example: x/1/y/2, instead could be Dealerships/1/Cars/2 what you are saying here is that the key=Dealerships, value=1 I am not sure that I would really want my variable name to be the name of the parent page.

3. I would actually prefer to have 2 separate dictionaries to use. That way I know which dictionary to ask: Querystring Parameter vs. a Url Parameter. Typically Url parameters are used to Identify a specific object to display, such as a Blogs/1, whereas Querystring parameters are typically used to filter, such as Blogs?category=food

4. The current method of module routing does not allow for the use of numbers to be used as Url Paths for pages. Using the delimiter segment "/module/" (or what ever you want to put there) fixes that issue.

5. I am not sure what you are saying with the NavigateUrl() Example, can you clarify? 
This is that I understand: When a page is routed then check to see if it is trying to access a page in the Specified module `if (module.ModuleId == moduleid && action != "")` it does not really matter how many modules you have on the page because the Url parameters are Module Specific. Meaning Url parameters only work inside of modules. So with regards to your example "http://www.domain.com/alias/path1/path2/3" you are saying that the equivalent would be  http://www.domain.com/alias/path1/path2?id=3 OR  http://www.domain.com/alias/path1/path2?category=3. However based on this PR the Modules would not even receive the Url Parameter 3 because you have to be inside a module, like so:  http://www.domain.com/alias/path1/path2/module/3/Blogs/3

Let me know if this makes sense or how I can better clarify. Then I can make the appropriate changes.

sbwalker:
In regards to 5. above... let me provide a more practical example... 

Let's say there is a page in a site which shows details about cars. The url for this page is /cars. 

On that page there 2 modules:

Module1 - shows a dropdown list of different car models. When a user selects a car from the list and clicks the View button they are navigated to NavigateUrl(PageState.Page.Path, "id=" + carid) which produces a url of /cars?id=##.

Module2 - shows a photo of the car based on the id parameter in the route. 

Note: the developer chose to implement it this way rather than creating a single consolidated module because they want to be able to use Module2 on other pages in the site where a dropdown is not needed - they simply want it to be controlled by a parameter.

This all works fine. Now let's say the developer wants to use Url parameters instead of Querystring parameters. The url they want to use is /cars/*/## - where ## is the car id.

Now lets say there is a Module3 they want to use on this same page. This module has nothing to do with cars - it is loan calculator used on many pages in the site and it relies on the default currency being passed as a parameter. It would interpret the url /cars/*/## and assume that ## was a default currency parameter. It would assume this because there is no way for Module3 to know that the parameter actually belongs to Module1 and Module2. 

This is the conflict scenario I was trying to describe earlier, as not all urls are scoped to a module - some urls are scoped to a page where there are often multiple modules.

So what I was suggesting is that if Url parameters also included the name in addition to the value it mitigates the potential for conflict. This is in fact exactly how the current Querystring parameters deal with this same issue. So in the example above Module1 and Module2 would expect a url of /cars/*/carid/## and Module3 would expect a url of /cars/*/currency/xx. 

Hopefully this example makes it more clear.

PoisnFang:
I understand what you are trying to convent now I believe. However it is my belief that in the example given Url parameters should not be used at all. I think that Url Parameters and Querystring parameters should be used more or less for different things. this PR was not designed to Replace the Querystring parameters, it was designed to ENHANCE and provide additional options for developers.  

> This all works fine. Now let's say the developer wants to use Url parameters instead of Querystring parameters. The url they want to use is /cars/module/## - where ## is the car id.

This is not the intended design. What you are suggesting is that the index page of the module FILTER values based on the Url. That is what a Querystring is perfect for. I designed this to route to specific pages INSIDE the module. So once the Url parses out the module/{ModuleId} then it is left with the route to take inside the Module.

In the case of your example, the Modules on that page are displaying the `Index.razor` page from each of the Modules. In which case I would use the database to save the ModuleId to the Car and use programming logic to pull the specific car based on the ModuleId linked to it.

Here is a real use case from the admin portal: http://localhost:44357/admin/pages shows a list of pages in the site. When you click "Edit" it takes you INSIDE the PageModule to http://localhost:44357/admin/pages/module/12/Edit?id=1 the /admin/pages/ is just aesthetic for the user to know where they are. Technically you could have any OR all modules display on that page. I would rewrite my module so that it would look like this instead http://localhost:44357/admin/pages/module/12/Pages/1 the Querystring would get transformed into a Url string `Pages` would be the Index.razor page which is this: http://localhost:44357/admin/pages/module/12 which like I said is the list of all the Pages on the site, then `1` would select the pageId of 1. 

**The main use of this PR was to be able to route to specific pages using Url routes rather than Querystrings INSIDE of Modules**

sbwalker:
One thing to keep in mind is that although you may have a preference on when/how Url Parameters should be used versus Querystring parameters, not all developers are going to share your perspective. So I am trying to approach this from a higher level, as regardless of whether parameters are in the Url or in the Querystring, their purpose is for passing stateful information which can be utilized within components. And whatever solution we offer in Oqtane needs to be super simple and intuitive for developers - we should be trying to avoid solutions which are inconsistent with the rest of that architecture or are only applicable in very specific use cases, as this just creates confusion. 

Before going any further I want to make it clear that I think Url parameters are an important feature to support, so I am glad that you raised it. And you have raised a number of very valid points in this thread which I had not considered in my initial analysis. So the only thing I am trying to do at this point is to help arrive at a solution which conforms to the architectural goals and supports as many use cases as possible.

Let me offer one more practical example of why Url parameters are just as important in the page level scenario as they are in the module level scenario. The sample Blog module ( https://github.com/oqtane/oqtane.blogs ) has an Index component which supports 2 renderings. The default rendering displays a list of blog entries with summary information and each entry has a Details button. If a user clicks the Details button it replaces the list of blog entries with the full content for the specific blog selected. It currently uses a Querystring parameter to accomplish this - and it is designed in this way for 3 reasons: 1) it does not want to use the standard modal experience of module administration, 2) to support deep linking from outside the site to a specific blog entry, and 3) to support search engine indexing of content. Rather than using a Querystring parameter, I would much rather use a Url parameter in this case as it would allow me to create Urls which are much better for SEO. For example rather than having a Url of /blog?id=3 I could have a Url of /blog/delimiter/title-of-this-blog/3. I hope this example helps to better explain why a parameter solution needs to support this use case as well as the module level use case.

Recommendations ( note that these differ slightly from some of the feedback I provided earlier because you made some valid points that need to be supported... also your PR already supports some of these recommendations so they may seem redundant however I am including them so that we can be sure we are on the exact same page ):

1. We should rely on the existing NavigateUrl() and EditUrl() methods for creating Urls. These methods already support overloads which allow a developer to pass parameters as a string value. Currently parameters are expected to be in Querystring format. However it would be trivial to enhance this to support Url parameters as well. Some examples:

"id=3" - Querystring format ( this is the current format )
"name1=3&name2=4" - Querystring format ( this is the current format )
"/3" - Url format
"/3/4" - Url format with multiple parameters
"/3?sort=1" = mixture of Url and Querystring parameters
"#5" - an anchor to < a name="5" > somewhere on a page

Note that there is no specific convention enforced for Url parameters - a developer has the freedom to use whatever format they want. This should satisfy your requirement of not wanting to expose the intent of your parameters to the outside world - which is a valid concern. Also it should be intuitive as it is using standard url syntax which all developers should be familiar with - basically what they specify will be included as a suffix on the urls generated.

2. We should use a simple parameter /delimiter/. Valid characters are "$-_.+!*'(),". We would want to choose something that is not likely to ever be used as a page path on its own. I am open to suggestions.

3. The existing module-based Url parameters for module id and action should be moved after the delimiter. This solves the problem you identified where you currently cannot have pages with numeric paths as they conflict with the router. For consistency I also think that rather than having 3 Url conventions, it would be better for consistency to simply have 2. So basically rather than supporting a Url of /3/edit and a Url of /3 ( where 3 is the moduleid ) we should include the action "index" - so /3 becomes /3/index. Internally the site router can determine how an action named "index" should be handled ( since "index" is reserved ). This change will make it much better for developers who referencing parameters by Url segment position as it will always be consistent. 

4. In the site router, whatever comes after the "/delimiter/" ( and before "?" ) are the Url parameters and would be populated in a PageState.UrlParameters property. This property should NOT include the module id or action segments as they are internal and already included as individual properties in PageState. also they are not consistent with the parameters values which the developer will provide to the NavigateUrl() or EditUrl() methods. So if the Url was /path/delimiter/1/edit/3, the UrlParameters property would contain "/3"

5. To make it simpler for developers to deal with Url parameters we should have a helper method in ModuleBase:

Dictionary<string, string> GetUrlParameters(string parameterTemplate)

So if the Url was /path/delimiter/1/edit/3, the developer could call:

GetUrlParameters("/{id}");

and it would return a dictionary with 1 element with a name of "id" and a value of "3" ( internally this method would ensure that the number of segments in the template match the number of segments in the PageState.UrlParameters ). This method would be for convenience only. A developer would also be free to call PageState.UrlParameters..Split('/') and reference the parameters by ordinal position if they want.

PoisnFang:
Ok I can agree with your current assessment and your practical example helps put things in perspective for me. I can start working on edits to satisfy all the necessary requirements. I do have a technical question with your example and how we might achieve the desired functionality. 

>For example rather than having a Url of /blog?id=3 I could have a Url of /blog/delimiter/title-of-this-blog/3

Currently the PR looks for the `delimiter` and it takes the value right after it as the `ModuleId` In your example however there is not a number after the delimiter which makes sense because you are not going inside the module. The question is, how would you expect to your example Url to be parsed in the system?

sbwalker:
In the current system, by default there is a 1:1 mapping an action in the Url and a razor component with the same name. However there are also situations where you do not want a 1:1 mapping, so there is also an Actions property in IModuleControl which optionally allows you to define a 1:N mapping. You do this by specifying a comma delimited list of actions. For example, the Blog example module on Github uses this approach - in Edit.razor it specifies public override string Actions => "Add,Edit"; Basically this means that when either "add" or "edit" are specified in the Url they should be mapped to this razor component. Since this property already has a very specific purpose in the system we should not change its meaning.

However you do raise a very valid point in regards to parsing.... the challenge is that there are 2 distinct types of routes which need to be supported - page level and module level. So the question is how to definitively recognize the existence of the /{ModuleId}/{Action} in the Url. A couple options come to mind:

1. Leave /{ModuleId}/{Action} in their current location BEFORE the delimiter ( but still include the "Index" action explicitly rather than suppressing it ). The router will function the same as it does currently and will determine the existence of a module-based route by looking at the final 2 segments prior to the delimiter and examining if they are an integer followed by a string. This is similar to how the system works today. The downside is that there is potential for it to conflict with some specific hierarchical path structures which may want to use an integer and a string in the final 2 segments ( this is still an improvement over the current system which does not allow you to have a Url with an integer as the final segment ).

/{pagepath}/{delimiter}/{UrlParameters}?{Querystring}
/{pagepath}/{ModuleId}/{Action}/{delimiter}/{UrlParameters}?{Querystring}

2. Use 2 delimiters in the Url when the ModuleId and Action is required. The benefit is that there would never be a conflict with any path structure. The downside is an extra delimiter in module-level Urls.

/{pagepath}/{delimiter}/{UrlParameters}?{Querystring}
/{pagepath}/{delimiter}/{ModuleId}/{Action}/{delimiter}/{UrlParameters}?{Querystring}

Note that I personally view the ModulePathPath and UrlParameters in your example above as the same thing - they are additional state provided by a module in the Url and should simply be referred to as UrlParameters.

PoisnFang:
The second option is the same solution I came up with. I feel like it has the potential for less conflicts with hierarchical paths. 

>Note that I personally view the ModulePathPath and UrlParameters in your example above as the same thing - they are additional state provided by a module in the Url and should simply be referred to as UrlParameters.

I agree with that conclusion as well after thinking about it a bit more.

This is what i was thinking for the Delimeters:

```csharp
        public const string ModuleDelimiter = "*";
        public const string UrlParametersDelimiter = "!";

sbwalker:
Do you think it would it be possible to use the same delimiter character for both scenarios?

PoisnFang:
My initial though it 'No'. A few use cases to consider below. The main conflict I believe will be trying to parse the difference between the first two.

Page page with Url Parameters but not inside Module:
`/{pagepath}/{delimiter}/{UrlParameters}?{Querystring}`

Page page with Inside Module but without Url Parameters:
`/{pagepath}/{delimiter}/{ModuleId}/{Action}?{Querystring}`

Page page with Inside Module with Url Parameters:
`/{pagepath}/{delimiter}/{ModuleId}/{Action}/{delimiter}/{UrlParameters}?{Querystring}`

sbwalker:
You are correct.

PoisnFang:
It appears that the consistent name for the querystring parameters variable is `parameters`
`public static string NavigateUrl(string alias, string path, string parameters)`

in the `ActionLink.razor`:
`private string _parameters = string.Empty;`
`public string Parameters { get; set; }`

I would like to rename these variable/properties to `querystring` (`_querystring`,`QueryString`) I feel it provides better clarity. It will also hep match with the property in `PageState.cs` ` public Dictionary<string, string> QueryString { get; set; }`

I will name the Url Parameters variable/properties to: `urlparameters` (`_urlparameters`,`UrlParameters`)

**I realize that this will be a BREAKING change to existing Modules.**

sbwalker:
We should avoid a breaking change. We should leave the NavigateUrl() and EditUrl() methods in their current form. I already outlined earlier in this thread some of the different values that could be passed in the "parameters" parameter. It it not difficult to parse/process them intelligently - they do not need to be passed as 2 different parameters.

sbwalker:
There were the examples that could be passed in the "parameters" parameter:

"id=3" - Querystring format ( this is the current format )
"name1=3&name2=4" - Querystring format ( this is the current format )
"/3" - Url format
"/3/4" - Url format with multiple parameters
"/3?sort=1" = mixture of Url and Querystring parameters
"#5" - an anchor to < a name="5" > somewhere on a page

For backward compatibility, if the value contains "=" and does not contain "?" then it is a querystring and should have a "?" prefix added.
  
In all other cases the value will simply be appended to the url generated.

PoisnFang:
Ok, I am glad I ran it by you first. I can do that.

>"#5" - an anchor to < a name="5" > somewhere on a page

Can you give an example of how this would look combined with the Url and Querystring?

sbwalker:
An anchor is always added at the very end of a Url. This should not matter to the implementation as whatever value is provided in the "parameter" parameter is appended to the Url generated.

NavigateUrl("path", "#23") = /path#23
NavigateUrl("path", "/5#23") = /path/5#23
NavigateUrl("path", "id=5#23") = /path?id=5#23

PoisnFang:
Ok, awesome that helps a lot! I am just a very visual person.

PoisnFang:
I have ran into a bit a a snag with parsing. Here are all of the possible parameter strings I could think of:

/urlparameters /urlparameters?Id=1 
/urlparameters#5 
/urlparameters?Id=1#5 
/urlparameters?reload#5
Id=1 Id=1#5
reload#5 
reload
#5

In order to easily parse between the 3 types we have to make it so that the querystring must contain `=` otherwise the parsing is very difficult.

sbwalker:
I am not sure what you are trying to parse? Are you referring to the site router? There is already logic for parsing a querystring which should not need to change.

sbwalker:
The site router should evaluate the url in from the end to the beginning:
- if it contains # then remove the # and everything after it and ignore it
- if it contains ? then remove everything everything after it and parse the querystring
- if it contains the url delimiter then remove everything after it and assign to UrlParameter
- if it contains the module delimiter then remove and populate ModuleId and Action
- you are left with alias and path

PoisnFang:
Here is what I am trying to do:

```csharp
                var chars = new[] { '?', '#' };
                var parameters = string.Empty;
                var parametersArr = parameters.Split(chars, StringSplitOptions.RemoveEmptyEntries);

                for (int i = 0; i < parametersArr.Length; i++)
                {
                    if (parametersArr[i].Contains('/'))
                    {
                        urlparameters = parametersArr[i];
                    }
                    else if (parametersArr[i].Contains('='))
                    {
                        querystring = parametersArr[i];
                    }
                    else
                    {
                        querystring = parametersArr[i];
                    }
                }

PoisnFang:
I committed the changes based on our conversation and I did some testing. The changes work in my Module without breaking anything. Take a look and see if I covered all of your suggestions and if there is anything else I need to do.

sbwalker:
This looks good... only 2 comments:

1. There is a variable named "ulrParametersPos" which appears to be a typo - it should be "urlParametersPos"

2. Anchors are known as "fragment identifiers" and are not considered part of the URL ( according to the HTML spec ). The anchor tag is never sent as part of the HTTP request to the server, it is only interpreted locally within the browser. So although it is useful for us to support them via NavigateUrl() and EditUrl() I am not sure if it makes sense to have a dedicated property for them in PageState?

PoisnFang:
Oh oops I'll fix that. Yeah I honestly wasnt too sure about the anchor, should I just remove that from the PageState?

sbwalker:
Yes, I think we should remove it. If a developer wants to utilize it it is available through the PageState.Uri.Fragment property already.

sbwalker:
> Edit: You will notice that in the GetUrlParameters method I still used brackets `"/[id]"` instead of curly braces `"/{id}"` that was done on purpose so the devloper can take advantage of the the use of the dollar sign with curly brackets in a string `$"/[{VariableName}]"`

I was actually thinking that it would be more intuitive for .NET developers if the url parameter template used the same syntax as other .NET route templates ie. Blazor uses @page "/{id}", and MVC uses HttpGet("{id}"). I realize it would restrict the use of string interpolation ($) but I feel that using consistent coding conventions should be a higher priority than being able to use a specific string management feature.

PoisnFang:
I am going to have to disagree with that one because I use constant string for my variables to avoid magic strings so I Need the interpolation. Thise example use the curly braces because interpolation is built in. Our would not have that which would be an inconsistency.

 We aren't automatically assigning  values to properties like blazor routing does. That alone could actually confuse developers because it doesn't work the same way.

sbwalker:
Fair enough. We can use [ ] as token identifiers.

sbwalker:
But you realize that you can use {{ or }} in string interpolation to produce a brace character? ie. if string name = "id"; then "/{{{name}}}" produces "{id}" - so using { } would not obstruct any C# feature

PoisnFang:
Alright I see your point. I am not a fan of how that really looks but you are right it will keep a certain level of consistency.

PoisnFang:
That should complete all of the changes.

sbwalker:
I merged the PR but there seems to be a bug somewhere. The first thing I tried was going to the Home page in a default installation. ( ie. the Url is http://localhost:44357/ ). I go into Edit Mode and then I click the Edit button on the Html module at the top of the page. The modal is not displayed for editing content. I also tried using the module action menu to Manage Settings and it does not work. The problem seems to be localized to the Home page which has a Path of "".

PoisnFang:
hmmm, I am debugging right now.

PoisnFang:
Found it, its on line 229 in the SiteRouter

`path = path.Replace("/" + segments[modIdPos - 1] + "/" + segments[modIdPos] + "/" + segments[actionPos], "");`

because the path is ""*/1/Edit/"" it wont replace because of the "/"

PoisnFang:
I fixed it I will open a new PR

PR #640, opened 2020-06-30
Url parameter helper enhancements

PoisnFang:
Currently the url parameter segments that are not variables are being discarded. It would be helpful to developers to be able to access the "action" segments, example:
`$"{{{parentId}}}/Edit/{{{childId}}}"` and `$"{{{parentId}}}/View/{{{childId}}}"`
Saving the "action" segments allows the developer to parse which "page" they are on if they want to ie: `Edit` or `View`

```csharp
            if (UrlParamerters.ContainsKey("action1"))
            {
                switch (UrlParamerters["action1"])
                {
                    case "Edit":
                        //set parameters
                        break;

                    case "View":
                        //set parameters
                        break;

                    default:
                        break;
                }
            }

### DIFF
--- a/Oqtane.Client/Modules/ModuleBase.cs
+++ b/Oqtane.Client/Modules/ModuleBase.cs
@@ -44,6 +44,26 @@ namespace Oqtane.Modules
 
         public virtual List<Resource> Resources { get; set; }
 
+        public virtual string UrlParametersTemplate { get; set; } = "";
+        public virtual Dictionary<string, string> UrlParamerters
+        {
+            get
+            {
+                var urlparameters = new Dictionary<string, string>();
+
+                var templates = UrlParametersTemplate.Split(',', StringSplitOptions.RemoveEmptyEntries);
+                foreach (var template in templates)
+                {
+                    urlparameters = GetUrlParameters(template);
+
+                    if (urlparameters.Count > 0) goto Return;
+                }
+
+                Return:
+                return urlparameters;
+            }
+        }
+
         // base lifecycle method for handling JSInterop script registration
 
         protected override async Task OnAfterRenderAsync(bool firstRender)
@@ -116,7 +136,7 @@ namespace Oqtane.Modules
             return Utilities.ContentUrl(PageState.Alias, fileid);
         }
 
-        public Dictionary<string, string> GetUrlParameters(string parameterTemplate)
+        public virtual Dictionary<string, string> GetUrlParameters(string parameterTemplate)
         {
             var urlParameters = new Dictionary<string, string>();
 
@@ -131,7 +151,7 @@ namespace Oqtane.Modules
                     {
                         if (templateSegments[i] == parameters[i])
                         {
-                            urlParameters.TryAdd("action" + actionId, parameters[i]);
+                            urlParameters.TryAdd("parameter" + actionId, parameters[i]);
                             actionId += 1;
                         }
                         else if (templateSegments[i].StartsWith("{") && templateSegments[i].EndsWith("}"))
--- a/Oqtane.Client/Modules/ModuleBase.cs
+++ b/Oqtane.Client/Modules/ModuleBase.cs
@@ -44,26 +44,6 @@ namespace Oqtane.Modules
 
         public virtual List<Resource> Resources { get; set; }
 
-        public virtual string UrlParametersTemplate { get; set; } = "";
-        public virtual Dictionary<string, string> UrlParamerters
-        {
-            get
-            {
-                var urlparameters = new Dictionary<string, string>();
-
-                var templates = UrlParametersTemplate.Split(',', StringSplitOptions.RemoveEmptyEntries);
-                foreach (var template in templates)
-                {
-                    urlparameters = GetUrlParameters(template);
-
-                    if (urlparameters.Count > 0) goto Return;
-                }
-
-                Return:
-                return urlparameters;
-            }
-        }
-
         // base lifecycle method for handling JSInterop script registration
 
         protected override async Task OnAfterRenderAsync(bool firstRender)
@@ -136,37 +116,51 @@ namespace Oqtane.Modules
             return Utilities.ContentUrl(PageState.Alias, fileid);
         }
 
-        public virtual Dictionary<string, string> GetUrlParameters(string parameterTemplate)
+        public virtual Dictionary<string, string> GetUrlParameters(string parametersTemplate = "")
         {
             var urlParameters = new Dictionary<string, string>();
-
-            var templateSegments = parameterTemplate.Split('/', StringSplitOptions.RemoveEmptyEntries);
+            string[] templateSegments;
             var parameters = PageState.UrlParameters.Split('/', StringSplitOptions.RemoveEmptyEntries);
-            var actionId = 1;
-            if (parameters.Length == templateSegments.Length)
+            var parameterId = 0;
+
+            if (string.IsNullOrEmpty(parametersTemplate))
             {
                 for (int i = 0; i < parameters.Length; i++)
                 {
-                    if (parameters.Length > i)
+                    urlParameters.TryAdd("parameter" + i, parameters[i]);
+                }
+            }
+            else
+            {
+                templateSegments = parametersTemplate.Split('/', StringSplitOptions.RemoveEmptyEntries);
+
+                if (parameters.Length == templateSegments.Length)
+                {
+                    for (int i = 0; i < parameters.Length; i++)
                     {
-                        if (templateSegments[i] == parameters[i])
+                        if (parameters.Length > i)
                         {
-                            urlParameters.TryAdd("parameter" + actionId, parameters[i]);
-                            actionId += 1;
-                        }
-                        else if (templateSegments[i].StartsWith("{") && templateSegments[i].EndsWith("}"))
-                        {
-                            var key = templateSegments[i].Replace("{", "");
-                            key = key.Replace("}", "");
-                            urlParameters.TryAdd(key, parameters[i]);
-                        }
-                        else
-                        {
-                            i = parameters.Length;
+                            if (templateSegments[i] == parameters[i])
+                            {
+                                urlParameters.TryAdd("parameter" + parameterId, parameters[i]);
+                                parameterId++;
+                            }
+                            else if (templateSegments[i].StartsWith("{") && templateSegments[i].EndsWith("}"))
+                            {
+                                var key = templateSegments[i].Replace("{", "");
+                                key = key.Replace("}", "");
+                                urlParameters.TryAdd(key, parameters[i]);
+                            }
+                            else
+                            {
+                                i = parameters.Length;
+                                urlParameters.Clear();
+                            }
                         }
                     }
                 }
             }
+
             return urlParameters;
         }
 
--- a/Oqtane.Client/Modules/ModuleBase.cs
+++ b/Oqtane.Client/Modules/ModuleBase.cs
@@ -122,7 +122,7 @@ namespace Oqtane.Modules
 
             var templateSegments = parameterTemplate.Split('/', StringSplitOptions.RemoveEmptyEntries);
             var parameters = PageState.UrlParameters.Split('/', StringSplitOptions.RemoveEmptyEntries);
-
+            var actionId = 1;
             if (parameters.Length == templateSegments.Length)
             {
                 for (int i = 0; i < parameters.Length; i++)
@@ -131,6 +131,8 @@ namespace Oqtane.Modules
                     {
                         if (templateSegments[i] == parameters[i])
                         {
+                            urlParameters.TryAdd("action" + actionId, parameters[i]);
+                            actionId += 1;
                         }
                         else if (templateSegments[i].StartsWith("{") && templateSegments[i].EndsWith("}"))
                         {

### CODE REVIEW

**Comment by PoisnFang** [comment]
Also should the method in ModuleBase.cs `public Dictionary<string, string> GetUrlParameters(string parameterTemplate)` be `virtual`?

**Comment by sbwalker** [comment]
There are already PageState.ModuleId and PageState.Action properties that exist for this exact purpose. They are already being used by modules such as the Blog sample module on Githib.

**Comment by sbwalker** [comment]
This was explained as part of our conversation in #638

“4. In the site router, whatever comes after the "/delimiter/" ( and before "?" ) are the Url parameters and would be populated in a PageState.UrlParameters property. This property should NOT include the module id or action segments as >>> they are internal and already included as individual properties in PageState <<<<. also they are not consistent with the parameters values which the developer will provide to the NavigateUrl() or EditUrl() methods. So if the Url was /path/delimiter/1/edit/3, the UrlParameters property would contain "/3"”

**Comment by PoisnFang** [comment]
Let me clarify, here is an absolute URL:
`http://localhost:44357/mypage/*/27/FunCars/!/Edit/Cars/1?id=3#5` which follows this format `http://localhost:44357/mypage/{ModuelDelimeter}/{ModuleId}/ManageForms/{ModuleDelimeter}/{action1}/{action2}/{id}?{querystring}#{anchor}`

the `action1` and `action2` string are not being saved in the  `Dictionary<string, string> GetUrlParameters`

ModuleBase.cs line 132: 

```csharp
                        if (templateSegments[i] == parameters[i])
                        {
                            // do nothing with this segment
                           // This is where you can add segment to dictionary as a urlparameter action
                        }
                        else if (templateSegments[i].StartsWith("{") && templateSegments[i].EndsWith("}"))
                        {
                            var key = templateSegments[i].Replace("{", "");
                            key = key.Replace("}", "");
                            urlParameters.TryAdd(key, parameters[i]);
                        }
                        else
                        {
                            i = parameters.Length;
                        }

**Comment by sbwalker** [comment]
My apologies... I saw the term “action” and thought you were referring to the action segment of the url. This is exactly why choosing names which are non-confusing is really important. The segments that are part of the url parameters should not be called actions - they should probably just be called parameter1, parameter2, etc... so that they are not confused with the actual action url segment. Regardless I do understand the purpose of this enhancement and agree that all segment values should be retained.

**Comment by PoisnFang** [comment]
Yeah my bad on the naming

**Comment by PoisnFang** [comment]
I also added this into my Inherited Class of the ModuleBase.cs, not sure if you think it would be good to add to the main project?

```csharp
        public virtual string UrlParametersTemplate { get; set; } = "";
        public virtual Dictionary<string, string> UrlParamerters
        {
            get
            {
                var urlparameters = new Dictionary<string, string>();

                var templates = UrlParametersTemplate.Split(',', StringSplitOptions.RemoveEmptyEntries);
                foreach (var template in templates)
                {
                    urlparameters = GetUrlParameters(template);

                    if (urlparameters.Count > 0) goto Return;
                }

                Return:
                return urlparameters;
            }
            set { }
        }
```
I found it useful in my Module Page Items.cs like so:

```csharp
        public override string UrlParametersTemplate
        {
            get
            {
                return $"Add," + //Add 
                    $"{{{parentId}}}," + //Edit item
                    $"{{{parentId}}}/Children," + //Manage Children
                    $"{{{parentId}}}/Children/Add," + //Add Child item
                    $"{{{parentId}}}/Children/{{{childId}}}"; //Edit Child item
            }
        }

**Comment by sbwalker** [comment]
after your enhancement, another option for getting an item based on its ordinal position is:

var dict = GetUrlParamwters(“/{id}/edit/{name}”);
string parameter1 = dict[dict.ElementAt(1).Key];

With this code the developer does not need to know the item is named “parameter1”

**Comment by PoisnFang** [comment]
Oh you make a good point. Do you think it would be good to have the Dictionary property in ModuleBase that auto checks UrlParameters, that way you don't have to call GetUrlParameters() on every module page?

**Comment by sbwalker** [comment]
I was thinking about this more last night... and I think we are thinking about the url template differently. The original idea of a GetUrlParameters() helper in ModuleBase was that it could provide a consistent dictionary structure so that a developer could treat Url Parameters similar to Querystring Parameters ( ie. by referencing parameters by name, checking for parameter existence, etc... ). 

The concept was intended to be very simple - the template provides the "name" elements which can be mapped to the values in the Url to provide name/value pairs in the dictionary. What I envisioned is that if you had a Url Parameter of "/1/edit/2" then your url template would be "/{parentid}/{action}/{childid}" which produces a dictionary with parentid=1, action=edit, and childid=2. You can see how it was intended to be a very simple mapping to map names to values. I had never imagined someone using hard-coded names in the url template - and I still do not understand why you would want to do that as it defeats the whole point of GetUrlParameters().

The only thing I can guess - which is not 100% clear based on the code, is that you are trying to use the url template both as way to parse parameters in GetUrlParameters() as well as a string interpolation template for creating the "parameter" string to pass to EditUrl()?

**Comment by sbwalker** [comment]
It also appears that you are not using the Action concept the way it is intended to be used and you are duplicate its functionality with Url Parameters. Let me explain...

If you have a razor component called Manager.razor that you want to handle both Add and Edit "actions" then you should use the Actions property in your razor component :

public override string Actions { get { return "Add,Edit"; } }

This tells the framework that if the url has an Action segment of either "Add" or "Edit" that it will be handled by Manage.razor. And you would use the EditUrl() framework method to create links by using its Action parameter. 

ie. EditUrl("Add"); or EditUrl("Edit", parameters);

The module is then able to rely on PageState.Action to perform conditional logic.

I realize that we cant prevent developers from trying to use the Url Parameters or Querystring Parameters for this same purpose but we should definitely not encourage it.

**Comment by sbwalker** [comment]
Is this essentially what you are trying to do in your module?

// define a url template
string urltemplate = "/{parentid}/{action}/{childid}";
...

// create dictionary based on url parameters and use its elements
var dict = GetUrlParameters(urltemplate);
parentid = dict["parentid"];
action = dict["action"];
...

// create a url based on the url template
string url = EditUrl("Edit", string.Format(urltemplate, parentid, action, childid));
// unfortunately C# does not support the following syntax 
// string url = EditUrl("Edit", $(urltemplate));

**Comment by sbwalker** [comment]
I think where I confused matters is by specifying a url template example which included braces ie. "/{parentid}/{action}/{childid}" which you interpreted to mean that it was intended to support dynamic replacement. That was never my intention - I only used braces because other route templates in .NET use them. If I had instead provided an example of "/parentid/action/childid" then it would have been clear it was only for mapping purposes to populate the dictionary with names to go along with the values in the url.

**Comment by chlupac** [comment]
I have thought about this topic. I think it's a bit recombined and it doesn't solve the basic problem - parameters (status) for more modules on the page. My idea is to create a PageStatus in the form [pagemoduleid1: [moduleParameter1: value1, moduleparameter2: value2], ....] serialize it into json and use Base64UrlEncode to create a PageStatus Url value. And it doesn't matter if it is used as a url part or url parameter.
This would allow independent status settings for individual modules on the page with simple json serialize/deserialize.

for example: http://localhost:44357/mypage/Ki8yNy9GdW5DYXJzLyEvRWRpdC9DYXJzLzE_aWQ9MyM1
or http://localhost:44357/mypage?state=Ki8yNy9GdW5DYXJzLyEvRWRpdC9DYXJzLzE_aWQ9MyM1

Just an idea when this topic is at stake.

**Comment by sbwalker** [comment]
@chlupac its a clever solution to the page level parameter conflict issue I outlined in the other thread. However it would not be a viable general solution as it eliminates the ability for urls to be clean and intuitive, and would not be good for SEO. One thing that is interesting is that the same parameter conflict issue has existed in DNN for 18 years... and it is rarely ever reported as a problem - it is a theoretical problem but its probability of occurring is very low. That being said, DNN does not support developer specified "url parameters" - it forces developers to use querystring parameters and then it uses a url rewriter to transform them into url parameters ie. path?id=3 becomes path/id/3. And because every parameter has an explicit name and value segment in the url it greatly reduces the probability of conflicts from occurring - essentially parameters are "namespaced". This new enhancement to Oqtane increases the probability of conflicts occurring for page level parameters as it has no "namespace" concept. Hopefully this does not become a major problem - we will need to monitor it closely.

**Comment by PoisnFang** [comment]
Wow a lot to wake up to read... But ok here goes @sbwalker 
My intent from the beginning was to always be able to have Url page routing inside the module. What was so nice about the solution we came up with is that it allows for both what you are describing and what I am doing. I am hardcoding the "action" (now parameter) into the Url parameters template. However I am also using the `PageState.Action` which I use to route to my `Page.cs` in the Module. I am using the Url Parameters template to logically parse the Url inside the Module because I want for Url to look very nested. But I also realized that I had a lot of code that was the same across the different .cs files, so by consolidating it I was have to manage less code.

All in all, IMO the beauty of the system is it allows for both scenarios. The way you are describing and the way I am using it. I am using a single Page.razor.cs as basically my own "Dynamic" page in my module. Which is simple because I can have multiple Page.razor components. Which is exactly what blazor was intended for. The Url template string for me become the "site map" of the "dynamic" page in my module.

**Comment by sbwalker** [comment]
Thank you for providing the additional context.... it actually helps explain why you have so many parameters in your modules ( which was something I never understood previously as it is not typical in any of the core modules ). The way you are developing modules does not follow the pattern used in the core framework but I can understand your arguments - especially in terms of reducing the number of razor files you need to maintain. And you are correct that one of the overarching themes in Oqtane is flexibility - so it is beneficial that the system is flexible enough to support multiple methods of module development. 

Getting back to the confusion about this specific PR... the first thing to keep in mind is that the GetUrlParameters() method in ModuleBase is not required in order to use Url parameters. So removing it altogether could be considered a viable solution as it would keep ModuleBase leaner and simpler and would result in less code to maintain in the core framework. 

In order to retain it we need to share a common understanding of how it will be used. When I suggested the GetUrlParameters() method I was only considering it to be a simple helper method for creating a dictionary from Url parameters. I was not thinking about it as part of a module routing solution. However as part of this discussion I can now see how a Url parameter template could be beneficial for creating urls as well.

For example, if there was property in ModuleBase for UrlParameterTemplate a module could specify:

public override string UrlParametersTemplate => "/{parentid}/{action}/{childid}";

The main rule is that all "dynamic" elements MUST be identified with braces {} and all other items will be treated as static parts of the url parameters string.

A developer would be able to call GetUrlParameters() and it would use the UrlParametersTemplate and PageState.UrlParameters to construct a dictionary. However this dictionary would only contain the dynamic elements from the template. So if you want the action parameter to be present in the dictionary then you need to wrap it in braces {action} and use a variable rather than hardcoding a value.

var dict = GetUrlParameters();
int parentid = (int)dict["parentid"];
string action = dict["action"];
int childid = (int)dict["childid"];

A developer could also call SetUrlParameters(params object[] parameters) and it would use the UrlParametersTemplate to create Url parameters which can be passed to the standard Url methods:

string url = EditUrl("Module", SetUrlParameters(parentid, action, childid));
string url = NavigateUrl("Module", SetUrlParameters(1, "edit", "2"));

So basically this provides a simple mechanism for ensuring url parameters are handled in a consistent manner within a module - both when creating and parsing url parameters. You will notice that string interpolation is not being used at all in what I described above - so there is no need for extra escaping of elements in the template ( ie. "{{{id}}}" ).  

I also believe supporting multiple url parameter templates withing the same module is out of scope for the core framework - as it gets too far away from a simple, generic solution.

**Comment by PoisnFang** [comment]
Yeah... my module is actually a business application platform, so it huge and complex. 
That's fine I don't mind simplifying the `GetUrlParameters()` so the you MUST use { } for every segment, as long as we keep it virtual. I am assuming that `SetUrlParameters(1, "edit", "2")` would return `"/1/edit/2`?

To simply the `GetUrlParameters(string parameterTemplate)` we pretty much just need to remove this section that scans for hardcoded values:

```csharp
                        if (templateSegments[i] == parameters[i])
                        {
                            urlParameters.TryAdd("parameter" + parameterId, parameters[i]);
                            parameterId += 1;
                        }
```
Also if we have the `UrlParametersTemplate ` Property do we need to pass a string into the `GetUrlParameters`?

sbwalker:
Sorry for going back and forth on this... trying to make it very simple and generic:

add to ModuleBase:

UrlParametersTemplate;

Within module:

public override string UrlParametersTemplate => "/{parentid}/edit/{childid}";

( this would be optional and UrlParametersTemplate would be initialized to "" by default in ModuleBase )

A developer would be able to call GetUrlParameters() and it would use the PageState.UrlParameters to construct a dictionary. Each element in the dictionary maps to a url segment. If a {token} exists for a url segment in UrlParametersTemplate then use it as the name. If a {token} was not provided for a segment in UrlParametersTemplate use "parameter#" as the name. If UrlParametersTemplate was "" then populate the dictionary with parameter0, parameter1, etc... 

var dict = GetUrlParameters();
int parentid = (int)dict["parentid"];
string action = dict.ElementAt(1).Value;
int childid = (int)dict["childid"];

A developer could call EditUrl() and NavigateUrl() new overload methods params object[] UrlParameters which would use the UrlParametersTemplate ( or if UrlParametersTemplate was "" it would just create a string delimited with "/" characters )

string url = EditUrl("Module", parentid, action, childid);
string url = NavigateUrl("Module", 1, "edit", 2); 

What are your thoughts?

PoisnFang:
That's fine its good to flesh it out. I know I keep coming back with things, I am not trying to be difficult. 
I can change GetUrlParameters() to this `GetUrlParameters(string parametersTemplate = "")` auto setting the variable to "" and then 
```csharp
            string[] templateSegments;
            if (string.IsNullOrEmpty(parametersTemplate))
            {
                templateSegments = parametersTemplate.Split('/', StringSplitOptions.RemoveEmptyEntries);
            }
            else
            {
                templateSegments = UrlParametersTemplate.Split('/', StringSplitOptions.RemoveEmptyEntries);
            }
```
This allows a developer to have multiple Urls in the UrlParametersTemplate property `UrlParametersTemplate => "/{parentid}/edit/{childid}"` OR `UrlParametersTemplate => "/{parentid}/edit/{childid},/edit/{id}"` Because GetUrlParameters() does not split UrlParametersTemplate by comma. Which splitting allows for:

```csharp
                var templates = UrlParametersTemplate.Split(',', StringSplitOptions.RemoveEmptyEntries);
                foreach (var template in templates)
                {
                    urlparameters = GetUrlParameters(template);
                }
```
> A developer could call EditUrl() and NavigateUrl() new overload methods params object[] UrlParameters which would use the UrlParametersTemplate ( or if UrlParametersTemplate was "" it would just create a string delimited with "/" characters )

I am a bit confused with this? What does the UrlParametersTemplate  have to do with EditUrl() and NavigateUrl()?

When you pass `parameters` into `NavigateUrl()` it looks like this "/1/edit/2?id=3#" then the method parses it out in a complete url based on the `alias` and `path`

sbwalker:
I feel like although all this back and forth discussion is useful, it is getting us further away from the simplest possible solution - and in general the simplest solution is usually the best because it is the easiest to adapt to different use cases which may come up in the future. So my final recommendation here is that we go with a simple method in ModuleBase for:

public virtual Dictionary<string, string> GetUrlParameters(string parameterTemplate)

It will do exactly what it does currently - parses UrlParameters into a dictionary and uses the {token} values in the template as dictionary item keys where specified, and uses "parameter#" as dictionary item keys where they is not specified. In the case where a developer calls GetUrlParameters("") it should return a dictionary that contains all segments with "parameter#" item keys ( this allows it to be useful even without a template ).

ie.

Example UrlParameter value: "/2/edit/3"

GetUrlParameters("/{id1}/{name}/{id2}");
Dictionary: ["id1","2"],["name","edit"],["id2","3"]

GetUrlParameters("/{id1}/edit/{id2}");
Dictionary: ["id1","2"],["parameter1","edit"],["id2","3"]

GetUrlParameters("");
Dictionary: ["parameter0","2"],["parameter1","edit"],["parameter2","3"]


Module developers will be responsible for keeping track of their own template ( or templates ).

PoisnFang:
Ah, ok thanks for the clarification. I can do that. Sorry I dragged this one out.

PoisnFang:
That should do it. I tested all 3 scenarios in my project.

sbwalker:
I really appreciate your patience working through this Url Parameter enhancement. I feel it is an elegant solution which provides a lot more flexibility for developers and improves the overall framework capabilities.

PoisnFang:
No problem, I feel the same. This solution has already greatly helped organize my Module and increased my efficiency in it.

PR #641, opened 2020-07-01
Notification Repository Breaking change fix

chlupac:
I wondered why my users complains than messages had not been delivered.

### DIFF
--- a/Oqtane.Server/Repository/NotificationRepository.cs
+++ b/Oqtane.Server/Repository/NotificationRepository.cs
@@ -21,7 +21,7 @@ namespace Oqtane.Repository
                 return _db.Notification
                     .Where(item => item.SiteId == siteId)
                     .Where(item => item.IsDelivered == false)
-                    .Where(item => item.SendOn < System.DateTime.UtcNow)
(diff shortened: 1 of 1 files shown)


sbwalker:
I integrated a PR enhancement #625 but it appears that it affected existing system behavior. Noting @mikecasas so that he is aware as he created the PR.

sbwalker:
@chlupac did this affect all notifications or just items created prior to the PR being integrated?

chlupac:
All notifications which not fill SendOn field. My system stop to send custom forgot password messages. Corrected now.

sbwalker:
Sorry... I should have caught this, as you are correct that any custom app that was using the Notification API would not have implemented the new property - so it would have been null - and the system should have handled this use case.

chlupac:
No problem, my complains department solved this quickly :) Mike was trying to correct this bug, but on wrong place.

mikecasas:
Wouldn't the sql update statement have fixed this? Should I have declared the property in the db as non null?

chlupac:
@mikecasas  no Sql do not fixed it, it requires change in existing code, because when notification is created, field is not filled.  
But for case when I want to send message immediately, null is ok. It's fixed now, no further changes required.

PR #643, opened 2020-07-03
Notification job optimalization

chlupac:


sbwalker:
@chlupac After the changes in this PR, there is no status messages in the job log. This is a problem as there needs to be feedback that the job is actually doing something. Also after this PR, the job is retrieving the notifications even if the SMTP settings have not been configured yet by the Administrator - which basically means it is doing work that is unnecessary. This does not seem to be an optimization - in fact it seems to have the opposite effect?

chlupac:
Yes, it's mistake because I do not recognize than log entry is created regardless of job output. 

but
some comments:

+ Logging of this should be done on debug level. Log is bloated with useless repeating messages.
+ no indication of error or status in job, job returns only string with accumulated messages from all sites (what if I have 10, 20, 500 sites ..... ????) maybe setting for site `MessagingJobEnabled = true` (#629 custom settings for entities) would help
+ I do not want messaging on all sites, read again and again than smtp is not configured is useless. Error message about SMTP server will show when first message for site is in queue.

sbwalker:
The production site scenario is the environment where I personally would want ALL log messages to be recorded as I want a record of everything that is happening ( or not happening ) in production. 

The job API definition is purposely very generic so that it can accommodate any type of job. It is up to each individual job to decide what to do with errors, how much logging to include, etc...

A site setting for enabling/disabling notifications could be added to the existing section in Site Settings which asks for SMTP details. #629 is not required to implement this.

sbwalker:
Another option if you want to remove all logging in your specific installation, simply create your own Notification Job by copying the current one and modifying it. Disable the default job and enable your own.

Issue #653, opened 2020-07-12
Bug: CreateAuthPolicyUrl


mikecasas:
In the external module, the private method ```CreateAuthPolicyUrl``` in the ```[Module]Service``` class adds the entityid parameter twice, so that when ```PermissionHandler``` trys to parse the entity id (the querystring value is 27,27 ) it throws the ```System.FormatException: 'Input string was not in a correct format.'```

I think the fix should be in the ```Add[Module]Async``` method. I think entityid should be moduleid?

mikecasas:
Track in #654

PR #654, opened 2020-07-12
Should be moduleid as the entity id is added in the CreateAuthPolicyU…

mikecasas:
Fix for #653


sbwalker:
Actually the proper fix should be to remove the entityid parameter as it is now being added by the CreateAuthPolicyUrl method:


PostJsonAsync<[Module]>(CreateAuthPolicyUrl($"{Apiurl}", [Module].ModuleId), [Module]);

mikecasas:
It was a 50/50 and I guessed wrong.

Issue #699, opened 2020-08-22
Path column fixed at 50 chars.  Increase to 256?


nohorse:
Folks adding pages 3-deep are running into the path length limitation.  I'm suggesting we increase the field length to something more reasonable like 256.  I think the spec for URLs is like 2048, so that's an option as well.

sbwalker:
DNN has used nvarchar(256) for this purpose for 18 years so I think it should be safe to assume this length should be sufficient. There is no point in defining this as 2048 as you are correct that the max Url character limit is based on all elements of the url including querystring parameters. In this case we are only focused on the url path.

PR #700, opened 2020-08-22
Create Tenant.01.00.02.02.sql

nohorse:
Proposed fixed for #699


nohorse:
yaya!  I'm officially an Oqtane contributor!

PR #714, opened 2020-08-28
added support for dynamic inclusion of global resources in _host.cshtml ( ie. global stylesheets and scripts such as those required by UI component suites )

sbwalker:


ADefWebserver:
I posted a Blog that illustrates this: [Using Radzen In Oqtane Modules](https://blazorhelpwebsite.com/ViewBlogPost/46)

PR #739, opened 2020-09-12
From razor to code behind (UI)

hishamco:
This is the first PR from many PRs that I will push to move the code from razor to code-behind, I divide the PRs to manageable chunks of small PRs so they can easily reviewed


sbwalker:
#736 was intended to be a discussion on the larger enhancements which could be tackled in the next major version. Refactoring to use code behind is actually not necessary and does not align with official Microsoft guidance so I was hoping the community would weigh in to determine if they think it is worthwhile or not. So I am holding off on merging these PRs for now.

hishamco:
Ok Shaun, but AFAIK mixing code with the markup is not preferred from the WebForms days, that is why code behind has been introduced, as well as RazorPages page models

sbwalker:
I asked Microsoft this question and this was the response: "Either component authoring style is fine. They’re mostly interchangeable. We prefer the inline style, so that’s what we use in our templates and samples. We disagree that this encourages bad coding practices any more so then if you had a code behind file. A .razor file represents a single component class so any UI logic is part of that component’s concern and makes sense to live in the same file. You shouldn’t mix business logic in your UI layer, but that can happen just as easily if you’re using a code-behind file. The Vue community uses the inline style and do just fine. But we also understand that ASP.NET devs have been taught for a long time to use a code-behind file and many prefer it, which is totally fine."

hishamco:
> A .razor file represents a single component class so any UI logic is part of that component’s concern and makes sense to live in the same file

Agree if the component logic is small, but if it's large, this will be hard to maintain, at the end the code-behind is nothing but a partial class, so at the compile time both will goes into the same file

Thanks

kevind2000:
Yes, "either is fine", preferring the single file for UI component since it's chosen. For later, there might be possibility to use XAML not Razor syntax to define UI elements to support native mobile and desktop client, even C# code directly for that matter, at that time refactoring to code-behind might be making sense - this is way down the road IMHO.

sbwalker:
Closing as we will not be migrating to code behind

sbwalker:
Closing as we will not be migrating to code behind

PR #741, opened 2020-09-12
From razor to code (Oqtane Theme)

hishamco:


sbwalker:
Closing as we will not be migrating to code behind

PR #747, opened 2020-09-16
Refactored to method to make the code more DRY.

mikecasas:
This PR is an example of what I mentioned in #736. Just wondering if you want refactorings/PRs like this.

### DIFF
--- a/Oqtane.Client/Modules/Controls/FileManager.razor
+++ b/Oqtane.Client/Modules/Controls/FileManager.razor
@@ -290,13 +290,13 @@
                 else
                 {
                     await logger.LogError("File Upload Failed For {Files}", result.Replace(",", ", "));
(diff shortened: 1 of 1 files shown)


hishamco:
I prefer to create a UI component for that, let me share what I did ...

hishamco:
I created a draft PR #750, I will try to replace all the strings that made an alert UI from the code, I think this will be better, like what we did in other components

sbwalker:
A component already exists for this functionality - the ModuleMessage component. It is used throughout the UI already- usually by calling the AddModuleMessage helper method - but also as a standard component.

mikecasas:
My original intent was just to clean up some of the repetitive code.

sbwalker:
@mikecasas its a good idea to eliminate repetitive code... you could use the existing ModuleMessage component in the FileManager

PR #750, opened 2020-09-17
Use ModuleMessage component everywhere

hishamco:


hishamco:
@sbwalker I just refactor the File Manager to you use the proposed alert component, if sound good to you I can refactor all the places to use the new component

sbwalker:
A component already exists for this functionality - the ModuleMessage component. It is used throughout the UI already- usually by calling the AddModuleMessage helper method - but also as a standard component.

hishamco:
I didn't know about it, let me check ...

hishamco:
@sbwalker it's ready from my side, just one question why you create a `Get` and `Set` methods on the component while you can get or set parameters directly from the component instance?!!

sbwalker:
@hishamco the SetModuleMessage() method is only intended to be used in one specific case - when a module developer calls the AddModuleMessage() helper which then needs to traverse to the ModuleInstance parent component and set the Message/Type properties on its ModuleMessage component. There may be a better way to implement this. Regardless, in all other cases the ModuleMessage component should be used in the standard way - by setting the parameter values directly and not using the SetModuleMessage() method. Make sense?

hishamco:
> by setting the parameter values directly and not using the SetModuleMessage() method. Make sense

I did that at the beginning but I got some warnings when I set the parameters directly, let me change it in all the places

hishamco:
Done!!

PR #752, opened 2020-09-18
Refactored to using the module message.

mikecasas:
Just wanted to change one line and see if you are ok with it. Using ModuleMessage moves the message placement from below the control to the top of the module. I'm not sure if like that?

I think it's better because the placement of messages is more consistent - meaning always at the top.

If you are ok with this then I will change all the other messages in FileManager.

### DIFF
--- a/Oqtane.Client/Modules/Controls/FileManager.razor
+++ b/Oqtane.Client/Modules/Controls/FileManager.razor
@@ -301,7 +301,7 @@
         }
         else
         {
-            _message = "<br /><div class=\"alert alert-warning\" role=\"alert\">You Have Not Selected A File To Upload</div>";
+            AddModuleMessage("You Have Not Selected A File To Upload", MessageType.Warning);
         }
     }
 


sbwalker:
the other options is that you can replace @((MarkupString) _message) with <ModuleMessage Message="@_message" Type="@_type" />

hishamco:
Seems your Shaun's comment in both PRs leads to confuse us, I already started a PR, so @mikecasas if you like to continue with this PR I will close mine or I will push my commits

sbwalker:
@mikecasas for further clarification, the AddModuleMessage helper method is only intended to be used within Modules - it allows a developer to interact with the ModuleMessage component which is included in the ModuleInstance component by default ( as opposed to forcing a Module developer to include the ModuleMessage component explicitly in each module component ). 

The FileManager component is NOT a Module - it is a generic component which can be used by any module. As a result it should not use the AddModuleMessage helper method - it should use the ModuleMessage component directly.

hishamco:
@mikecasas any  updates on this or shall I update my PR

mikecasas:
I'll close, you can update yours.

hishamco:
Ok @mikecasas

PR #756, opened 2020-09-21
Update README.md

hishamco:
Addresses https://github.com/oqtane/oqtane.framework/discussions/755


sbwalker:
I dont think this change addresses the OP's question in #755. He was asking WHY Oqtane recommends Visual Studio 2019 Preview. Adding the words "or higher" does not answer his question.

hishamco:
I was answering the question before, but adding the word "higher" may needed, coz the minimal requirement is VS 2019 Preview

PR #762, opened 2020-09-29
Localization support

hishamco:
@sbwalker this WIP PR, but it will make it easier for you to test the localization support in both client & server. Just few commits and you will be able to see localization in action with the current Oqtane project, then I will move forward to write some notes for what the changes are required to make the module localizable, after that I will add server-side localization suuport

Regarding the supported culture as you mentioned earlier it should come from `appsettings.json` or database, let us start with `appsettings.json` then nothing will prevent us to move to database if it's required


hishamco:
@sbwalker at this point you can test both:

1. Oqtane.Client

1.1 Add "Resources" folder including your resources
1.2 Add supported culture on Oqtane.Server using `LocalizationSettings.SupportedCultures.Add()`

1. Oqtane.Blogs.Client

1.1 Add "Resources" folder including your resources
1.2 Set ResourceNamespace attribute with `Oqtane.Blogs.Client`

hishamco:
**Note:**

Blazor WebAssembly is looking for the resources and display the appropriate values with the respect current culture, no supported culture is there AFAIK, but we may search if we can fallback to default culture if the culture is not supported in Oqtane

hishamco:
Again, seems Server-Side Blazor doesn't support all the localization features, as well as the culture can be set using one of the following approaches:

- Cookies

- Provide UI to choose the culture

Setting the localization is the easiest and simplest option, but we should have a component that may need to add later to allow users to switch between the supported cultures

hishamco:
@sbwalker at this point you can test server-side localization too:

3. Oqtane.Server

3.1 Setup default culture by settings `LocalizationSettings.DefaultCulture`
3.2 Add supported culture on Oqtane.Server using `LocalizationSettings.SupportedCultures.Add()`

hishamco:
The PR is now almost complete according the requirements that we are discussed https://github.com/oqtane/oqtane.framework/discussions/749

I may polish this PR if it's required, but it's good so far

Oqtane.Server/Infrastructure/Localization/LocalizationSettings.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections.Generic;
+using Oqtane.Shared;
+
+namespace Oqtane.Infrastructure.Localization
+{
+    public static class LocalizationSettings
```
hishamco, line 6: It would be nice if we avoid `static` and use options pattern, but the issue is we need a way to read a settings from within the extension methods

hishamco:
No it's the time to move foreword ;)

PR #767, opened 2020-09-30
add default class 'container' for div in pane to avoid content squishing

PoisnFang:
Currently the Modules look a little squished like [This](https://i.imgur.com/TRJROTQ.png) because of  [This](https://i.imgur.com/JyXc1e4.png). This PR solves this by adding [This](https://i.imgur.com/MoEDxEM.png) so that it looks like [This](https://i.imgur.com/iHMuwa2.png)

### DIFF
--- a/Oqtane.Client/UI/Pane.razor
+++ b/Oqtane.Client/UI/Pane.razor
@@ -1,18 +1,23 @@
-﻿@using Microsoft.AspNetCore.Components.Rendering 
+﻿@using Microsoft.AspNetCore.Components.Rendering
 @namespace Oqtane.UI
 @inject IUserService UserService
 @inject IModuleService ModuleService
 @inject IModuleDefinitionService ModuleDefinitionService
 
-<div class="@_paneadminborder">
-    @if (_panetitle != "")
-    {
+@if (_useadminborder)
+{
+    <div class="@_paneadminborder">
         @((MarkupString)_panetitle)
-    }
+        @DynamicComponent
+    </div>
+}
+else
+{
     @DynamicComponent
-</div>
+}
 
 @code {
+    private bool _useadminborder = false;
     private string _paneadminborder = "container";
     private string _panetitle = "";
 
@@ -26,8 +31,9 @@
 
     protected override void OnParametersSet()
     {
-        if (PageState.EditMode && UserSecurity.IsAuthorized(PageState.User,PermissionNames.Edit, PageState.Page.Permissions) && Name != Constants.AdminPane)
+        if (PageState.EditMode && UserSecurity.IsAuthorized(PageState.User, PermissionNames.Edit, PageState.Page.Permissions) && Name != Constants.AdminPane)
         {
+            _useadminborder = true;
             _paneadminborder = "app-pane-admin-border";
             _panetitle = "<div class=\"app-pane-admin-title\">" + Name + " Pane</div>";
         }
@@ -47,8 +53,8 @@
                     if (module != null && !module.IsDeleted)
                     {
                         var typename = module.ModuleType;
-                        // check for core module actions component
-                        if (Constants.DefaultModuleActions.Contains(PageState.Action))
+                    // check for core module actions component
+                    if (Constants.DefaultModuleActions.Contains(PageState.Action))
                         {
                             typename = Constants.DefaultModuleActionsTemplate.Replace(Constants.ActionToken, PageState.Action);
                         }
@@ -94,8 +100,8 @@
                         }
                         else
                         {
-                            // module control does not exist with name specified
-                        }
+                        // module control does not exist with name specified
+                    }
                     }
                 }
             }
@@ -106,8 +112,8 @@
                     Module module = PageState.Modules.FirstOrDefault(item => item.ModuleId == PageState.ModuleId);
                     if (module != null && module.Pane.ToLower() == Name.ToLower() && !module.IsDeleted)
                     {
-                        // check if user is authorized to view module
-                        if (UserSecurity.IsAuthorized(PageState.User, PermissionNames.View, module.Permissions))
+                    // check if user is authorized to view module
+                    if (UserSecurity.IsAuthorized(PageState.User, PermissionNames.View, module.Permissions))
                         {
                             CreateComponent(builder, module);
                         }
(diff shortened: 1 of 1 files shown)


sbwalker:
If you look back through the history of the Pane.razor component you will notice that there has been a lot of churn related to the class property. It was initially undefined, then set to "container", then rolled back... and now this PR adds it again. All of these changes have happened because different people have different opinions on what the default behavior should be. Ultimately when it comes to design elements such as this, they should not be hard-coded in the core framework - they should be specified in themes. And if there are limitations to setting this class in themes, or the general community consensus is that the behavior in the default theme should be different, then that is where this modification should be done.

PoisnFang:
Sounds good. Then in the Oqtane theme could we set it to container? Several of the child divs use container, 
which is why when you add it the content expands, because that is the design of bootstrap container.  
Having div containers inside a div that has a blank class completely defeats the purpose of using container.

sbwalker:
@PoisnFang the bigger question is if the Pane component should even include a div wrapper in the default scenario. Typically when you create a theme you would include your specific div layout and probably do not want any extra markup injected from the framework. Correct?

PoisnFang:
@sbwalker That is a good question. Currently the div wrapper solely exists in order to add  `app-pane-admin-border`. Should be customizable in a theme?

sbwalker:
The app-pane-admin-border div is only needed if a user is logged in and in edit mode. Otherwise it is not needed.

PoisnFang:
so are you suggesting dynamically adding in the div wrapper when this condition is met?

```csharp
        if (PageState.EditMode && UserSecurity.IsAuthorized(PageState.User, PermissionNames.Edit, PageState.Page.Permissions) && Name != Constants.AdminPane)
        {
            _paneadminborder = "app-pane-admin-border";
            _panetitle = "<div class=\"app-pane-admin-title\">" + Name + " Pane</div>";
        }
```

sbwalker:
correct

PoisnFang:
Ok I can look into making that happen. Should I use something like an if statement or the RenderTreeBuilder?

hishamco:
I disagree to control or add styles and css classes from within the code, I will revise the issue and write down my thoughts

sbwalker:
@PoisnFang I would just use @if in the markup

sbwalker:
@hishamco if you re-read this thread I think you will notice that we are talking about removing CSS classes, styles, and markup from the code - not adding them.

hishamco:
> I think you will notice that we are talking about removing CSS classes, styles, and markup from the code - not adding them.

That's why I said I will revise the issue ;) If the aim is to remove them I'm totally agree, just go for it

thabaum:
There are many places I think need removed classes to get desired results.  I had to remove many to get 100% width and the "container" class could of been OK to get that result sorta that was a regression.

I see problems here... and the problem starts prior to pane and after too in order to get desired results.

I have a solution I was going to propose in the near future I am hoping when I get time... not sure if it will be accepted but I think it will because it is a "cure all" approach.

Here is what I am thinking....

Anywhere "Oqtane" declares ANY classes for styles it is a class loaded from the db.  and Anywhere a class needs added to the framework again it was going to be from the db.

I was going to introduce a "non-breaking" PR in order to keep "container" class from breaking any existing themes but then I seen it was regressed instead...

Still big issue and always will be until properly handled IMO.  

My solution was going to move the "Container" class to the database and then it can be changed without breaking anything for v1.

@chlupac was correct saying it should not be in the middle of a "row"

This is layed out like a table... instead of a framework that can be modified as desired.

To get my results of 100% width I had to remove a number of classes from the framework not just that one "container" then everything was beautiful.  However nothing could get selected.

I was going to leave the container class as a default class in the db that could be then changed.  But since it turned into a tough topic with lots of people issues one way or the other...

My proposal is to make ALL the selectors available in a Admin component that the framework uses so that you can add/change/remove them as desired.

Sound good?  I thought a lot about this and tested over the past few months while dormant the idea and I think it is the best solution.  Does this at all make sense to anyone?

sbwalker:
@thabaum I appreciate the thought you put into the solution you described above. And I apologize if I do not have a full understanding of it, but from a high level perspective I am not in favor of embedding styles in code or in a database. In my opinion the framework should embrace standard and consistent web design/development methodologies ie. CSS overrides.

PoisnFang:
@sbwalker My latest commit should fix it

thabaum:
>  In my opinion the framework should embrace standard and consistent web design/development methodologies ie. CSS overrides.

@sbwalker with all due respect to your opinion, isn't this against what we are shooting for then?

 `  _paneadminborder = "app-pane-admin-border";`

Shouldn't this just be a class inserted with no option to remove then? 

And the component or theme handling the css styles and selectors?   Or is this a special case for admin component pages?

The css would probably be `app-pane-border` and this class utilizing a CSS file from within the component to style the page if desired same as any other module or component a mod developer would create?

sbwalker:
@thabaum let me elaborate on what I mean by the framework should embrace standard and consistent web design/development methodologies ie. CSS overrides....

Web applications have a user interface. That user interface is comprised of many elements which are required in order to provide a cohesive user experience. These required elements will have a default appearance but there is often a need to allow developers/designers to customize their appearance. The standard way to allow this is via CSS overrides. And in order to allow the CSS overrides to be targeted appropriately it is important to include unique classes at a granular level ( ie. to avoid the CSS override from affecting other areas of the UI ).

So if we consider your example above, there is a required element which is injected by Oqtane for displaying a border around a pane when a user is in Edit mode. This border is a visual cue of where the various panes in the theme are located. The element has a CSS class of "app-pane-admin-border". And if you look in the app.css you will notice that app-pane-admin-border has a default implementation:

```
.app-pane-admin-border {
    width: 100%;
    border-width: 1px;
    border-style: dashed;
    border-color: gray;
}
```
If you want to customize the characteristics of this required element, you should include include an override for the app-pane-admin-border class. This override should be done in a CSS file which is part of your Theme. 

This approach follows standard web UI styling patterns which most developers/designers are familiar with. As soon as you deviate from this standard approach and start introducing switches in the code or database configuration it creates a blocker as developers/designers will be forced to adapt to a non-standard pattern.

sbwalker:
On a related note, there is also a scenario where a semi-technical user may want to customize the user interface. They may not have the skillset necessary to create or modify a custom Theme. However they may be familiar enough with CSS that they could add a CSS override. In DNN we had the concept of a CSS override at the Site level. Basically this was a unique CSS file stored on the file system for each "portal". By default it was empty. However a user could include custom CSS overrides in that file, and they would override styles includes by the app or theme. Perhaps Oqtane needs this feature as well.

PR #770, opened 2020-10-01
Allows page to find Custom Index page in Module from Actions Property

PoisnFang:
added after this discussion post was made: #765
This if statement just checks if the route is inside a Module, we do not need to check that anymore because the Index Page should always be displayed if we are not in the Module. ` if (module.ModuleId == moduleid && action != "")`  Which is why we set this `var action = Constants.DefaultAction;` Which basically the same thing that this was doing: `else{module.ModuleType = typename.Replace(Constants.ActionToken, Constants.DefaultAction);}`
I did test this with an External Module and the internal HtmlText Module, I did not find any errors.

### DIFF
--- a/Oqtane.Client/UI/SiteRouter.razor
+++ b/Oqtane.Client/UI/SiteRouter.razor
@@ -76,7 +76,7 @@
         User user = null;
         List<Module> modules;
         var moduleid = -1;
-        var action = string.Empty;
+        var action = Constants.DefaultAction;
         var urlparameters = string.Empty;
         var editmode = false;
         var reload = Reload.None;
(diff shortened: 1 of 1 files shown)


sbwalker:
see comments in #765

sbwalker:
@PoisnFang I think there may be an issue with this PR. Oqtane supports 3 scenarios:

1. ModuleId = "" - displays all modules ( displays Index component )
2. ModuleId != "" && Action = "" - displays only the module specified by ModuleId ( displays Index component )
3. ModuleId != "" && Action != "" - displays only the module specified by ModuleId ( displays component corresponding to Action )

I am concerned that scenario 2 is no longer supported in this PR

PoisnFang:
@sbwalker What would be an example url for scenario 2 that I can use as reference?

sbwalker:
http://localhost:44357/*/1

PoisnFang:
Looks like the change, was fine. Just need to update the HtmlText Module to not use the admin theme in a different PR?

sbwalker:
I am fixing the issue now. It is not related to the HtmlText module.

PR #775, opened 2020-10-01
Blazor WebAssembly localization test

hishamco:
@sbwalker please try this PR to your end to make sure that localization is work at your end

Please add Arabic, French and Spanish languages to your browser and try to switch between then while the application is running, hit a refresh after each change and you should see "Hello Oqtane" on  every culture

### DIFF
--- a/Oqtane.Client/Resources/Oqtane/Themes/OqtaneTheme/Default.ar.resx
+++ b/Oqtane.Client/Resources/Oqtane/Themes/OqtaneTheme/Default.ar.resx
@@ -1,0 +1,123 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<root>
+  <!-- 
+    Microsoft ResX Schema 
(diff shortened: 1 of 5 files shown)


hishamco:
After it works with you can basically change the runtime to  "Server" and set the browser culture to the default culture that has been set in `appsettings.json` then run the app

Please let me know if you face any issue or something is not work

hishamco:
@sbwalker did you test this PR? Don't forget to close it after you ensure if the localization is working at your side

hishamco:
@sbwalker seems this take so long, if the localization is already worked for you, I will close this PR

sbwalker:
I will do some localization testing today. I was hoping others in the community would also assist with localization but so far nobody has volunteered

hishamco:
No problem Shaun I can handle all the localization stuff as much as I can ;)

PR #783, opened 2020-10-06
Refactor to eliminate repetitive code.

mikecasas:
Simple improvement to keep the code DRY.

### DIFF
--- a/Oqtane.Client/Services/ServiceBase.cs
+++ b/Oqtane.Client/Services/ServiceBase.cs
@@ -173,13 +173,15 @@ namespace Oqtane.Services
         // add entityid parameter to url for custom authorization policy
         public string CreateAuthorizationPolicyUrl(string url, int entityId)
         {
+            string qs = "entityid=" + entityId.ToString();
+
             if (url.Contains("?"))
             {
-                return url + "&entityid=" + entityId.ToString();
+                return url + "&" + qs;
             }
             else
             {
-                return url + "?entityid=" + entityId.ToString();
+                return url + "?" + qs;
             }
         }
 


Oqtane.Client/Services/ServiceBase.cs
```diff
@@ -173,13 +173,15 @@ public string CreateApiUrl(Alias alias, string serviceName)
         // add entityid parameter to url for custom authorization policy
         public string CreateAuthorizationPolicyUrl(string url, int entityId)
         {
+            string qs = "entityid=" + entityId.ToString();
```
hishamco, line 176: What about

```csharp
return url.Contains("?") ? "&" : "?" + "entityid=" + entityId.ToString(); 
```

mikecasas: I go for the simple approach.  I know what the code you pasted does, but IMO it's so complex for a one-liner.

hishamco: Not that complex ;) coz the first letter depends on whether the url contains query string or not

jimspillane: Both ways are fine for me.  The boss indicated a preference for the longer form.
https://github.com/oqtane/oqtane.framework/pull/155#issuecomment-545851383

hishamco: I prefer inline condition in such case but if the condition is complex or has nested conditions I may prefer if block for readability

mikecasas: Thanks @jimspillane I thought I had read that somewhere. Plus at this point I don't want to change a lot of the code, meaning if it's an 'if block', then I will leave it as an if block.

hishamco: This take too long for a tiny change ;) could you please give a meaningful name instead of `qs` and I'm fine with that

PR #785, opened 2020-10-07
Update SiteRouter.razor

markdav-is:
not every class that inherits from ModuleBase implements iModule.  So it's possible to add a module without a ModuleDefinition.  In that case the default action should be Index.

### DIFF
--- a/Oqtane.Client/UI/SiteRouter.razor
+++ b/Oqtane.Client/UI/SiteRouter.razor
@@ -455,7 +455,7 @@
                 }
 
                 // handle default action
-                if (action == Constants.DefaultAction && !string.IsNullOrEmpty(module.ModuleDefinition.DefaultAction))
+                if (action == Constants.DefaultAction && !string.IsNullOrEmpty(module.ModuleDefinition?.DefaultAction))
                 {
                     action = module.ModuleDefinition.DefaultAction;
                 }


sbwalker:
Can you please explain the error or issue you were encountering? When I review the code in the ProcessModules() method there are many references to module.ModuleDefinition - so I do not understand why it would need to handle null in this one instance and not the others? Module is linked to ModuleDefinition via the ModuleDefinitionName - not by the existence of the IModule interface - see Oqtane.Server\Controllers\ModuleController.cs. So if your ModuleDefinition is null I think it is because you must have broken the linkage between Module and ModuleDefinition in the database - probably by renaming your module?

markdav-is:
Sure thing!  We have a multi-module called AkaMod.  It's a solution with many module namespaces defined but there is only one moduleinfo.cs and only one entry in the ModuleDefinition table.  Oqtane allow us to add modules to the pages by namespace and when i select a page that has one of those modules (sub modules?) it blows up with the lastest additions where it's looking for the default action.  I'm currently going thru and adding the IModule interface to each and every namespace (sub-module?) in the how that that will resolve the problem.  For other folks in this situation, the next rev might be a braking change.   This proposed change will help default those namespaces that are missing a module definition to still run and default to index.razor.  I'm sure I'm bending the definition of a module, but I always error on the side of more resilient code that can deal with missing pieces. I hope I was able to describe this properly!

sbwalker:
Thank you for the explanation. I would like to dig into this a bit further as I am not familiar the pattern you are using to develop modules. Using an unknown pattern represents a risk because new features are developed/tested to support only known patterns. 

I understand that you are developing multiple modules within the same project - which is fine.  In fact, the Oqtane.Client framework project contains many different modules within the same project - they are organized into separate folders under the /Admin folder and each module has its own unique namespace. And most of these modules do not implement the IModule interface. On startup the framework iterates through the assemblies and looks for all instances of IModuleControl and organizes them based on namespace. It also uses namespaces to determine if a module implements IModule - however this is optional.  

For example:

Oqtane.Client\Modules\Admin\Roles contains razor components with @namespace Oqtane.Modules.Admin.Roles
Oqtane.Client\Modules\Admin\Sites contains razor components with @namespace Oqtane.Modules.Admin.Sites
Neither of these modules implement IModule

If you look at Oqtane.Server\Controllers\ModuleController.cs you will see that the Get() method populates a collection of module instances and it eager loads the ModuleDefinition property on the Module object. It does this because the framework relies on the metadata in the ModuleDefinition object for a variety of functionality. In your case it appears that ModuleDefinition is null - which is a red flag as it means the framework is not able to locate the metadata for the module.

Can you please describe the pattern you are using in more detail... and perhaps why you are using the pattern?

sbwalker:
@markdav-is just to clarify a few more items... you should NOT need to add the IModule interface to each namespace - the interface is optional. And the change for DefaultAction should NOT be a breaking change. It seems to be causing an error in your implementation based on the pattern you are using to develop modules - however it is important to dig into this further as the fact that a Module object is not being associated to a ModuleDefinition object is a much larger problem than just the issue you identified in the router. Internally the association from a Module instance to a ModuleDefinition is mandatory - the ModuleController should actually be raising an exception when it cannot find the associated ModuleDefinition.

markdav-is:
Thanks so much for the additional comments. I'm learning a lot about the structure of modules.  It could be that my database is a little wonky.  I've been focused on some other issues this week, but hope to dig into this deeper as it's a breaking change for our team.  we are currently stuck on 1.0.4 until i can resolve this.  It's really nice to know that I don't need to create Imodule defs for all the namespaces.  I still plan on making them for namespaces with alternate default actions.

sbwalker:
@markdav-is yes I think there may be an integrity issue in your database if you encountering an issue with this new IModule property

markdav-is:
I'm researching when the database rows get written as I suspect it would happen when the module was installed.  Now that we've added so many modules,  just updating the DLLs,  we might need to re-install the module(s) to get those records populated.

PR #788, opened 2020-10-08
Changes for .NET 5

sbwalker:


sbwalker:
Install .NET 5 RC1 from here https://dotnet.microsoft.com/download/dotnet/5.0

hishamco:
I will try this today and come back to you

hishamco:
Seems my VS complains about `Oqtane.Client` I will try with VS Code

sbwalker:
Did you follow the  instructions for .NET 5 RC1 and install Visual Studio 2019 (v16.8, Preview 3). If you do not use that version of Visual Studio ( or higher ) it will not work. I am using Version 16.8.0 Preview 3.2 and it works fine.

hishamco:
>  it will not work

Exactly, I have a little bit old version `16.4.0`, I will have a try now with VS Code ...

hishamco:
Woohoo it's working with VS Code, also the localization working too

LGTM :shipit:

hishamco:
@sbwalker can we merge this?

hishamco:
I'm still suggesting to create a `dev` branch and leave the `master` to the stable release

We could also tag and create branches for older release, if you need a help or you don't have time for this, I can do it if you add me to the Oqtane team or give some permission

hishamco:
@sbwalker I think you mentioned few days ago that .NET Core 5.0 stuff will merge into `master` but we will continue to have .NET Core 3.1 in `dev` branch, so is this really happen, coz merging this in `master` will enforce everyone to use .NET Core 5.0 RC

sbwalker:
@hishamco you are correct that I did suggest a branching strategy in that other... however after thinking more about the fact that there are only 3 weeks until .NET 5 is released on Nov 10, and the fact that main already contains a bunch of other changes for Localization which are intended to be part of the 2.0 release, I think we need to move to .NET 5 now so that we can rally the community and get prepared for the release. I will provide a longer explanation in the branching discussion.

hishamco:
No problem I thought you will release `1.0.5` as you said, but seems `2.0.0` on it's way ;) BTW did you see my PR on `Oqtane.Blogs` repo and test other localization features that have been added, coz I need to know what's the remaining before release `2.0.0`

PR #792, opened 2020-10-10
Introduce PlatformInfo

hishamco:
This will be useful for many scenarios:

- Keep track of Oqtane runtime

- Hold Oqtane current version

- Probably keep track of the current environment

### DIFF
--- a/Oqtane.Client/Services/Interfaces/IPlatform.cs
+++ b/Oqtane.Client/Services/Interfaces/IPlatform.cs
@@ -1,0 +1,11 @@
+﻿using Oqtane.UI;
+
+namespace Oqtane.Services
+{
+    public interface IPlatform
+    {
+        Runtime Runtime { get; }
+
+        string Version { get; }
+    }
+}
--- a/Oqtane.Client/UI/PageState.cs
+++ b/Oqtane.Client/UI/PageState.cs
@@ -19,6 +19,8 @@ namespace Oqtane.UI
         public string Action { get; set; }
         public bool EditMode { get; set; }
         public DateTime LastSyncDate { get; set; }
+
+        [Obsolete("This property is deprecated and will be removed in the upcoming major release, try to use IPlatformRuntime instead.")]
         public Runtime Runtime { get; set; }
     }
 }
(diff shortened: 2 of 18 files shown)


hishamco:
After this PR no need to change the current version in `Constants.Version` after every release, also the `Runtime` in the `PageState` is no longer needed

Oqtane.Client/UI/PageState.cs
```diff
@@ -19,6 +19,8 @@ public class PageState
         public string Action { get; set; }
         public bool EditMode { get; set; }
         public DateTime LastSyncDate { get; set; }
+
+        [Obsolete("This property is deprecated and will be removed in the upcoming major release, try to use IPlatformRuntime instead.")]
```
hishamco, line 23: I just mark this property as `Obsolete` for backward compatibility

hishamco:
@sbwalker just one question if this PR looks good for you. How the `Runtime` has been passed from the `Oqtane.Server` to `Oqtane.Client`? coz the `IPlatform` need to be registered in the `Oqtane.Client` too

Oqtane.Client/Services/Interfaces/IPlatform.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Oqtane.UI;
+
+namespace Oqtane.Services
+{
+    public interface IPlatform
+    {
+        Runtime Runtime { get; }
+
+        string Version { get; }
```
hishamco, line 9: Version could be used as type `Version` to avoid unnecessary versions comparison

sbwalker:
I do not understand the problem this PR is trying to solve? It seems like a lot of code changes with no specific purpose?

hishamco:
As I mentioned earlier accessing `Runtime` require us to inject the `IConfiguration` and get its value also the version is a static value that need to be changed on every release

This PR introduced a platform information, so you can easily access the runtime and version and all things we may add to this class from both client and server

sbwalker:
Neither of the items you mentioned are actual "problems". The Version is a static value and it is located in a single place in the Oqtane.Shared project to make it easily accessible to module developers - in their Client, Server, and Shared projects. Runtime is part of the PageState,  so that it is also easily accessible to module developers ( ie, they don't need to figure out how to inject anything - it is immediately available to them ). This is a really important perspective in Oqtane - this is a platform, not an application - so the main focus should be on how easy it is for developers to build modules and themes.

hishamco:
I disagree about the version it could be retrieved from the current assembly instead of a static value to avoid the changes everytime. Do it once and forget ; )

Let us focus on the important thing then and please check the localization label issue I'm blocked for few days :(

PR #795, opened 2020-10-11
Organize and clean Oqtane startup classes

hishamco:
- [x]  Oqtane.Server
- [x]  Oqtane.Client

### DIFF
--- a/Oqtane.Server/Startup.cs
+++ b/Oqtane.Server/Startup.cs
@@ -1,11 +1,8 @@
 using System;
 using System.Collections;
 using System.IO;
-using System.Linq;
-using System.Net.Http;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
-using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Identity;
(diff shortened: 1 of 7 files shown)


hishamco:
No worry about the changes, tested and what I basically did is move everything into parts and make everything clean and clear. Taking the same order in consideration

Oqtane.Server/Startup.cs
```diff
@@ -180,43 +114,13 @@ public void ConfigureServices(IServiceCollection services)
             // register custom claims principal factory for role claims
             services.AddTransient<IUserClaimsPrincipalFactory<IdentityUser>, ClaimsPrincipalFactory<IdentityUser>>();
 
-            // register singleton scoped core services
             services.AddSingleton(Configuration);
```
hishamco, line 117: I don't think this line is necessary, but the reason for that is you are used `IConfigurationRoot` in `DatabaseMigration`

sbwalker:
I am VERY hesitant to merge this PR as, like I mentioned in a different PR, the startup logic is very fragile and the order of events is extremely important. There are at least 4 different installation scenarios which would need to be thoroughly regression tested to ensure they are not broken. The last time this code was "organized and cleaned up" as part of a PR, it caused almost a week of work to get things all fully working again. Since this code is all functional in its current form I would prefer to not introduce changes which could break functionality at this time. I would rather focus our time/effort on ensuring the new features which are promised for V2 ( ie. Localization ) are completely implemented.

hishamco:
No problem make this on hold, I didn't touch or change the order coz it's important as you mentioned

Regarding the localization I'm still waiting for your reply in the related issue

ADefWebserver:
This may break the functionality required to register component suits like Syncfusion. See "Register The Syncfusion License" at: https://blazorhelpwebsite.com/ViewBlogPost/47

hishamco:
@ADefWebserver this is an old PR it's replaced by #1245

hishamco:
Closed with respect of #1245

PR #796, opened 2020-10-12
Support label localization

### LINKED ISSUE #771: Support localization for Label component
As @sbwalker mentioned earlier

> @hishamco DNN has a custom Label control which handles the display of form labels and help text. The way that the label control handles localization is via an attribute named "ResourceKey". Essentially the value provided for this attribute was used as a prefix when retrieving localized values. So if ResourceKey="FirstName" then it would expect by convention that there would be localization name/value pairs for "FirstName.Text" ( the label ) and "FirstName.HelpText" ( the help text ). This made it very easy for developers to wire up localization - without having to write any code. Oqtane has a similar Label component which is used consistently throughout the admin UI. It would be great if the Label component had a parameter named "ResourceKey" which could be used for the same purpose.

Now we should think about the challenges and use cases to make this happen, hopefully I will share my thought soon ...

hishamco:
Fixes #771

### DIFF
--- a/Oqtane.Client/Modules/Controls/Label.razor
+++ b/Oqtane.Client/Modules/Controls/Label.razor
@@ -1,5 +1,8 @@
 ﻿@namespace Oqtane.Modules.Controls
 @inherits ModuleControlBase
+@using Microsoft.AspNetCore.Http
+@using Microsoft.Extensions.Localization
+@inject IHttpContextAccessor HttpContextAccessor
 
 @if (!string.IsNullOrEmpty(HelpText))
 {
@@ -26,6 +29,9 @@ else
     [Parameter]
     public string HelpText { get; set; } // optional - tooltip for this label
 
+    [Parameter]
+    public string ResourceKey { get; set; }
+
     protected override void OnParametersSet()
     {
         _openLabel = "<label";
@@ -40,5 +46,19 @@ else
         }
 
         _openLabel += ">";
+
+        if (!string.IsNullOrEmpty(ResourceKey))
+        {
+            if (ModuleState?.ModuleType != null)
+            {
+                var moduleType = Type.GetType(ModuleState.ModuleType);
+                var localizerTypeName = $"Microsoft.Extensions.Localization.IStringLocalizer`1[[{moduleType.AssemblyQualifiedName}]], Microsoft.Extensions.Localization.Abstractions";
+                var localizerType = Type.GetType(localizerTypeName);
+                var localizer = (IStringLocalizer)HttpContextAccessor.HttpContext.RequestServices.GetService(localizerType);
+                
+                ChildContent = @<text>@localizer[$"{ResourceKey}.Text"]</text>;
+                HelpText = localizer[$"{ResourceKey}.{nameof(HelpText)}"];
+            }
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


hishamco:
This show how the "Name" label is localized

Localizable Label

Oqtane.Client/Modules/Controls/Label.razor
```diff
@@ -40,5 +46,19 @@ else
...
+        if (!string.IsNullOrEmpty(ResourceKey))
+        {
+            if (ModuleState?.ModuleType != null)
+            {
+                var moduleType = Type.GetType(ModuleState.ModuleType);
+                var localizerTypeName = $"Microsoft.Extensions.Localization.IStringLocalizer`1[[{moduleType.AssemblyQualifiedName}]], Microsoft.Extensions.Localization.Abstractions";
+                var localizerType = Type.GetType(localizerTypeName);
+                var localizer = (IStringLocalizer)HttpContextAccessor.HttpContext.RequestServices.GetService(localizerType);
+                
+                ChildContent = @<text>@localizer[$"{ResourceKey}.Text"]</text>;
```
hishamco, line 59: @sbwalker I don't see any text property for a label component, but I find this line work to set the label text, please let me know your feedback about this line

sbwalker: this looks good! what would the RESX file look like to define the localized values?

hishamco: Basically as:

Name.Text => Nom

Name.HelpText => Entrez Le num do site

Where `Name` is the resource name provided in `ResourceKey` property

hishamco:
IMHO many components could be localized that way, and I'm still I see a value if we create a localization sample or localize both `Oqtane.Server` and `Oqtane.Client` I know the last one it needs time but at least we need a way to provide a way to make the users start quickly using localization without mess ;)

sbwalker:
One additional question I have related to localization is whether it is possible to create RESX files for the default language? For example, Oqtane will display the text in English by default - and currently it is embedded in the razor files - but it would be better if the English resources also came from RESX files so that it is possible to customize the terminology.

hishamco:
> it is possible to create RESX files for the default language? 

Yep we can, but no need ;) because the default behavior for the localizer is to return the string that passed in as a key

hishamco:
For labels while the actual keys something like `Name.Text` and `Name.HelpText` the locaizer will show them as its for that we may need a English resources to retrieve 'Name` and `Enter the name` instead

sbwalker:
Yeah, I am not a big fan of the default behavior of the localizer in .NET Core. It seems very brittle. For example if I add text to a razor file... and then people create alternate RESX localizations in various languages... and then I realize that I mispelled the text in the razor file so I fix the spelling mistake... it will break the localization in every related REX file because the content is the key. In every other aspect of software development you are taught to never use content as a key - a key should be an immutable value.

hishamco:
> Yeah, I am not a big fan of the default behavior of the localizer in .NET Core

Lol, I remembered the early beta's there's a suggestion to add a missing behavior, to use the same key or throw an exception and this I'm currently planned to do in my `JsonStringLocalizer` https://github.com/hishamco/My.Extensions.Localization.Json

> In every other aspect of software development you are taught to never use content as a key - a key should be an immutable value

Make sense

So, shall we merge this if looks good to you, and what about to support other components, we may consider it for future release at least. If this got merged there will be two simple things I need to finish to make a successful localization story in V2 ;)

PR #824, opened 2020-10-19
Localizable component

hishamco:
@sbwalker this tested and will make component localization more easier

### DIFF
--- a/Oqtane.Client/Modules/Controls/LocalizableComponent.cs
+++ b/Oqtane.Client/Modules/Controls/LocalizableComponent.cs
@@ -1,0 +1,34 @@
+using System;
+using Microsoft.AspNetCore.Components;
+using Microsoft.Extensions.Localization;
+using Oqtane.Shared;
+
+namespace Oqtane.Modules.Controls
+{
+    public class LocalizableComponent : ModuleControlBase
+    {
+        [Parameter]
+        public string ResourceKey { get; set; }
+
+        protected IStringLocalizer Localizer { get; private set; }
+
+        protected override void OnParametersSet()
+        {
+            if (!String.IsNullOrEmpty(ResourceKey))
+            {
+                if (ModuleState?.ModuleType != null)
+                {
+                    var moduleType = Type.GetType(ModuleState.ModuleType);
+                    var localizerTypeName = $"Microsoft.Extensions.Localization.IStringLocalizer`1[[{moduleType.AssemblyQualifiedName}]], Microsoft.Extensions.Localization.Abstractions";
+                    var localizerType = Type.GetType(localizerTypeName);
+
+                    // HACK: Use ServiceActivator instead of injecting IHttpContextAccessor, because HttpContext throws NRE in WebAssembly runtime
+                    using (var scope = ServiceActivator.GetScope())
+                    {
+                        Localizer = (IStringLocalizer)scope.ServiceProvider.GetService(localizerType);
+                    }
+                }
+            }
+        }
+    }
+}
--- a/Oqtane.Client/Modules/Controls/LocalizableComponent.cs
+++ b/Oqtane.Client/Modules/Controls/LocalizableComponent.cs
@@ -7,10 +7,17 @@ namespace Oqtane.Modules.Controls
 {
     public class LocalizableComponent : ModuleControlBase
     {
+        private IStringLocalizer _localizer;
+
         [Parameter]
         public string ResourceKey { get; set; }
 
-        protected IStringLocalizer Localizer { get; private set; }
+        protected string Localize(string name)
+        {
+            var key = $"{ResourceKey}.{name}";
+
+            return _localizer?[key] ?? key;
+        }
 
         protected override void OnParametersSet()
         {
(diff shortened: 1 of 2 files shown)


Oqtane.Client/Modules/Controls/LocalizableComponent.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace Oqtane.Modules.Controls
+{
+    public class LocalizableComponent : ModuleControlBase
+    {
+        private IStringLocalizer _localizer;
+
+        [Parameter]
+        public string ResourceKey { get; set; }
+
+        protected string Localize(string name)
```
hishamco, line 15: Just a little note here, I had seen `T` is common name for translation, so can we use it instead?

sbwalker: can you provide more detail on what you are asking?

hishamco: Just if we can rename `Localize` to `T` or just forget it

sbwalker: My preference is to use Localize() as its much more explicit/intuitive - I would have no idea what T means.

hishamco:
You can merge after the upcoming commit because there's a case needed where the resource key is not set

Oqtane.Client/Modules/Controls/LocalizableComponent.cs
hishamco: T for Translate, I'm with you, seems I need to add a check in every child component, cause we have 3 cases:

1. Localizable component and key is exist

2. Localizable component and key isn't exist

3. Non localizable component

I will do a final test before merge

hishamco: Localization

Everything works as except, regarding the above cases the result should be:

1. The translated text

2. `{ResourceKey}.{Property}`

3. Use the property value

hishamco:
@sbwalker it's ready to merge

Oqtane.Client/Modules/Controls/LocalizableComponent.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        protected override void OnParametersSet()
+        {
+            if (!String.IsNullOrEmpty(ResourceKey))
+            {
+                if (ModuleState?.ModuleType != null)
+                {
+                    var moduleType = Type.GetType(ModuleState.ModuleType);
+                    var localizerTypeName = $"Microsoft.Extensions.Localization.IStringLocalizer`1[[{moduleType.AssemblyQualifiedName}]], Microsoft.Extensions.Localization.Abstractions";
```
pranavkm, line 36: Are you able to use https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.localization.istringlocalizerfactory.create?view=dotnet-plat-ext-3.1#Microsoft_Extensions_Localization_IStringLocalizerFactory_Create_System_Type_ instead?

Oqtane.Client/Modules/Controls/LocalizableComponent.cs
```diff
@@ -0,0 +1,54 @@
...
+                    var localizerType = Type.GetType(localizerTypeName);
+
+                    // HACK: Use ServiceActivator instead of injecting IHttpContextAccessor, because HttpContext throws NRE in WebAssembly runtime
```
pranavkm, line 39: Using `IHttpContextAccessor` in a Blazor Server app could expose you to security vulnerabilities. Components do not guarantee they execute in the context of a user's request (think a timer that's updated in the background or a chat component that causes other instances of chat components to update when you post a message). You should not use it in general since it's an anti-pattern, but definitely not in a server app.

hishamco: Thanks @pranavkm for pointing to this

Oqtane.Client/Modules/Controls/LocalizableComponent.cs
hishamco: Oooh, how did I forgot that we already have `IStringLocalizerFactory.Create()` ;) I will try to simplify this soon

Thanks again

Oqtane.Client/Modules/Controls/LocalizableComponent.cs
sbwalker: @hishamco are you going to be able to take a look at this? Or should I make an attempt at it?

hishamco: @sbwalker if you meant `IStringLocalizerFactory` I already started today on this on, but one point I'm still think about, the factory will create the `IStringLocalizer` but my aim in the PR was to retrieve it from the DI. I will update you on this soon, before releasing `2.0.0`

Oqtane.Client/Modules/Controls/LocalizableComponent.cs
hishamco: I think the issue with `IStringLocalizerFactory.Create()` will create a new instance of the localizer at that point of time, but the aim to get the `IStringLocalizer` that already registered in the DI

Oqtane.Client/Modules/Controls/LocalizableComponent.cs
hishamco: Seems something broken in localization, I will try to figure out what happen into the latest few PRs

PR #836, opened 2020-10-20
Factor out Policy Names

TonyValenti:
Factor out Policy Names
Change AppDomain to AppContext (AppDomain just passes it through to AppContext and AppDomain is basically considered legacy).

### DIFF
--- a/Oqtane.Shared/Shared/PolicyNames.cs
+++ b/Oqtane.Shared/Shared/PolicyNames.cs
@@ -1,10 +1,19 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Text;
 
-namespace Oqtane.Shared {
-    public class PolicyNames {
+namespace Oqtane.Shared
+{
+    public class PolicyNames
+    {
+        public const string ViewPage = "ViewPage";
+        public const string EditPage = "EditPage";
+
         public const string ViewModule = "ViewModule";
         public const string EditModule = "EditModule";
+
+        public const string ViewFolder = "ViewFolder";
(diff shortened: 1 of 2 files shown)


Oqtane.Shared/Shared/PolicyNames.cs
```diff
@@ -1,10 +1,19 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Text;
 
-namespace Oqtane.Shared {
-    public class PolicyNames {
+namespace Oqtane.Shared
+{
+    public class PolicyNames
```
hishamco, line 7: If @sbwalker agree with this change I'd like to use `static readonly` field instead of `const`

TonyValenti: You can't use ```static readonly``` in attributes though.  Ie.:

```
        [HttpGet("{id}")]
        [Authorize(Policy = PolicyNames.ViewModule)]
        public List<HtmlTextInfo> Get(int id)
        {
            var list = new List<HtmlTextInfo>();
            try
            {
                HtmlTextInfo htmlText = null;
                if (_entityId == id)
                {
                    htmlText = _htmlText.GetHtmlText(id);
                    list.Add(htmlText);
                }
            }
            catch (Exception ex)
            {
                _logger.Log(LogLevel.Error, this, LogFunction.Read, ex, "Get Error {Error}", ex.Message);
                throw;
            }
            return list;
        }
```

hishamco: You are right I didn't notice they are using there

PR #839, opened 2020-10-20
Fix issue with ActionLink localization

hishamco:
Roles


hishamco:
@sbwalker ready to go ...

Please check it without localization resources, just run the page

PR #840, opened 2020-10-20
Factored out DataDirectory string replacements.

TonyValenti:
There were a number of places where the ```DataDirectory``` variable was being retrieved and then was having string replacements done.  I factored out these methods into their own classes and then updated the code so that everywhere calls the same methods.

### DIFF
--- a/Oqtane.Server/Extensions/WebHostBuilderExtensions.cs
+++ b/Oqtane.Server/Extensions/WebHostBuilderExtensions.cs
@@ -1,6 +1,7 @@
-﻿using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 using Oqtane.Infrastructure;
+using Oqtane.Shared;
 
 namespace Microsoft.AspNetCore.Hosting
 {
@@ -12,7 +13,7 @@ namespace Microsoft.AspNetCore.Hosting
             {
                 var config = context.Configuration;
 
-                services.Configure<LocalizationOptions>(config.GetSection("Localization"));
+                services.Configure<LocalizationOptions>(config.GetSection(SettingKeys.LocalizationSection));
                 services.AddSingleton(ctx => ctx.GetService<IOptions<LocalizationOptions>>().Value);
                 services.AddTransient<ILocalizationManager, LocalizationManager>();
             });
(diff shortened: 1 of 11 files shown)


Oqtane.Shared/Configuration/ConnectionString.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Oqtane.Configuration
+{
+    public static class ConnectionString
```
hishamco, line 9: Static classes, noooooo

hishamco:
Please @sbwalker don't merge this PR, there is a better way to make this happen without `static` classes

Thanks @TonyValenti for your contribution, but we should do it right way

TonyValenti:
@hishamco What is your recommendation?

Oqtane.Server/Extensions/WebHostBuilderExtensions.cs
```diff
@@ -12,7 +13,7 @@ public static IWebHostBuilder ConfigureLocalizationSettings(this IWebHostBuilder
             {
                 var config = context.Configuration;
 
-                services.Configure<LocalizationOptions>(config.GetSection("Localization"));
+                services.Configure<LocalizationOptions>(config.GetSection(SettingKeys.LocalizationSection));
```
hishamco, line 16: Seems this go beyond DataDirectory ;)

hishamco:
> @hishamco What is your recommendation?

Perhaps using options pattern for all things related to `appsettings.json`

But my advice as always @sbwalker remind me, let us think about V2 features and backlog, coz such PRs needs a review and perhaps breaking things, I already did one in the past ;) so please let us focus on V2, if we finish we can refactor and enhance the code

Thanks

Oqtane.Server/Extensions/WebHostBuilderExtensions.cs
TonyValenti: When I create a PR, it wraps in things in from subsequent commits as well.  You'll notice that is from the next commit.

sbwalker:
I am not going to merge this PR as it touches DatabaseManager - which has at least 4 distinct scenarios related to installation/upgrade which would be affected and I am fairly certain that not all of these scenarios were tested with these changes.

hishamco:
Agree with @sbwalker we may have a look after `2.0.0` with unit and functional tests provided to make sure nothing is broken

PR #857, opened 2020-11-03
fixed compatibility issue in .NET5/WebAssembly where assemblies were not being loaded into the default AppDomain, optimized service registration on WebAssembly, fixed spelling mistake for satellite assemblies constant and fixed issue in LocalizableComponent

sbwalker:


Oqtane.Shared/Shared/Constants.cs
```diff
@@ -68,7 +68,7 @@ public class Constants {
         };
         public static readonly string[] InvalidFileNameEndingChars = { ".", " " };
 
-        public static readonly string StalliteAssemblyExtension = ".resources.dll";
+        public static readonly string SatelliteAssemblyExtension = ".resources.dll";
```
hishamco, line 71: Good catch

Oqtane.Client/Modules/Controls/LocalizableComponent.cs
```diff
@@ -18,7 +18,7 @@ protected string Localize(string name)
         {
             if (!IsLocalizable)
             {
-                return null;
+                return name;
```
hishamco, line 21: Why?

sbwalker: For modules that have already been created for Oqtane 1.x they may have used the ActionDialog control. When trying to run these modules on Oqtane 2.0 (.NET5) the modules run fine however when the ActionDialog modal popup is displayed, the buttons do not have any text. I stepped through the code and it is because the Localize() method is returning null. This appears to be  a breaking change as it affects backward compatibility. I am not sure if the solution I checked in is valid - if not, then we will need some other type of solution to deal with this issue. It does seem to make sense to me that if a component is not localizable then the method should the text which was passed in?

hishamco: Aha, I did that because you mentioned this in the PR about hating default values in localization ;)

PR #859, opened 2020-11-04
optimize performance when running on WebAssembly by caching assembly payload that needs to be served to new clients

sbwalker:


hishamco:
Great!!

PR #871, opened 2020-11-09
Removed comment and added defensive logic in the eent that the moduletype is not valid. Also changed default behavior to display the name if the key is missing. Will need an option in the future to enable the display of missing keys.

sbwalker:


sbwalker:
@hishamco can you please review before I merge? The one significant change is that for the benefit of the user experience to end users I think the default name should be displayed in the UI if a localized translation has not been defined. I think in the future we need an option for ShowMissingResourceKeys which can be enabled by developers/translators to show any cases where a localized translation is missing.

hishamco:
Seems you merged before I reply ;) no problem I will check

hishamco:
IMHO setting this property by developer for every component is a headache, I have a better idea from UI/UX perspective, we could show the missing localization strings to the developers something like logs which is suitable for devs and not the end users

sbwalker:
I was not referring to a parameter on the localized component - I was referring to an application setting. DNN had a global config option to ShowMissingResourceKeys.

hishamco:
I see, sound good, I will try to implement this

sbwalker:
We probably don't need it for 2.0.0. It will be a bit tricky as it is needed in the Client project - so you cant use appsettings or other simple mechanisms

hishamco:
> We probably don't need it for 2.0.0

Sure

PR #938, opened 2020-11-20
missing IStringLocalizer<Index> causing compilation error

sbwalker:


hishamco:
Oops seems I miss one, thanks for the fix

PR #950, opened 2020-11-25
Fix localization issue in ActionLink

### LINKED ISSUE #945: Localization
@hishamco after your recent changes to implement localization throughout the Admin UI I am seeing:

image

My understanding was that if someone is not using localization, then the user interface should be displayed in the fallback default language (English). Is that your understanding as well?

hishamco:
Fixes #945


hishamco:
Profile

PR #952, opened 2020-11-26
Introduce IdentityStartup

hishamco:
Address https://github.com/oqtane/oqtane.framework/discussions/903

- Ability to toggle Identity service on / off

- Ability to replace the default Identity service implementation

### DIFF
--- a/Oqtane.Server/Startup.cs
+++ b/Oqtane.Server/Startup.cs
@@ -140,7 +140,7 @@ namespace Oqtane
                 ));
             services.AddDbContext<TenantDBContext>(options => { });
 
-            if (FeatureManager.IsEnabledAsync("Identity").GetAwaiter().GetResult())
+            if (FeatureManager.IsEnabled("Identity"))
             {
                 services.AddSingleton<IServerStartup, IdentityStartup>();
             }        
--- a/Oqtane.Server/Startup.cs
+++ b/Oqtane.Server/Startup.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
@@ -14,6 +13,7 @@ using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
+using Microsoft.FeatureManagement;
 using Microsoft.OpenApi.Models;
 using Oqtane.Extensions;
 using Oqtane.Infrastructure;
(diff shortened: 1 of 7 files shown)


Oqtane.Server/Startup.cs
```diff
@@ -136,10 +140,10 @@ public void ConfigureServices(IServiceCollection services)
                 ));
             services.AddDbContext<TenantDBContext>(options => { });
 
-            services.AddIdentityCore<IdentityUser>(options => { })
-                .AddEntityFrameworkStores<TenantDBContext>()
-                .AddSignInManager()
-                .AddDefaultTokenProviders();
+            if (FeatureManager.IsEnabled("Identity"))
+            {
+                services.AddSingleton<IServerStartup, IdentityStartup>();
```
hishamco, line 145: Is this should be Singleton?

hishamco:
@sbwalker this is a proof of concept of what you are suggested about using `IServerStartup` for the core services, I'm still vote to introduce a new interface `IIdentityStartup` and the default implementation will be `IdentityStartup` or provide a way to extend `IdentityStartup` by overriding the default method implementation

If you have another idea that ensure the registration order will not be changes, please shoot it to me

hishamco:
What if we did something like this:

```csharp
services.AddSingleton<IdentityStartup>();
```

Then if we want a custom implementation

```csharp
public class PasswordLessIdentityStartup : IServerStartup
{
    public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
    {

    }

    public void ConfigureMvc(IMvcBuilder mvcBuilder)
    {

    }

    public void ConfigureServices(IServiceCollection services)
    {
        services.AddSingleton<IdentityStartup, PasswordLessIdentityStartup>();
        services.AddIdentityCore<IdentityUser>()
            .AddEntityFrameworkStores<TenantDBContext>()
            .AddSignInManager()
            .AddDefaultTokenProviders()
            .AddPasswordlessLoginTotpTokenProvider();
    }
}
```

Is this will work?!!

Oqtane.Server/Startup.cs
hishamco: Nah seems I did it wrong, the `IServerStartup` will register it's services automatically, but how we can controller the registration in `Startup`?

sbwalker:
I understand the idea behind adding an interface for Identity, however there will be MANY different types of services which need to be registered in startup in the future, so having to introduce a new interface for each new scenario is simply not scalable. I would prefer to try and come up with a different approach.

hishamco:
> I would prefer to try and come up with a different approach

That's why I asked in the discussion, how this should be implemented taking in mind the priority of activating the services

PR #953, opened 2020-11-26
Prioritize IServerStartup & IClientServer

hishamco:
@sbwalker this will make #952 easier,  also will ensure that Oqtane will execute the services with respect of predefine order

Negative -> High (Oqtane Core Services)
Zero        -> Normal
Positive   -> Low


sbwalker:
Interfaces are immutable - so adding an additional property for Order to the existing interfaces is a breaking change, and will cause a runtime error for anyone who has implemented these interfaces in their solutions. However I do appreciate your thought process in terms of trying to come up with a solution for prioritizing services in startup. I think this feature needs a broader discussion before we decide on an implementation.

hishamco:
>Interfaces are immutable - so adding an additional property for Order to the existing interfaces is a breaking change, and will cause a runtime error for anyone who has implemented these interfaces in their solutions

You are right, this will be a braking change

> I think this feature needs a broader discussion before we decide on an implementation.

I already started in the discussion and waiting for your reply

hishamco:
@sbwalker introduing `IServerOrderedStartup` & `IClientOrderedStartup` is extending the current startup interfaces and make ordering startup possible. This will help us to utilize some of the core Oqtane services to use them as proof of concept that startup interfaces are very useful not for modules only but for Oqtane Framework itself

sbwalker:
Let's revisit this once we actually have a core feature that needs to utilize it - most likely when support for other auth schemes is introduced in a future version

PR #954, opened 2020-11-27
Localize log messages

### LINKED ISSUE #933: Localize Logs Messages
We should localize the static logs, coz the dynamics logs are tricky but we could may find a good solution to localize them

hishamco:
Fixes #933

### DIFF
--- a/Oqtane.Server/AssemblyInfo.cs
+++ b/Oqtane.Server/AssemblyInfo.cs
@@ -1,0 +1,3 @@
+﻿using Microsoft.Extensions.Localization;
+
+[assembly: RootNamespace("Oqtane")]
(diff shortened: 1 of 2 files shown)


hishamco:
Log

This is how the added role appear in Arabic

hishamco:
I added `Resources` folder so anyone after can add the resources without creating that folder

sbwalker:
Can you please explain why you needed to add the Root Namespace? It seems unrelated to this Logging enhancement. And it did not seem to be necessary for the foundational localization enhancements you add in 2.0.0.

hishamco:
> why you needed to add the Root Namespace

This because you change the rootnamesapce in `Oqtane.Server.csproj`

```
<RootNamespace>Oqtane</RootNamespace>
```

hishamco:
I already add it to `Oqtane.Client` but seems I miss it for `Oqtane.Server`, if you remember that is why I asked to change the Module Creator too to avoid adding this attribute

PR #955, opened 2020-11-27
Ensure that page action casing doesn't break the URL

### LINKED ISSUE #769: Page Actions in Modules are Case Sensitive
@sbwalker Previously in the gitter group you said:

>Shaun Walker @ sbwalker Jul 09 04:54
@ cody in my earlier post I suggested that alias names and page paths should be case insensitive, and parameters must be case sensitive. The case insensitive elements of the url should be converted to lowercase for consistency. There is no site setting that is needed for enabling or disabling this as it is not a “feature” - it is expected behavior.

I believe that actions on Page Actions in Modules should be case insensitive. I keep running into the problem where I will route to a page action use a lower case e.g. `add` vs `Add` and the browser caches the lower case version so if I try to route to the other it automatically changes back to lower case...

hishamco:
Fixes #769

### DIFF
--- a/Oqtane.Client/UI/SiteRouter.razor
+++ b/Oqtane.Client/UI/SiteRouter.razor
@@ -439,6 +439,11 @@
             module.UseAdminContainer = false;
             module.PaneModuleIndex = -1;
             module.PaneModuleCount = 0;
+            // Change the action casing to lower because TextInfo.ToTitleCase() doesn't change the upper casing strings
+            if (action.ToUpper() == action)
+            {
+                action = action.ToLower();
+            }
             // Ensure that the page action is a Title case to avoid break URLs
             action = CultureInfo.CurrentUICulture.TextInfo.ToTitleCase(action);
 
--- a/Oqtane.Client/UI/SiteRouter.razor
+++ b/Oqtane.Client/UI/SiteRouter.razor
@@ -1,4 +1,5 @@
 @using System.Diagnostics.CodeAnalysis
+@using System.Globalization 
 @using System.Runtime.InteropServices
 @namespace Oqtane.UI
 @inject AuthenticationStateProvider AuthenticationStateProvider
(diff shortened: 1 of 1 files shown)


PoisnFang:
this scenario is not fixed: http://localhost:44357/admin/users/*/13/ADD

PoisnFang:
Also, using ToTitleCase() would be assume that ALL developers will be using title case as their page actions. This will take a bit more to fully flesh out all the given scenarios.And truly fix all the URL case sensitivity in Oqtane

hishamco:
> Also, using ToTitleCase() would be assume that ALL developers will be using title case as their page actions.

Nope, this will enforce the page action to become `Action` not `action` nor `ACTION`

> this scenario is not fixed: http://localhost:44357/admin/users/*/13/ADD

I already checked the one will lower casing, but this not much differ from that, but I will test it too

hishamco:
@PoisnFang as I said before it should works without any issues

PoisnFang:
Not working on mine... where "ADD" is all uppercase. 
image

hishamco:
it will be converted into `Add`, could you please try my PR or I will try to add some unit tests if it's possible

PoisnFang:
Like I said, I did try your PR and it does NOT convert to Add because it is all uppercase and the function thinks its an acronym...

hishamco:
Strange!! I tried http://localhost:44357/admin/roles/*/13/ADD and it works

PoisnFang:
I don't  see how given that the function you use explicitly says that it wont convert words that are all upper case..

hishamco:
I don't know how it works with me, but in both ways we could do an extra check if the action is UPPER case it will be converted into lower case then the `TextInfo` will fix the casing

PoisnFang:
That change should fix it. Are you also planning on including the fix for pages?

hishamco:
The PR fixing the issue you are filed, regarding the pages I'm not sure if the blazor page routing are case sensitive or this managed by Oqtane

sbwalker:
I believe #968 addresses the issues in a more comprehensive manner. The challenge with module action resolution is that the application uses the action parameter from the Url to dynamically create a type name. So I don't think trying to enforce any specific type of casing is the optimal solution. The logic in #968 uses a case insensitive overload of Type.GetType() to attempt to load the type and if successful it persists the actual type name in ModuleState for any downstream operations so that developers do not need to worry about casing.

PR #958, opened 2020-11-28
Oqtane Services Extensibility

hishamco:


hishamco:
@sbwalker this a working prototype that allow to replace or extend Oqtane services with order in mind, regardless of the services I added in `IOqtaneServices`. To test this you could add a new implementation like:

```csharp
public class CustomOqtaneServerServices : DefaultOqtaneServerServices
{
    public override void AddIdentity(IServiceCollection services)
    {
        services.AddIdentityCore<IdentityUser>()
            .AddEntityFrameworkStores<TenantDBContext>()
            .AddSignInManager()
            .AddDefaultTokenProviders()
            .AddPasswordlessLoginTotpTokenProvider();
    }
}
```

Also add

```csharp
[assembly: OqtaneServices(typeof(CustomOqtaneServerServices))]
```

so the `CustomOqtaneServerServices` could be discoverable, after running all the default services will be called except `AddIdentity` which will be called from the new implementation

sbwalker:
I need some time to evaluate this prototype. Based on the code I now understand what you mean when you said that the framework would control the ordering of service registration. One observation is that the heavy reliance on reflection would impact startup performance - I actually had a conversation with folks from Microsoft last week where they said that the biggest contributor to startup performance issues is excessive use of reflection. There are probably alternatives available to avoid so much reflection. That being said, since it is actually not even practical to allow replacement of the majority of services in Oqtane startup currently, I am not even sure if such an elaborate solution is needed at all. That is what I am trying to establish with the questions in the Discussion thread ie. what are the actual requirements we need to satisfy.

hishamco:
> One observation is that the heavy reliance on reflection would impact startup performance - I actually had a conversation with folks from Microsoft last week where they said that the biggest contributor to startup performance issues is excessive use of reflection

We can improve this by using type descriptors, as I said this is just a prototype ;)

> That being said, since it is actually not even practical to allow replacement of the majority of services in Oqtane startup currently, I am not even sure if such an elaborate solution is needed at all. That is what I am trying to establish with the questions in the Discussion thread ie. what are the actual requirements we need to satisfy.

From what I saw in Orchard Core almost the services if not all are replaceable, because we don't know all the uses cases that developers have. IMHO allow the extensibility is one of the main features of any framework, and while the Oqtane itself taking the responsibility of registering the services at the right time and place, no worry at all

BTW I will try to improve the prototype and rid off the reflection if it's possible, also you could take a deep look and please don't be hesitate to share your feedback even it's negative ;)

Thanks

PR #967, opened 2020-12-01
Add Language switcher

### LINKED ISSUE #772: Create Language Switcher component
@sbwalker it would be nice to create a language switch components to allow the Oqtane users select their favorite language instead of dealing with cookies

I will push a PR if sounds looks good to you

hishamco:
Fixes #772

- [x] Blazor Server

- [x] Blazor WebAssembly

### DIFF
--- a/Oqtane.Client/Oqtane.Client.csproj
+++ b/Oqtane.Client/Oqtane.Client.csproj
@@ -17,6 +17,7 @@
     <PackageReleaseNotes>https://github.com/oqtane/oqtane.framework/releases/tag/v2.0.0</PackageReleaseNotes>
     <RootNamespace>Oqtane</RootNamespace>
     <IsPackable>true</IsPackable>
+    <BlazorWebAssemblyLoadAllGlobalizationData>true</BlazorWebAssemblyLoadAllGlobalizationData>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Oqtane.Client/Oqtane.Client.csproj
+++ b/Oqtane.Client/Oqtane.Client.csproj
@@ -31,6 +31,7 @@
     <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly" Version="5.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly.DevServer" Version="5.0.0" PrivateAssets="all" />
     <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly.Authentication" Version="5.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Localization" Version="2.2.0" />
     <PackageReference Include="Microsoft.Extensions.Localization" Version="5.0.0" />
     <PackageReference Include="System.Net.Http.Json" Version="5.0.0" />
   </ItemGroup>
--- a/Oqtane.Client/Program.cs
+++ b/Oqtane.Client/Program.cs
@@ -95,12 +95,15 @@ namespace Oqtane.Client
             var jsRuntime = host.Services.GetRequiredService<IJSRuntime>();
             var interop = new Interop(jsRuntime);
             var culture = await interop.GetLocalStorage("OqtaneCulture");
-            if (culture != null)
+            var localizationService = host.Services.GetRequiredService<ILocalizationService>();
+            var cultures = await localizationService.GetCulturesAsync();
+
+            if (culture == null || !cultures.Any(c => c.Name.Equals(culture, StringComparison.OrdinalIgnoreCase)))
             {
-                var cultureInfo = CultureInfo.GetCultureInfo(culture);
-                CultureInfo.DefaultThreadCurrentCulture = cultureInfo;
-                CultureInfo.DefaultThreadCurrentUICulture = cultureInfo;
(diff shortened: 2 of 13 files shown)


hishamco:
This is how it looks like

Language Switcher

hishamco:
@sbwalker please if you have any feedback or thoughts let me know

sbwalker:
Thank you @hishamco... this is looking good! I have some feedback after reviewing the code...

1. One of the goals in Oqtane is to support both Blazor hosting models in the same application. The preferable way to do this is to minimize the amount of conditional logic ( ie. if WebAssembly... ) and rely on approaches which are supported in both environments. In general, if solutions are built for WebAssembly ( ie. client/server ), they will also work on Blazor Server. So it should be possible to use a common persistence mechanism that works for both models.

2. Oqtane uses REST-style controllers by default. The localization controller uses an RPC-style approach. Is there a reason for the difference in API approach in this case?

3. The LanguageSwitcher component was added to the ControlPanel. The benefit to this approach is that every theme will get access to this feature without having to explicitly add the new component. I think this is the correct approach. The only thing I might suggest is a property on the LanguageSwitcher to enable/disable its display in their theme ( ie. there may be a case where someone wants to hide it even if there are multiple supported cultures )

4. We should not introduce new specialized getCulture and setCulture JSInterop methods when we can use the generic methods that already exist. We already have getCookie and setCookie JSInterop methods. And if we do not want to use cookies ( it would be useful to know why not ) then we should either use the generic localstorage methods introduced in .NET 5 or create our own generic localstorage methods.

sbwalker:
There are a few other things which may affect the implementation of the language switcher, so I will mention them here for further discussion...

- Should there be an option in the Control Panel for Manage Cultures which would allow the Administrator to enable cultures for the current site and specify which culture is the default? The default culture would be used for anonymous users. 

- Should an authenticated user be able to specify their preferred culture in their user profile?

hishamco:
Thanks for your feedback:

1. I will try to unify the solution as much as I can, one thing I need to ask you @sbwalker which I twitted yesterday, I don't know why `CultureInfo.CurrentUICulture.DisplayName` shows the culture name not the display name

2. I will revise that

3. We could add `Visible` property, so the display is controled by the developer

4.  I was thinking to use `getCookie` and `setCookie` but while I started with local storage, I may look into generic localstorage methods introduced in .NET 5

hishamco:
> Should there be an option in the Control Panel for Manage Cultures which would allow the Administrator to enable cultures for the current site and specify which culture is the default? The default culture would be used for anonymous users.

This good as enhancement

> Should an authenticated user be able to specify their preferred culture in their user profile? When a user logs in the system will use the culture they specified rather than the site default culture.

This is HOT, after merge this we should differentiate bewteen site culture and admin cultures because they are totally different, thanks for mentioning this

sbwalker:
I have no idea why CultureInfo.CurrentUICulture.DisplayName shows the culture name not the display name - this seems like a bug?

One other question I have related to Localization is the web console reports:

Cookie “.AspNetCore.Culture” will be soon rejected because it has the “SameSite” attribute

Do you know where this is coming from and if it can be resolved?

hishamco:
> I have no idea why CultureInfo.CurrentUICulture.DisplayName shows the culture name not the display name - this seems like a bug?

May be, that is why I'm thinking to retrieve the display names from the server which is a bad idea, but it solve the issue

> Cookie “.AspNetCore.Culture” will be soon rejected because it has the “SameSite” attribute

`SameSite` is related to GPR if i'm not wrong, where should I see the web console reports, I will try to fix it now if this show up at my end

sbwalker:
I believe it comes from this code in _host.cshtml:

@{
    // Set localization cookie
    var localizationCookieValue = CookieRequestCultureProvider.MakeCookieValue(new RequestCulture(CultureInfo.CurrentCulture, CultureInfo.CurrentUICulture));
    HttpContext.Response.Cookies.Append(CookieRequestCultureProvider.DefaultCookieName, localizationCookieValue);
}

What is the purpose of this logic?

hishamco:
This is set a cookie with the current culture, so one the cookie will be persist and the localization middleware will set the current culture based on the cookie culture

No worry for now, I may delete these line after this PR is ready, I will conider you comments and continue to keep this ready ASAP

hishamco:
> Oqtane uses REST-style controllers by default. The localization controller uses an RPC-style approach. Is there a reason for the difference in API approach in this case?

@sbwalker regarding localization APIs, I followed the style in `InstallationController`, one reason because both functions uses `GET` without parameters

hishamco:
Hooray, final I made cookies run everywhere ;) your final review @sbwalker

hishamco:
Now it's ready to merge, then we can work on your enhancements

chlupac:
Good job with localization, @hisamco. I like it.  I will add Czech resources.

hishamco:
Hi @chlupac welcome back dude ;) long time, hope you are ok, regarding the resources I have an idea that I will post it soon

chlupac:
Yes, I know, it's a shame. I was busy with other projects and unfortunately didn't have time for Oqtane.

hishamco:
Please check #971

PR #968, opened 2020-12-02
Fixes to make site alias, page path, and module action Url resolutions case insensitive. Optimized logic for Settings component which fixed localization rendering issue.

sbwalker:


hishamco:
Which localization rendering issue this PR is fixed?

sbwalker:
@hishamco the module settings component was not localizing its labels correctly on the Module Settings or Permissions tabs. This was because the ModuleType was not set to a valid type in the router - it is an edge case which only affected Module Settings because of the way that custom settings are dynamically embedded within another component.

PR #986, opened 2020-12-07
Fix - return to url during login fail when returnurl contains "/"

chlupac:
Url was transformed to 
https://xxxx.yyy/login/!/returnurl=test/Detail
instead of
https://xxxx.yyy/login?returnurl=test/Detail

### DIFF
--- a/Oqtane.Client/Themes/Controls/LoginBase.cs
+++ b/Oqtane.Client/Themes/Controls/LoginBase.cs
@@ -22,7 +22,7 @@ namespace Oqtane.Themes.Controls
             {
                 returnurl += "/" + PageState.Page.Path;
             }
-            NavigationManager.NavigateTo(NavigateUrl("login", "returnurl=" + returnurl));
+            NavigationManager.NavigateTo(NavigateUrl("login", "?returnurl=" + returnurl));
         }
 
         protected async Task LogoutUser()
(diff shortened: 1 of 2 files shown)


sbwalker:
I am not sure if this is the correct fix as hardcoding the parameter name of "returnurl=" in the logic seems far too specific. It seems there is a more general problem that if a querystring parameter has a value which contains a "/" ( ie. a path delimiter ) then the utility method fails to interpret it correctly. Perhaps the returnurl value should be Url encoded in this case?

chlupac:
Yes, URL encoding /decoding will be correct way, but I'ts more complicated and I'm not sure where should be decoding/encoding implemented.

sbwalker:
See my edited comment above... I don't think url encoding is necessary after reviewing the code.

chlupac:
ok. I will do it your way. Updated version.

sbwalker:
Thank you @chlupac. We really need some better developer documentation to explain the usage for some framework APIs. @PoisnFang did a great job on the Url Parameter enhancement, but clearly there are a few edge cases which can cause challenges.

PoisnFang:
Yes I see what is happening here... not something I had thoughy about during dev

sbwalker:
@PoisnFang I dont think there is anything wrong with your solution - there are 2 fairly trivial options to handle this - either prefix the parameters with "?" or url encode the values so that they do not contain "/"

PR #987, opened 2020-12-09
Localizable Component Enhancement

hishamco:


hishamco:
@sbwalker one related question is there's a use case to not introduce a `Text` property for a `Label` component, because I see it's hard to get the value that has been set into the `RenderFragment`

sbwalker:
The Text property for the Label component is probably the most common localization scenario for module developers and it adds significant benefits in terms of convenience. My understanding is that it is working well which is why you updated all of the Admin module components to use it.... or am I missing something?

Oqtane.Client/Modules/Controls/Label.razor
```diff
@@ -43,16 +43,12 @@ else
 
         _openLabel += ">";
 
-        if (IsLocalizable)
+        var text = Localize("Text", String.Empty);
```
hishamco, line 46: It's working @sbwalker but there's an extra work here coz I'm unable to read the `ChildContent`, I'm asking why it's not string property?!!

sbwalker: I think I was unable to figure out how to utilize @ChildContent within a tag:

```
<label for="name" class="class">@ChildContent</label> 
```
The razor compiler could not handle it so I had to use an ugly hack. If you are able to create a better implementation which has the same interface and produces the same output result, please go ahead.

hishamco: I suffered when I worked with the render tree ;) so it's fine for now, all the cases are handled

PR #997, opened 2020-12-12
FileController - content disposition

chlupac:
Add ability to direct content disposition of file

`api/file/download/{fileid}`  inline
`api/file/download/{fileid}/attach`  attachment

### DIFF
--- a/Oqtane.Server/Controllers/FileController.cs
+++ b/Oqtane.Server/Controllers/FileController.cs
@@ -66,7 +66,7 @@ namespace Oqtane.Controllers
                     {
                         foreach (string file in Directory.GetFiles(folder))
                         {
-                            files.Add(new Models.File { Name = Path.GetFileName(file), Extension = Path.GetExtension(file)?.Replace(".", "") });
+                            files.Add(new Models.File {Name = Path.GetFileName(file), Extension = Path.GetExtension(file)?.Replace(".", "")});
                         }
                     }
                 }
@@ -147,8 +147,10 @@ namespace Oqtane.Controllers
                     {
                         Directory.CreateDirectory(folderpath);
                     }
+
                     System.IO.File.Move(Path.Combine(GetFolderPath(_file.Folder), _file.Name), Path.Combine(folderpath, file.Name));
                 }
+
                 file.Extension = Path.GetExtension(file.Name).ToLower().Replace(".", "");
                 file = _files.UpdateFile(file);
                 _logger.Log(LogLevel.Information, this, LogFunction.Update, "File Updated {File}", file);
@@ -221,7 +223,7 @@ namespace Oqtane.Controllers
             {
                 _logger.Log(LogLevel.Error, this, LogFunction.Create,
                     "File Could Not Be Downloaded From Url Due To Its File Extension {Url}", url);
-                HttpContext.Response.StatusCode = (int)HttpStatusCode.Conflict;
+                HttpContext.Response.StatusCode = (int) HttpStatusCode.Conflict;
                 return file;
             }
 
@@ -229,7 +231,7 @@ namespace Oqtane.Controllers
             {
                 _logger.Log(LogLevel.Error, this, LogFunction.Create,
                     $"File Could Not Be Downloaded From Url Due To Its File Name Not Allowed {url}");
-                HttpContext.Response.StatusCode = (int)HttpStatusCode.Conflict;
+                HttpContext.Response.StatusCode = (int) HttpStatusCode.Conflict;
                 return file;
             }
 
@@ -266,7 +268,7 @@ namespace Oqtane.Controllers
 
             if (!file.FileName.IsPathOrFileValid())
             {
-                HttpContext.Response.StatusCode = (int)HttpStatusCode.Conflict;
+                HttpContext.Response.StatusCode = (int) HttpStatusCode.Conflict;
                 return;
             }
 
@@ -432,9 +434,38 @@ namespace Oqtane.Controllers
             return canaccess;
         }
 
+
+        /// <summary>
+        /// Get file with header
+        /// Content-Disposition: inline
+        /// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition
+        /// </summary>
+        /// <param name="id">File Id from Oqtane filesystem </param>
+        /// <returns>file content</returns>
+
         // GET api/<controller>/download/5
         [HttpGet("download/{id}")]
-        public IActionResult Download(int id)
+        public IActionResult DownloadInline(int id)
+        {
+            return Download(id, false);
+        }
+        /// <summary>
+        /// Get file with header
+        /// Content-Disposition: attachment; filename="filename.jpg"
+        /// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition
+        ///
+        /// </summary>
+        /// <param name="id">File Id from Oqtane filesystem</param>
+        /// <returns></returns>
+
+        // GET api/<controller>/download/5/attach
+        [HttpGet("download/{id}/attach")]
+        public IActionResult DownloadAttachment(int id)
+        {
+            return Download(id, true);
+        }
+
+        private IActionResult Download(int id, bool asAttachment)
         {
             var file = _files.GetFile(id);
             if (file != null)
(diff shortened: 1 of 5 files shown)


Oqtane.Server/Controllers/FileController.cs
```diff
@@ -433,8 +433,8 @@ private bool CanAccessFiles(string[] files)
         }
 
         // GET api/<controller>/download/5
-        [HttpGet("download/{id}")]
-        public IActionResult Download(int id)
+        [HttpGet("download/{id}/{mode?}")]
+        public IActionResult Download(int id,string mode = null)
```
hishamco, line 437:
```csharp
 public IActionResult Download(int id, string mode = null)
```

Oqtane.Server/Controllers/FileController.cs
```diff
@@ -444,7 +444,10 @@ public IActionResult Download(int id)
                     var filepath = Path.Combine(GetFolderPath(file.Folder), file.Name);
                     if (System.IO.File.Exists(filepath))
                     {
-                        return PhysicalFile(filepath, file.Name.GetMimeType(), file.Name);
+                        var result = mode==Constants.ContentDispositionAttachment
```
hishamco, line 447:
```csharp
var result = mode == Constants.ContentDispositionAttachment
```

Oqtane.Server/Extensions/StringExtensions.cs
```diff
@@ -27,5 +28,10 @@ public static string GetMimeType(this string fileName)
 
             return contentType;
         }
+
+        public static string GetMimeType(this File file)
```
hishamco, line 32: I suggest to move this into `FileExtensions` also you could use inline lambda expression body

hishamco (changes requested):

sbwalker:
@hishamco I believe you are commenting on something that is unrelated to this PR... this PR is focused on adding content-disposition functionality

sbwalker:
My main feedback on this PR is that we should not be using optional parameters with default values in APIs as these have a significant risk of causing breaking API changes in the long term. I realize there are other areas in the framework where I have allowed this approach ( despite my better judgement ) but we need to avoid this pattern going forward. I participated in a code review discussion with Microsoft this week where this approach caused some major problems and Microsoft is looking for ways to improve the tooling to warn developers of the problem as most developers like the convenience of optional parameters and are not aware that they can cause breaking changes. This a great blog on the topic if you want more information:

https://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/

hishamco:
I just commenting on the syntax and formatting issues and left the content disposition to you ;) 

Regarding the optional parameter I'm agree with you if it's in a public APIs not in controllers

chlupac:
it's true, and it comes in more elegant solution
```c#
        // GET api/<controller>/download/5
        [HttpGet("download/{id}")]       //for compatibility
        [HttpGet("inline/{id}")]
        public IActionResult DownloadInline(int id)
        {
            return Download(id, false);
        }

        // GET api/<controller>/download/5/attach
        [HttpGet("attach/{id}")]
        public IActionResult DownloadAttachment(int id)
        {
            return Download(id, true);
        }

        private IActionResult Download(int id, bool asAttachment)
        {
.......
}
```

chlupac:
@sbwalker you can merge it if you like it

sbwalker:
Sorry for the delay in getting back to this... I should have provided more details...

The issue with optional parameters only manifests itself in public APIs which are statically linked to your consuming application code. For example, in Oqtane we have Service classes in the Client project which provide strongly typed methods that module developers can use to interact with the backend APIs in their components. These Service classes are statically linked to the components during compilation. So if the Service class methods use optional parameters and later change their method signatures they can result in run-time breaking changes. This is the scenario I was commenting on for the original PR as it appeared that a FileService method had been modified to use an optional parameter. I no longer see this so perhaps it was removed in the modifications you made to the PR.

Web APIs do not have the same issue when it comes to optional parameters ( which is what I think @hishamco was trying to say above ). Based on the fact that Web APIs are not statically linked to application code, there is no run-time binary compatibility issue. So Web APIs can have optional parameters - but you obviously still need to be careful to preserve the legacy behavior as there may be clients which have referenced the legacy Url pattern.

Based on what I described above, I think it would be preferable to follow the original approach you suggested at the beginning of this PR:

api/file/download/{fileid} inline
api/file/download/{fileid}/attach attachment

This preserves the existing Url pattern. There is obviously a behavioral ( not binary ) breaking change if we do this as the original method was erroneously including attach content headers - however I believe there is no client code which will be impacted as nothing was relying on the attach behavior. @iJungleboy was advocating that the default method should have inline content disposition behavior and attach should be reserved for special cases when it is required. With this approach there is really no need to introduce another contstant - and I would prefer if we did not.

The FileService would need an addiitional overload method to call the new API method with the attach parameter. The current method be inline.

I do also agree with @hishamco that the GetMimeType is not a StringExtension. It is only used in scenarios where you are converting a filename to a mimetype which has nothing to do with strings - it is related to files.

chlupac:
> I do also agree with @hishamco that the GetMimeType is not a StringExtension. It is only used in scenarios where you are converting a filename to a mimetype which has nothing to do with strings - it is related to files. 

Me too. Where you like to have it? Best place is as a method of File class, but it is in Shared. Logical place will be MimeUtilities class somewhere in namespace. We have Utilities class on Shared, but I do not want introduce new library reference here.

> api/file/download/{fileid} inline
api/file/download/{fileid}/attach attachment
This preserves the existing Url pattern. There is obviously a behavioral ( not binary ) breaking change if we do this as the original 

My proposal is 
`api/file/inline/{fileid}` for inline
`api/file/attach/{fileid}` or `api/file/download/{fileid}`  for attachment - without breaking change

but I can reverse to my original proposal, it's no  problem

>The FileService would need an addiitional overload method to call the new API method with the attach parameter. The current method would be inline.

In which scenario you need distinguish between inline and attachment in this method? Maybe attachment method should be used here, but it does not matter. 
```c#
        public async Task<byte[]> DownloadFileAsync(int fileId)
        {
            return await GetByteArrayAsync($"{Apiurl}/download/{fileId}");
        }
```

sbwalker:
Ahh... good point. Since the Oqtane.Shared assembly is used by Oqtane.Client you want to keep dependencies to a minimum as you don't want to force additional libraries to be downloaded when running on WebAssembly. It is not very likely that a MIME type will ever need to be determined in client components - it seems to be a more server-focused scenario. So I would think this class belongs in the Server project. Why not a FileExtensions class in the same folder as StringExtensions?

sbwalker:
My preference is 

api/file/download/{fileid} inline
api/file/download/{fileid}/attach attachment

and I am in favor of making the behaviorial breaking change and having the original method serve content inline as that was always the intent of the method ( the fact it was adding the attach header was a bug ).

I think this approach requires fewer changes, has less of an overall impact, introduces less baggage ( ie. obsolete), and affects API consumers less.

It would be good to hear from @iJungleboy as he was the person who raised this issue of content disposition in the first place.

chlupac:
>Why not a FileExtensions class in the same folder as StringExtensions?

```c#
// string extension
        public static string GetMimeType(this string fileName)
        {
            var provider = new FileExtensionContentTypeProvider();
            if (!provider.TryGetContentType(fileName, out var contentType))
            {
                contentType = "application/octet-stream";
// WE CAN ADD ADDITIONAL MAPPINGS HERE
            }
            return contentType;
        }

// Oqtane.Models.File extension
        public static string GetMimeType(this File file)
        {
            return GetMimeType(file?.Name);
        }

// maybe add System.IO.File ?????
```

sbwalker:
By including this method in Models.File wouldn't it create a dependency from the Shared project to the Server project? This would be a problem.

iJungleboy:
i agree with @sbwalker that the default behavior should be inline. That's what 99% of all url do so it should be the standard use case. 

Not sure about the terms though - `attach` sounds like an e-mail behavior. I guess `download` would be better but that`s confusing because the url already says download ATM, and that's what people would expect when they see the urls. 

I think I personally prefer REST style paths where you address a resource and then optionally say what you want. I would imagine something like this: 

`api/file/id/{fileid}` for inline
`api/file/id/{fileid}?download` or `api/file/id/{fileid}?download=true` or `api/file/id/{fileid}?action=download` for download

In future this would then be expanded by another pattern which contains paths/file names, which will look more normal like
`api/file/path/{filepath}/{filepath}/{filepath}/{filename}.ext` for inline
`api/file/path/{filepath}/{filepath}/{filepath}/{filename}.ext?download` (or alternatives) for download

Even if this could co-exist with the previous schema, I would deprecate that since it doesn't feel REST-ful.

sbwalker:
Oqtane does use REST-style controllers:

api/file/{fileid} is already in use:

```
        // GET api/<controller>/5
        [HttpGet("{id}")]
```
it returns a File object - as the FileController is related to the File domain model

sbwalker:
We do not expect to add support for methods such as:

api/file/path/{filepath}/{filepath}/{filepath}/{filename}.ext for inline
api/file/path/{filepath}/{filepath}/{filepath}/{filename}.ext?download (or alternatives) for download

this is because file paths and names are not immutable and therefore not good candidates for APIs.

chlupac:
@sbwalker updated simplified version

PR #1021, opened 2020-12-27
Settings Fix

chlupac:
Settings of module is not saved because UpdateSettings method cannot be invoked

### DIFF
--- a/Oqtane.Client/Modules/Admin/Modules/Settings.razor
+++ b/Oqtane.Client/Modules/Admin/Modules/Settings.razor
@@ -1,4 +1,5 @@
 @namespace Oqtane.Modules.Admin.Modules
+@using Oqtane.Interfaces
 @inherits ModuleBase
 @inject NavigationManager NavigationManager
 @inject IThemeService ThemeService
@@ -137,7 +138,7 @@
                 builder.AddComponentReferenceCapture(1, inst => { _settings = Convert.ChangeType(inst, _settingsModuleType); });
                 builder.CloseComponent();
             };
-        }
+       }
     }
 
     private async Task SaveModule()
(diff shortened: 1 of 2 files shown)


Oqtane.Client/Modules/Admin/Modules/Settings.razor
```diff
@@ -162,14 +162,7 @@
...
 
-        if (_settingsModuleType != null)
-        {
-            var moduleType = Type.GetType(ModuleState.ModuleType);
-            if (moduleType != null)
-            {
-                moduleType.GetMethod("UpdateSettings")?.Invoke(_settings, null); // method must be public in settings component
-            }
-        }
+        _settings?.GetType().GetMethod("UpdateSettings")?.Invoke(_settings, null); // method must be public in settings component
```
chlupac, line 165: We have to introduce interface for settings form (ISettingForm or ISettingControl),
I have now created one setup form and have had trouble remembering how it works.

sbwalker:
This was caused by PR #968 - the way that Settings components are passed via ModuleState was changed which resulted in a regression issue. The proper fix should be:

```
        if (_settingsModuleType != null)
        {
            _settingsModuleType.GetMethod("UpdateSettings")?.Invoke(_settings, null); // method must be public in settings component
        }
```

If you need to see an example of how this works you can look at the Blog example module:

https://github.com/oqtane/oqtane.blogs/blob/master/Client/Settings.razor

Basically it uses a convention that if a module has a component named "Settings" then it will dynamically load that component as a tab within the Manage Settings UI. The benefit of this approach is that the developer has full flexibility over the layout and UI interactions of the component - they are not restricted in any way in what they can do within the component. That being said, in order to be consistent with the user experience in other areas of the framework, there is only a single Save button in the Manage Settings UI ( rather than a Save button in each tab ). When a user selects the Save button the system needs to update all of the settings regardless of which tab they are displayed on. The way this is currently implemented is that a module Settings component must implement a public method:

```
public async Task UpdateSettings()
```

However this is not based on an actual interface - so it is not as intuitive from a developer perspective, as there is no "contract". So I do see the benefit of your suggestion of having an ISettingControl interface which contains an UpdateSettings() method to make this contract more explicit. The question is if it is truly worth it to introduce the interface or not as the current method would still need to be preserved for backward compatibility - so it would result in 2 methods which would need to be supported/maintained in the long term. Could this be more simply improved in other ways - ie. better documentation, examples, etc..

chlupac:
Your fix code is in another notation, but result is same. I will do interface implementation with compatibility fallback tomorow, it is about add one if statement to current code.

chlupac:
Bug fixed,
interface ISettingsControl introduced
ready to check in

sbwalker:
Do we want to implement this in the Module Creator templates so that developers use the new method rather than the old method?

chlupac:
yes, it's only about introduce interface in Settings.razor
```c#
@implements Oqtane.Interfaces.ISettingsControl 
```

PR #1033, opened 2021-01-03
Order users by display name

### LINKED ISSUE #1027: Lists in administrative interface should be ordered alphabetically
image

It's difficult to find correct user here with few hundreds of users.

hishamco:
Fixes #1027


hishamco:
@chlupac seems this is the only page that needs to be sorted, I just scan almost the pages in the admin

Issue #989, opened 2020-12-09
Some comments to localization


chlupac:
@hishamco, great job with localization.
 
I know  than it's work in progress and some of this points are planned or obsolete but:

My recommendation in this topic:
1. Allow use Language Switcher as standalone control in skin. Take it out from Control panel or allow hide it with control panel parameter. I'm not able without modification of Oqtane source switch off Language Switcher in skin.
2. Make default language and supported languages property of site, add this information to settings similarly to SMTP settings.
3. Allow set only one language for site without switcher (I have different sites for different languages where content is not 1:1 translation)

New settings for site:
SupportedLanguages string;
DefaultLanguage string;
ShowSwitcher bool;

.

hishamco:
Hi @chlupac nice to see you again, regarding localization there are some features still in progress or not planned yet

1. `LanguageSwitcher` is already has a `Visible` property as requested by @sbwalker so you can toggle the visibility programmatically also it's located in Skins if I'm not wrong

2. Localization options are stored in `appsettings.json` it could not be in a setting because localization middleware runs early in the pipeline, but we could add it to the settings but change them required to restart the app

3. Let me think about localization options per tenant

sbwalker:
There were some comments in #967 which I will move here so they are easier to find:

- The LanguageSwitcher component was added to the ControlPanel. The benefit to this approach is that every theme will get access to this feature without having to explicitly add the new component. I think this is the correct approach ( as opposed to creating a new LanguageSwitcher component which would need to be explicitly added to every theme ). The only thing I suggest is a property on the LanguageSwitcher to enable/disable its display in a theme. I realize that the language switcher is only visible when multiple supported cultures are defined but there may be cases where someone wants to hide it altogether.

- There be an option in the Control Panel for Manage Cultures which would allow the Site Administrator to enable cultures for the current site and specify which culture is the default, The default culture would be used for anonymous users. The current method of storing this information in appsettings.json was a short term solution and does not satisfy our multi-tenancy needs.

- An authenticated user be able to specify their preferred culture in their user profile. When a user logs in to the system it would use the culture they specified in their profile rather than the site default culture. This allows a site administrator to interact with the site in their native language even if the site caters to users who speak other languages.

hishamco:
We could open a separated issues for each so I can closed it when it's done or we could add a checklist here so it could be easy for tracking

chlupac:
1. Control panel should have parameter ShowLangSwitcher="true" or this should be option for site.
[...]

PR #1045, opened 2021-01-10
Add Languages Management

hishamco:
Fixes part of #989

LanguageManagement

### DIFF
--- a/Oqtane.Client/Modules/Admin/Languages/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Languages/Add.razor
@@ -1,6 +1,7 @@
 @namespace Oqtane.Modules.Admin.Languages
 @inherits ModuleBase
 @using System.Globalization
+@using Microsoft.AspNetCore.Localization
 @inject NavigationManager NavigationManager
 @inject ILocalizationService LocalizationService
 @inject ILanguageService LanguageService
@@ -65,6 +66,11 @@
         {
             language = await LanguageService.AddLanguageAsync(language);
 
+            if (language.IsCurrent)
+            {
+                await SetCultureAsync(language.Code);
+            }
+
             await logger.LogInformation("Language Added {Language}", language);
 
             NavigationManager.NavigateTo(NavigateUrl());
@@ -74,6 +80,18 @@
             await logger.LogError(ex, "Error Adding Language {Language} {Error}", language, ex.Message);
 
             AddModuleMessage(Localizer["Error Adding Language"], MessageType.Error);
+        }
+    }
+
+    private async Task SetCultureAsync(string culture)
+    {
+        if (culture != CultureInfo.CurrentUICulture.Name)
+        {
+            var interop = new Interop(JSRuntime);
+            var localizationCookieValue = CookieRequestCultureProvider.MakeCookieValue(new RequestCulture(culture));
+            await interop.SetCookie(CookieRequestCultureProvider.DefaultCookieName, localizationCookieValue, 360);
+
+            NavigationManager.NavigateTo(NavigationManager.Uri, forceLoad: true);
         }
     }
 }
--- a/Oqtane.Client/Modules/Admin/Languages/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Languages/Add.razor
@@ -2,6 +2,7 @@
 @inherits ModuleBase
 @using System.Globalization
 @inject NavigationManager NavigationManager
+@inject ILocalizationService LocalizationService
 @inject ILanguageService LanguageService
 @inject IStringLocalizer<Add> Localizer
 
@@ -11,12 +12,15 @@
             <Label For="name" HelpText="Name Of The Langauage" ResourceKey="Name">Name:</Label>
         </td>
         <td>
-            <select id="_code" class="form-control" @bind="@_code">
-                @foreach (var culture in GetCultures())
-                {
-                    <option value="@culture.Name">@culture.DisplayName</option>
-                }
-            </select>
+            @if (_supportedCultures?.Count() > 1)
+            {
+                <select id="_code" class="form-control" @bind="@_code">
+                    @foreach (var culture in _supportedCultures)
+                    {
+                        <option value="@culture.Name">@culture.DisplayName</option>
+                    }
+                </select>
+            }
         </td>
     </tr>
     <tr>
@@ -31,7 +35,7 @@
         </td>
     </tr>
 </table>
-<button type="button" class="btn btn-success" @onclick="SaveLanguage">@Localizer["Save"]</button>
+<button type="button" class="btn btn-success @(_supportedCultures?.Count() > 1 ? String.Empty : "disabled")" @onclick="SaveLanguage">@Localizer["Save"]</button>
 <NavLink class="btn btn-secondary" href="@NavigateUrl()">@Localizer["Cancel"]</NavLink>
 
 @code {
@@ -40,9 +44,12 @@
 
     public override SecurityAccessLevel SecurityAccessLevel => SecurityAccessLevel.Admin;
 
-    private static IEnumerable<Culture> GetCultures()
-        => CultureInfo.GetCultures(CultureTypes.AllCultures)
-            .Select(c => new Culture { Name = c.Name, DisplayName = c.DisplayName });
+    private IEnumerable<Culture> _supportedCultures;
+
+    protected override async Task OnParametersSetAsync()
+    {
+        _supportedCultures = await LocalizationService.GetCulturesAsync();
+    }
 
     private async Task SaveLanguage()
     {
--- a/Oqtane.Client/Modules/Admin/Languages/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Languages/Add.razor
@@ -1,0 +1,72 @@
+@namespace Oqtane.Modules.Admin.Languages
+@inherits ModuleBase
+@using System.Globalization
+@inject NavigationManager NavigationManager
+@inject ILanguageService LanguageService
+@inject IStringLocalizer<Add> Localizer
+
+<table class="table table-borderless">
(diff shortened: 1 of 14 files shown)


hishamco:
@sbwalker I'm almost finish, one question to make sure that what I did will work as expected, regarding the `Startup` is it executed for the master tenant only, or whenever the a new tenant created it is executed, because the languages show be different from tenant to another, which implies a localization middleware should register the default and supported cultures for each

sbwalker:
On startup I would expect that the system would register any culture for which localization resources exist ( independent of multi-tenancy ). And then when the system responds to a UI request, it would load the languages for the current Site from the database ( based on alias/multi-tenant resolution ) and it would use that list of languages for UI functions such as the language picker, etc...

hishamco:
> On startup I would expect that the system would register all cultures for which localization resources exist

I think this weird because the number of cultures are HUGE and different from OS to another, another issue here, whenever someone sit the language header the ASP.NET Core middleware will localize the app based on that culture because HTTP headers have a precedency

I'm thinking for another apprach using custom `IServerStartup`, the idea in my mind is to run that startup whenever a tenant is created, this way each tenant is isolated from localization perspective

sbwalker:
I don't think you are understanding what I am saying... let me elaborate...

Oqtane supports localization - however in order for localization to work correctly it requires satellite assemblies to be deployed into the Oqtane installation. So by default Oqtane supports English - but in order for it to support French, the satellite assemblies for French need to be installed. These resources are installed at the application level and have nothing to do with tenancy. So what I was saying is that during startup the system simply needs to know which languages have been previously installed. From a technical perspective it could determine this automatically based on the existence of satellite assemblies - then there would be no need to manually register them in a database or deal with keeping a list in sync. Basically the end result is a "global" list of the languages which map to satellite assemblies - as these are the only languages which are enabled for the installation.

It is only the UI where multi-tenancy plays a role - and it is important not to note that when I say "multi-tenancy" I am referring to Oqtane's ability to organize users and services in a virtual grouping referred to as a Site ( not the actual Tenant table in the Oqtane database ). When a user is interacting with Oqtane it is via a Url which maps to a Site. A Site can specify which Language(s) are enabled - and these are the languages that would be available in the UI.

hishamco:
> It is only the UI where multi-tenancy plays a role - and it is important not to note that when I say "multi-tenancy" I am referring to Oqtane's ability to organize users and services in a virtual grouping referred to as a Site

Got it!! but again the scenario that I mentioned eariler may possible to occurs, so let me complete the PR based on

> So what I was saying is that during startup the system simply needs to know which languages have been installed

I already did this in a very old PR

> A Site can specify which Language(s) are enabled - and these are the languages that would be available in the UI.

This mean the languages for each site is nothing but a subset of the supported cultures

hishamco:
Another question is a default culture for each site, this can't be managed using ASP.NET Core localization cookie, because one site may ovveride the other one. Your thought?

sbwalker:
I am not familiar with the technical details of the ASP.NET Core Localization cookie but from a higher level I am familiar with the behavior of cookies and how they store information in the client browser for each domain name. So my assumption is that when a client browser navigates to an ASP.NET Core application, some logic needs to be executed which deals with the various fallback scenarios for the app. For example, if a user is authenticated then they may have specified their preferred language in their user profile (which still needs to be implemented)... else it needs to look for a language specification in the url, etc... and eventually one of the fallbacks would be the default language defined for the Site - and only if it is not specified then it would ultimately fallback to the DefaultRequestCulture ( for which there can only be one for the entire application ).

chlupac:
I have bad feeling. My programmer instinct. Something is missing here. We are solving translations but here is another aspect of current culture - date format, number format, decimal and currency separator, i do not know how this is handled....

sbwalker:
@chlupac that is known as Globalization - which is different than Localization.

hishamco:
> We are solving translations but here is another aspect of current culture - date format, number format, decimal and currency separator, i do not know how this is handled....

@chlupac ASP.NET Core will handle the globalization stuff based on the culture request

chlupac:
ok, thanks

hishamco:
IMHO managing the prefered cultures for a site from a subset of supported cultures is suited for this PR, no need to set a default culture because it will not take place because the `Startup` is already ran

I think the second PR should be focus on profile language that @sbwalker refered to

sbwalker:
@hishamco I agree that it makes sense to create a different PR for the user profile language preference. 

One question I have is what the "Is Current" column means in the screen shot you posted above? Does it actually mean "Is Default" for the Site?

hishamco:
> Does it actually mean "Is Default" for the Site?

That's was my aim before, but after our discussion this will not applicable in Oqtane with the current implementation because the startup runs once, while in Orchard Core for instance we have a startup for each module, that will guarantee the localization middleware runs for every tenant

So I will delete it from the UI

sbwalker:
Actually, this is one of the requirements outlined above - we need a way for a Site to be able to specify which Languages it supports - and a Site should be able to specify which of those languages is the default.

hishamco:
Let me try something, but I'm sure ASP.NET Core need to setup both default and supported cultures at early stage during the startup

One question to you @sbwalker regarding multisite aka multi-tenancy in Oqtane how it works internally, is there any blog post you wrote, I need to understand something here to make sure if this can work fine with the current implementation or not

sbwalker:
I have never written a detailed blog about multi-tenancy in Oqtane. A high level summary is that it is managed by the router. The router parses the browser Url and makes a call to the AliasController which resolves the Site/Tenant based on the Url and passes back an Alias object. The Alias object is then stored in a Scoped Service so that it is available to all other client services ( which use it to construct API Urls where the AliasId is passed to the server for tenant resolution when accessing the database ).

sbwalker:
I felt like I was missing something in this conversation about localization so I read the official docs:

https://docs.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-5.0

Blazor behaves much differently than a standard .NET Core web application when it comes to Localization - and there are also differences depending on the hosting model. This reminds me of the effort in getting Cookie Authentication to work in both Server and WebAssembly scenarios ( which I documented in the blog: https://www.oqtane.org/Resources/Blog/PostId/527/exploring-authentication-in-blazor )

hishamco:
> A high level summary is that it is managed by the router

As expected, seems things different from Orchard Core, where each tenant has it's own AUTH, localization, modules .. etc

> Blazor behaves much differently than a standard .NET Core web application when it comes to Localization

Ya, that's why I suffered at the beginning to make localization happen in both client & server ;)

sbwalker:
Oqtane has nothing in common with Orchard - it uses a completely different architecture

hishamco:
Ya I mentioned that to make it clear that the localization default and supported cultures need to be run on `Startup`, anyhow I will push some commits to finalize this PR, your feedback is always welcome

hishamco:
@sbwalker please review the PR

The scenario as the following:

1. The Supported Cultures should be added in `appsettings.json` by the host Why?
   1.1 The Satellite assemblies should be complied to be supported
   1.2 No need to add a masive of unused cultures

2. The admin can manage the languages per site

3. The `LanguageSwitcher` should respect the languages for the site

hishamco:
I will try to add a default culture, but I'm sure it may be a problematic when we have multiple sites

sbwalker:
The Tenant.02.00.02.00.sql script has an issue... the relationship should be between the Language and Site tables - not the Language and Tenant tables ( the Tenant table is part of the "master" database and will not exist in each "tenant" database )

sbwalker:
The PageTemplate changes in the SiteRepository seem to have the wrong permissions specified - the Language Management feature should be available to Administrators - not just Hosts:

```
                    new Permission(PermissionNames.View, RoleNames.Host, true),
                    new Permission(PermissionNames.Edit, RoleNames.Host, true)
```

sbwalker:
if you look at any of the other Controllers which are targeted at Administrators ( ie. Page, Role ) you will notice that you don't need to require both the Admin and Host role - you only need to specify the Admin role. This is because a user with host rights will automatically have admin rights in any site. So the Add Admins Role changeset should be modified to only include the Admin role ( follow the patterns in the framework ).

hishamco:
Sure, seems I didn't notice the other controllers

Thanks

hishamco:
Shaun could please try the PR at your end, I need to test few cases, especially with the `IsCurrent` coz still I'm sure that may don't work as expected for the reason I mentioned earlier

sbwalker:
We should probably name the field/property IsDefault. I will do some testing later

hishamco:
BTW @sbwalker I notice that I can create a new tenant when I add a new site, is the difference that the site information stored in another database or is there any other differeneces, furthermore when I tried to create a new tenant a password error shows up all the time, is there any restrictions on the password?

sbwalker:
@hishamco when you create a site you have the ability to specify which "Tenant" it belongs to. In this context, "Tenant" actually means Database ( I admit that this was a bad choice in terminology as it creates a lot of confusion ). Multiple sites can be managed in the same database, or you can have an isolated database for each site. There are some differences - there is only one Master database which is defined by the connection string in appsettings.json and it contains the global application tables. Tenant databases contain the tables which are applicable to a site. This is why there are separate Master and Tenant database scripts. In a default installation the Master and Tenant tables are created in the same physical database for simplicity - which does not make it clear that they are logically separated.

The only restriction on passwords is that they need to be 6 characters long ( defined in Startup,cs ). The other thing to note is that the password you enter when creating a new Tenant needs to be correct password for the 'host' user account that you created when you installed Oqtane.

hishamco:
Ya the terminology are confusing but I knew you mean either use one or multiple databases

> The other thing to note is that the password you enter when creating a new Tenant needs to be correct password for the 'host' user account that you created when you installed Oqtane

Aha, I will try it .. thanks

PR #1056, opened 2021-01-15
Job definition improvement - list of available jobs

chlupac:
image

### DIFF
--- a/Oqtane.Client/Modules/Admin/Jobs/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Jobs/Add.razor
@@ -18,7 +18,12 @@
             <Label For="type" HelpText="Enter the job type" ResourceKey="Type">Type: </Label>
         </td>
         <td>
-            <input id="type" class="form-control" @bind="@_jobType" />
+            <select id="type" class="form-control" @bind="@_jobType">
+                @foreach (var item in _jobTypeList)
+                {
+                    <option value=@item>@item</option>
+                }
+            </select>
         </td>
     </tr>
     <tr>
@@ -83,8 +88,14 @@
     private string _startDate = string.Empty;
     private string _endDate = string.Empty;
     private string _retentionHistory = "10";
+    private List<string> _jobTypeList = new();
 
     public override SecurityAccessLevel SecurityAccessLevel => SecurityAccessLevel.Host;
+
+    protected override async Task OnInitializedAsync()
+    {
+        _jobTypeList = await JobService.GetJobTypesAsync();
+    }
 
     private async Task SaveJob()
     {
--- a/Oqtane.Server/Controllers/JobController.cs
+++ b/Oqtane.Server/Controllers/JobController.cs
@@ -4,6 +4,8 @@ using Microsoft.AspNetCore.Authorization;
 using Oqtane.Models;
 using Oqtane.Shared;
 using System;
+using System.Linq;
+using System.Reflection;
 using Microsoft.Extensions.Hosting;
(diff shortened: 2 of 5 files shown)


Oqtane.Client/Modules/Admin/Jobs/Add.razor
```diff
@@ -18,7 +18,12 @@
             <Label For="type" HelpText="Enter the job type" ResourceKey="Type">Type: </Label>
         </td>
         <td>
-            <input id="type" class="form-control" @bind="@_jobType" />
+            <select id="type" class="form-control" @bind="@_jobType">
+                @foreach (var item in _jobTypeList)
+                {
+                    <option value=@item>@item</option>
```
hishamco, line 24: Surround the `@item` with double quotes

Oqtane.Server/Controllers/JobController.cs
```diff
@@ -34,6 +36,18 @@ public IEnumerable<Job> Get()
             return _jobs.GetJobs();
         }
 
+        // GET: api/<controller>
+        [HttpGet("list")]
+        [Authorize(Roles = RoleNames.Host)]
+        public IEnumerable<string> List()
+        {
+            var types = AppDomain.CurrentDomain
```
hishamco, line 44: Can we introduce a `static` method that retrieved a hosted services, because this may an expenssive operation

hishamco (changes requested):

Oqtane.Client/Modules/Admin/Jobs/Add.razor
chlupac: You are right, I will fix it.

Oqtane.Server/Controllers/JobController.cs
chlupac: Not critical operation, how often this api will be called? No reason to do it. You can cache results somewhere, but why?

Oqtane.Client/Modules/Admin/Jobs/Add.razor
chlupac: Fixed.

Oqtane.Server/Controllers/JobController.cs
hishamco: > how often this api will be called? 

I expect every time the jobs page is called

sbwalker: @chlupac  is correct - the Add Job UI is very rarely called so this list does not need to be cached.

hishamco: If is that no need to change this

Oqtane.Client/Modules/Admin/Jobs/Add.razor
hishamco: Need to be fixed

Oqtane.Server/Controllers/JobController.cs
sbwalker: @hishamco if we wanted to optimize for performance, the best place to introduce a static would in the GetOqtaneAssemblies extension method, as this is called repeatedly during startup and during other operations.

Oqtane.Client/Modules/Admin/Jobs/Add.razor
chlupac: no matter if double quotes is here or not, resulting html is same. Blazor renders attribute correctly.
image

chlupac:
Closing this, because it is obsolete, solved another way.

PR #1059, opened 2021-01-17
auto registration of scheduled jobs

sbwalker:


Oqtane.Server/Infrastructure/Jobs/HostedServiceBase.cs
```diff
@@ -153,27 +162,52 @@ public Task StartAsync(CancellationToken cancellationToken)
...
                         job.IsExecuting = false;
                         jobs.UpdateJob(job);
                     }
+                    else
+                    {
+                        // auto registration
+                        job = new Job { JobType = jobTypeName };
+                        // optional properties
+                        var jobType = Type.GetType(jobTypeName);
+                        var jobObject = ActivatorUtilities.CreateInstance(scope.ServiceProvider, jobType) as HostedServiceBase;
```
chlupac, line 180: Is it necessary to build new instance here to get defaults? this.Name must work.

sbwalker: The StartAsync() method of IHostedService is only executed once when a Host starts up

chlupac: > 
> 
> The StartAsync() method of IHostedService is only executed once when a Host starts up
my fault, I do not recognize method where is it called. This is correct place.

PR #1064, opened 2021-01-18
Added ModuleCreator template for External (Partial, Code-Behind)

Jayson-Furr:
I have added a partial/code-behind copy of the external template. I have thought about creating a new module creator to generate multiple entities and their respective services based on a list, separate from the module name, specified in the creator. Does this sound like a good idea? Or, are modules based on single entity by design?


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=1064) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=1064) before we can accept your contribution.<br/><br/>:x: Jayson-Furr [sign now](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=1064)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/oqtane/oqtane.framework?pullRequest=1064) it.</sub>

sbwalker:
@Jayson-Furr the purpose of the Module Creator in the framework is to provide developers with a basic functional starting point for module development - so the fewer number of options the better or else it becomes confusing and the defeats the intended purpose ( some would argue that even having 2 templates - Internal and External - is too confusing ). In addition, each new template requires ongoing maintenance. My opinion is that it would be awesome to have a more advanced Module Creator... but it should not be in the core framework. It should be created as a custom module that can be installed through Module Management similar to the Blog module.

chlupac:
Best way for this should be create standard `dotnet new` Net framework project template. 
https://docs.microsoft.com/en-us/dotnet/core/tutorials/cli-templates-create-project-template.

sbwalker:
I agree that support via the .NET CLI or even a custom Oqtane CLI ( ie. "oqtane new ..." ) would be a great addition to the project.

Jayson-Furr:
I think the CLI would be a great addition.

PR #1066, opened 2021-01-20
Fixed a bug where the caret / chevron was still displayed when a child page was deleted.

Jayson-Furr:
Fixed a bug where the caret / chevron was still displayed when a child page was deleted. I also added HasNavigationChildren beside HasChildren to better support menus when you dont want the child pages visible in the menu. This will allow a page to have non-navigation children and not show the caret / chevron. The naming may need to change over time?


sbwalker:
@Jayson-Furr can you please submit the menu changes as a separate PR from the module creator template? I have also noticed that the horizontal menu component is buggy. If I add a new page called Test which is a child of Home and specify navigation = true and add to end, the menu displays "Home Test Private ..." - which is incorrect. The menu should only display items at the current Level. The issue you raised is also a problem - where it is not ignoring Deleted pages - however I am not sure if a new property is necessary to fix the problem - I think the menu component is not interpreting the existing properties properly.

Jayson-Furr:
Yes sir, I will re-fork it and get it this weekend.

PR #1067, opened 2021-01-21
Localization fixes - table definition, SQL script naming, SQL script not marked as Embedded Resource, changed column name from IsCurrrent to IsDefault to reflect intent, set default language for site in _Host

sbwalker:


sbwalker:
@hishamco I did some testing of localization and fixed a few issues... and also researched how to set the default language per site in Blazor. I determined that it can be set in _Host - please take a look at the code in this PR.

Oqtane.Server/Oqtane.Server.csproj
```diff
@@ -35,6 +35,7 @@
     <EmbeddedResource Include="Scripts\Tenant.01.00.02.01.sql" />
     <EmbeddedResource Include="Scripts\Tenant.02.00.00.01.sql" />
     <EmbeddedResource Include="Scripts\Tenant.02.00.01.01.sql" />
+    <EmbeddedResource Include="Scripts\Tenant.02.00.01.02.sql" />
```
hishamco, line 38: Nice catch

Oqtane.Server/Pages/_Host.cshtml.cs
```diff
@@ -24,6 +41,24 @@ public void OnGet()
...
             }
+
+            // if framework is installed 
+            if (!string.IsNullOrEmpty(_configuration.GetConnectionString("DefaultConnection")))
+            {
+                Uri uri = new Uri(Request.GetDisplayUrl());
+                var alias = _aliases.GetAlias(uri.Authority + "/" + uri.LocalPath.Substring(1));
+                _state.Alias = alias;
+
+                // set default language for site
```
hishamco, line 52: If you looked to the commits history this change was here, but I remove it later one, when we support both client and server Blazor, I don't think we need it anymore. Could you please let me know what's the reason that enforce you to add it?

hishamco: Seems you want to set the default language, that stored in the database. I need to try your PR in my laptop before we merge

Oqtane.Server/Repository/Interfaces/IAliasRepository.cs
```diff
@@ -9,6 +9,7 @@ public interface IAliasRepository
         Alias AddAlias(Alias alias);
         Alias UpdateAlias(Alias alias);
         Alias GetAlias(int aliasId);
+        Alias GetAlias(string name);
```
hishamco, line 12: This is a breaking change!!

hishamco (changes requested):

Oqtane.Server/Pages/_Host.cshtml.cs
sbwalker: The requirement described previously was for a Site to be able to specify the Languages that it supports and specify which of those Languages is the Default. Then when a user browses to a Site, the framework should set the culture to the default language for the site.

hishamco: I knew but this will break the old localization if I'm not wrong, because server and client blazor using a special cookie name

sbwalker: I do not understand what you mean by "old localization". The code for setting the cookie came from the official docs https://docs.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-5.0. I assume that both client and server Blazor use the same cookie name?

Oqtane.Server/Repository/Interfaces/IAliasRepository.cs
sbwalker: The Repository is not considered part of the public API in Oqtane - only the actual Web API is the public API. Adding an additional method should be fine - in fact it is the expected way to enhance an application ( rather than modifying or deleting methods ). Or are you referring to the fact that this method is defined as a method overload?

Oqtane.Server/Pages/_Host.cshtml.cs
hishamco: I mean the current localization, that could you show may brings some issue, tonight I will test and revise your PR ..

sbwalker: I am going to merge this PR as it contains some fixes which are required for Languages to install properly. We can always back out the changes to _Host if they are problematic - although I think it is the right approach for setting a default culture per site.

Oqtane.Server/Repository/Interfaces/IAliasRepository.cs
hishamco: > The Repository is not considered part of the public API in Oqtane

Ok

> Or are you referring to the fact that this method is defined as a method overload?

Also it could be an extension method if it's possible

sbwalker: I am not sure why it would need to be an extension method?

hishamco: No need while repository APIs not considered to be a breaking changes. BTW could could you please list all the breaking changes in Oqtane in `CONTRIB.md` so everyone will know what are the breaking changes in Oqtane before submit a PR

Oqtane.Server/Pages/_Host.cshtml.cs
hishamco: @sbwalker as I told you this breaks the localization, try to change a language via the switcher, it will not take an affect coz you enforce  the "Host" to set the default language all the time

I'm trying to fix the issue, that why I told you to hold before you merge

PR #1071, opened 2021-01-23
Set default language if the culture not supported

hishamco:
Fixes localization bug introduced in #1067

**Behavior**

The language switcher has been broken

**Steps to Reproduce:**

1. Add two or more languages for a site
2. Set one of them as `IsDefault`
3. Use `LanguageSwitcher` to switch to non default language


hishamco:
The solution I come up with is to set the default language if the `CurrentUICulture` is not belong to the site supported cultures

- If there's a default language it will be used otherwise
- The first language will be set

To test the new behavior create two sites as follows:

Site 1: ar, fr (default)

Site 2: ar  (default), en

1. Open the two sites in the same tab
2. try to switch to Site 1 -> fr
3. Go to the site 2, ar will be set (because it's active)

sbwalker:
I realize that #1067 broke the Language Picker but am trying to understand if this PR solves the problem in a way that achieves the desired behavior. Basically, the goal is that if a new user visits a site ( ie. they have never been to the site before ) then their default culture should be set to the Default Language for that Site... and if no Default Language has been set for the Site then it should fallback and use the default language for the installation ( ie. English ).  If the user then decides to change the culture using the Language Picker, the language they select should become the active language.

hishamco:
> and if no Default Language has been set for the Site then it should fallback and use the default language for the installation ( ie. English )

As explained before if there's no default language set per site, the first one will picked up not English, because the languages SHOULD be in the site supported cultures. I think it should fallback to English if there's no languages set for the site. Am I right?

sbwalker:
I guess I am wondering if it’s possible to simply check in _Host if the Localization cookie is equal to null... and if so then get the Default Language for the current Site and set the cookie. This approach would avoid multiple database calls on the most requests to _Host resulting in better overall performance.

hishamco:
> and if so then get the Default Language for the current Site and set the cookie

What if there's no default language has been set

sbwalker:
Then it would set the cookie to the default language for the installation - English

hishamco:
> which is specified in the appsettings.config ( "DefaultCulture" )

Why not pickup one of the supported cultures for the site, which is make sense, but if there's no language supported, it should fallback to the default in `appsettings.json`

sbwalker:
If an Administrator has specified some supported Languages in the Language Management UI for a Site, then it should be a requirement that 1 of those languages has been specified as the Default. We should not allow an Administrator to specify languages without also specifying a default.

hishamco:
> We should not allow an Administrator to specify languages without also specifying a default.

This could be done by marking the first one as default, then the admin can change it if he wants to

sbwalker:
@hishamco please look at the changes in #1088 which check for the existence of the culture cookie and only if it has a null value does it use the fallback logic to set it.

hishamco:
I was waiting for you respond, I will have a look to your changes in the other PR

PR #1088, opened 2021-02-04
performance and user experience improvements

sbwalker:
There are a number of items in this PR:

- improved performance for module definitions and fixed siteid caching bug 
- improved module deletion user experience so that a restart is not required
- improved module definition deletion to handle situation where asset manifest does not exist
- improved performance for loading themes
- improved theme deletion user experience so that a restart is not required
- improved theme deletion to handle situation where asset manifest does not exist
- added ability to specify a custom admin container for a site
- removed tenant management UI and added tenant info to Site management to reduce confusion 
- changed asset manifest to use relative paths rather than absolute
- allowed modules to include an asset manifest in their nuget package
- modified localization logic so that it only sets cookie if it is null
- add first and last options to pager and made default toolbar display above grid so that position is consistent and does not move based on number of items in list
- converted menu code behind files to inline


sbwalker:
There are a number of items in this PR:

- improved performance for module definitions and fixed siteid caching bug 
- improved module deletion user experience so that a restart is not required
- improved module definition deletion to handle situation where asset manifest does not exist
- improved performance for loading themes
- improved theme deletion user experience so that a restart is not required
- improved theme deletion to handle situation where asset manifest does not exist
- added ability to specify a custom admin container for a site
- removed tenant management UI and added tenant info to Site management to reduce confusion 
- changed asset manifest to use relative paths rather than absolute
- allowed modules to include an asset manifest in their nuget package
- modified localization logic so that it only sets cookie if it is null
- add first and last options to pager and made default toolbar display above grid so that position is consistent and does not move based on number of items in list
- converted menu code behind files to inline

Oqtane.Server/Pages/_Host.cshtml.cs
```diff
@@ -42,8 +42,8 @@ public void OnGet()
                 ProcessThemeControls(assembly);
             }
 
-            // if framework is installed 
-            if (!string.IsNullOrEmpty(_configuration.GetConnectionString("DefaultConnection")))
+            // if culture not specified and framework is installed 
+            if (HttpContext.Request.Cookies[CookieRequestCultureProvider.DefaultCookieName] == null && !string.IsNullOrEmpty(_configuration.GetConnectionString("DefaultConnection")))
```
hishamco, line 46: IMHO cookie check should be inside, I will cover this case in my PR

Issue #1120, opened 2021-02-18
Upload SVG throw error


erw13n:
When upload `.svg` file from File Manager or upload Logo it throw an error :

![image](https://user-images.githubusercontent.com/792847/108285442-ad945500-71b9-11eb-94ef-19d12d93d943.png)

![image](https://user-images.githubusercontent.com/792847/108285524-d6b4e580-71b9-11eb-9c1b-f6fc8f0949d7.png)

Not much information on Logs, but when debugging, it throw error at `FileController > MergeFile`

What's weird is that the file uploaded to Physical Folder, but not listed in `FileManager`
![image](https://user-images.githubusercontent.com/792847/108286405-98202a80-71bb-11eb-9bca-d0db96e81373.png)

leigh-pointer:
The issue is with svg files.  A svg image, is an object in the form of XML there for will not hold these attributes.

```
FileStream stream = new FileStream(filepath, FileMode.Open, FileAccess.Read);
                using (var image = Image.FromStream(stream))
                {
                    file.ImageHeight = image.Height;
                    file.ImageWidth = image.Width;
                }
```
I have put a fix forward for review.

sbwalker:
Question on this item... there is a constant in the framework: 

ImageFiles = "jpg,jpeg,jpe,gif,bmp,png,svg,ico";

This constant is used as a filter in some areas to restrict the files which can be selected/uploaded by a user to only allow image files ( ie. when choosing a logo for a site, etc... ). This constant is also used in the FileController to determine whether the system should dynamically obtain the width and height of an image. Currently SVG is part of this list... but my question is whether this is valid or not?

erw13n:
Can it just read height and width from the object xml, Svg-xml should have this attribute

sbwalker:
@hishamco raised a valid point - SVG files pose a security risk https://wpglorify.com/wordpress-svg-support/. Perhaps the best short term option is to remove SVG from the ImageFiles filter.

hishamco:
Long time we may sanitize the svg content to make it safe to upload such files

sbwalker:
SVG removed as a short term solution

PR #1121, opened 2021-02-18
Fix Upload SVG throw error #1120

leigh-pointer:
Added check if svg then do not read the image height and width properties.


hishamco:
I think we need a lot of work here, beucase `svg` may introduce a security issues

sbwalker:
@hishamco you make a very good point. For reference see https://wpglorify.com/wordpress-svg-support/. I think we should probably remove SVG from the list of image types.

hishamco:
We should remove it or sanitize the content

Issue #1145, opened 2021-02-27
2.0.1 not able to install on Azure using automated Azure Deploy button


ADefWebserver:
I can't install https://github.com/oqtane/oqtane.framework/releases/tag/v2.0.1 on Azure

1) Create Azure website
2) Deploy Oqtane.Framework.2.0.1.Install.zip
3) Install wizard comes up, enter information (including Host information)
4) Database is created
5) When I try to go to site the install wizard comes up (again)
6) no errors in [oqtane-blazor].[dbo].[Log] (it is empty)
7) There are no users in [AspNetUsers] or [User] tables (**this appears to be the problem**?)

sbwalker:
Were you able to install 2.0.0 on Azure?

ADefWebserver:
This was the last version I tried yesterday: (it was the latest version in prod); https://github.com/oqtane/oqtane.framework/issues/1140

sbwalker:
I have installed the 2.0.1 release package successfully on IIS locally - so this must be specifically an Azure deployment issue.

sbwalker:
I am wondering if it might be related to an Azure environment behavior issue similar to https://github.com/oqtane/oqtane.framework/issues/545

ADefWebserver:
> I am wondering if it might be related to an Azure environment behavior issue similar to #545

Note I am trying the manual.zip install so I can use my existing azure app.

1) it fully installs and updates the appsetings.Json

2) I cleared the settings and installed again. Everything installs (tables are created) but no host user is created.

sbwalker:
Do the Tenant, Alias, and Site tables have records in them after install?

ADefWebserver:
> Do the Tenant, Alias, and Site tables have records in them after install?

![image](https://user-images.githubusercontent.com/1857799/109407595-eb6e4780-7936-11eb-9fa9-3bc13e317907.png)

sbwalker:
You are certain this is a clean install of 2.0.1 - without any other module assemblies, site templates, or customizations to the appsettings.json? It appears that the CreateSite() method in DatabaseManager.cs is not executing or is throwing an error during your installation - but I do not know why.

sbwalker:
Can you also share the contents of your SchemaVersions table?

ADefWebserver:
[...]

PR #1146, opened 2021-02-28
Secure ConnectionString with Data protection

hishamco:
While see the `2.0.1` updates (issues & PRs), more specifically https://github.com/oqtane/oqtane.framework/issues/1145?notification_referrer_id=MDE4Ok5vdGlmaWNhdGlvblRocmVhZDE2MzE5OTYzMTM6MzIzNzI2Ng%3D%3D&notifications_query=repo%3Aoqtane%2Foqtane.framework#issuecomment-787326260 I just notice that the connection string is not secure, I think it will be a good to keep safe even in the data base

This PR introduces `DataProtector` service, to secure connection strings, passwords and sensitive data

/cc @sbwalker

### DIFF
--- a/Oqtane.Client/Program.cs
+++ b/Oqtane.Client/Program.cs
@@ -15,7 +15,6 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.JSInterop;
 using Oqtane.Modules;
 using Oqtane.Providers;
-using Oqtane.Security;
 using Oqtane.Services;
 using Oqtane.Shared;
 using Oqtane.UI;
@@ -70,8 +69,7 @@ namespace Oqtane.Client
             builder.Services.AddScoped<ILocalizationService, LocalizationService>();
             builder.Services.AddScoped<ILanguageService, LanguageService>();
 
-            builder.Services.AddDataProtection();
-            builder.Services.AddSingleton<DataProtector>();
+            builder.Services.AddOqtaneDataProtection();
 
             await LoadClientAssemblies(httpClient);
 
--- a/Oqtane.Client/Program.cs
+++ b/Oqtane.Client/Program.cs
@@ -15,6 +15,7 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.JSInterop;
 using Oqtane.Modules;
 using Oqtane.Providers;
+using Oqtane.Security;
 using Oqtane.Services;
 using Oqtane.Shared;
 using Oqtane.UI;
@@ -68,6 +69,9 @@ namespace Oqtane.Client
             builder.Services.AddScoped<ISystemService, SystemService>();
             builder.Services.AddScoped<ILocalizationService, LocalizationService>();
             builder.Services.AddScoped<ILanguageService, LanguageService>();
+
+            builder.Services.AddDataProtection();
+            builder.Services.AddSingleton<DataProtector>();
 
             await LoadClientAssemblies(httpClient);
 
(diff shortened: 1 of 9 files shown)


hishamco:
Please don't merge, there's a weired case I notice especially in `WebAssembly` that may need a further investigation, seems `KeyRingBasedDataProtector` doesn't support in this platform

You could try it if you have time

hishamco:
@sbwalker  FYI I tweeted @blowdart and he told me that the `KeyRingBasedDataProtector` will not work at Blazor Web Assembly, also he said the whole idea for data protection to be done on the server

@blowdart what if we create a custom `IDataProtectionProvider` to make it works for both client and server?

blowdart:
As I said on twitter protecting the keyring is key. You cannot do that client side. Data protection needs a copy of the keyring to work, and taking a copy client side where it can't be protected against an attacker breaks the whole security model.

In addition you'd have an api to exchange the keyring between client and server because of new key generation. So now an attacker has a way to get a copy
of the keyring outside of the browser.

So, to put it simply **don't**

chlupac:
Connection string is server issue, you do not WANT it on client.

hishamco:
Thanks @blowdart

> Connection string is server issue, you do not WANT it on client.

Nah!! @chlupac if you look to the code above you will see that I used default data protection provider which is a `KeyRingBasedDataProtector`, the protection is working fine with Blazor Server-Side model but not for WebAssembly, as you know everything we added in Oqtane we need to ensure it works for both models

blowdart:
My personal opinion is if you want secrets nothing can be client side, as someone can just f12 and bring up browser tools and see everything that's stored. Secret stuff lives server side and you proxy calls through your server, unless you're doing something like gathering a short lived oauth token and keeping it in memory, rather than storage, so it dies with the current session, then you can do things like call graph or other services that support non-confidential clients. Those are few and far between though.

sbwalker:
@hishamco this pull request was never clear to me in the first place in terms of what "problem" it was intended to solve. If the goal was to encrypt sensitive database fields then it probably should have been surfaced as an RFC for broader discussion on the approach, etc... You indicated that the PR should not be merged, but I did review the code anyways, and I came to the same conclusion as @chlupac and @blowdart - encryption should never occur client-side. The reason why your PR is problematic is because you implemented encryption in the Shared project - and the Client project is dependent on the Shared project so when running on WebAssembly it would be pulled down to down to the client. The encryption logic should only be in the Server project. Either way, I am not even sure if we should implement this feature at all - at the very least it would likely need a switch to enable it or disable as it is not convenient for all environments.

hishamco:
Thanks again @blowdart I will wait for @sbwalker thought, the implmentation I will did will be SERVER SIDE, but I need to registered into DI, so the `IDataProtectionProvider` will be resolved then

hishamco:
> @hishamco this pull request was never clear to me in the first place in terms of what "problem" it was intended to solve.

The issue is the connectionStrings are stored as plain text in the database which is harmful, the idea to use data protection APIs to encrypt such information



>  The reason why your PR is problematic is because you implemented encryption in the Shared project 

I just realize know why it's a problematic ;), so it should be move into the server project, now either I can move it to the server or open a discussion thread to discuss this issue and possible solutions

Thanks all

hishamco:
@chlupac seems I miss understood your point, but as @sbwalker said the problem is when I implement this in `Shared` project

hishamco:
> I am not even sure if we should implement this feature at all - at the very least it would likely need a switch to enable it or disable as it is not convenient for all environments.

@sbwalker if you agree to include this feature I will update this PR, otheriwse I will close it

sbwalker:
@hishamco I think I would prefer to not move forward with this enhancement - so I will close it.

PR #1161, opened 2021-03-05
Rows -> Maximum Records

### LINKED ISSUE #1152: Event log Rows Drop down not changing the number of rows displayed
When changing the rows drop down in the Event Log the display is not displaying the total number of rows selected.

hishamco:
Fixes #1152

### DIFF
--- a/Oqtane.Client/Modules/Admin/Logs/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Logs/Index.razor
@@ -36,7 +36,7 @@ else
                 </select>
             </td>
             <td>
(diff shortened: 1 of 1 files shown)


thabaum (changes requested):
Would "Max Rows" be a better solution to utilize less screen space and still make it more clear?

hishamco:
I was thinking about that, before I push the PR ;)

sbwalker:
It is the word "Rows" which is causing the confusion described in #1152 - that is why the suggestion was to change the label to "Maximum Records" so that it is clear that the dropdown has no correlation to the number of rows displayed on the screen.

PR #1162, opened 2021-03-05
Log error when alias is not found

### LINKED ISSUE #1156: Exception thrown when alias not found from requested url in _Host.cshtml.cs
Before version 2.0.1 works fine. I'm using copy of database from live server for development. Alias of development machine is not present in aliases table. 

_Host.cshtml.cs:
image

hishamco:
Fixes #1156

### DIFF
--- a/Oqtane.Server/Pages/_Host.cshtml.cs
+++ b/Oqtane.Server/Pages/_Host.cshtml.cs
@@ -12,6 +12,7 @@ using Microsoft.AspNetCore.Http.Extensions;
 using Oqtane.Repository;
 using Microsoft.AspNetCore.Localization;
 using Microsoft.Extensions.Configuration;
+using Oqtane.Enums;
 
 namespace Oqtane.Pages
 {
@@ -22,13 +23,15 @@ namespace Oqtane.Pages
         private readonly IAliasRepository _aliases;
         private readonly ILocalizationManager _localizationManager;
         private readonly ILanguageRepository _languages;
+        private readonly ILogManager _logger;
 
         public HostModel(
             IConfiguration configuration,
             SiteState state,
             IAliasRepository aliases,
             ILocalizationManager localizationManager,
-            ILanguageRepository languages)
+            ILanguageRepository languages,
+            ILogManager logger)
         {
             _configuration = configuration;
             _state = state;
@@ -55,6 +58,13 @@ namespace Oqtane.Pages
             {
                 var uri = new Uri(Request.GetDisplayUrl());
                 var alias = _aliases.GetAlias(uri.Authority + "/" + uri.LocalPath.Substring(1));
+                if (alias == null)
+                {
+                    _logger.Log(LogLevel.Error, this, LogFunction.Read, "Site Not Found {Alias}", alias);
+
+                    return;
+                }
+
                 _state.Alias = alias;
 
                 // set default language for site if the culture is not supported


sbwalker:
@hishamco you asked @chlupac how he caused this error to occur - so its not clear to me if you actually tested this code  change before submitting the PR?

hishamco:
@sbwalker I made the PR as draft, I did the fixed according to the issue, but still waiting for @chlupac to mention what are the use cases to raise the issue, coz I did few tests without see the actual exception

chlupac:
@hishamco content of my alias table
```
INSERT INTO mfl.dbo.Alias (AliasId, Name, TenantId, SiteId, CreatedBy, CreatedOn, ModifiedBy, ModifiedOn) VALUES (1, N'cortonso.cz', 1, 1, N'host', N'2020-06-12 10:31:52.930', N'host', N'2020-06-12 10:31:52.930');
INSERT INTO mfl.dbo.Alias (AliasId, Name, TenantId, SiteId, CreatedBy, CreatedOn, ModifiedBy, ModifiedOn) VALUES (5, N'cortonso.cz/en', 1, 3, N'', N'2020-06-12 10:33:01.843', N'host', N'2020-06-12 10:33:02.933');
INSERT INTO mfl.dbo.Alias (AliasId, Name, TenantId, SiteId, CreatedBy, CreatedOn, ModifiedBy, ModifiedOn) VALUES (6, N'www.cortonso.cz/en', 1, 3, N'host', N'2020-11-17 14:59:23.437', N'host', N'2020-11-17 14:59:23.437');
```

sbwalker:
@hishamco originally the framework did not require an explicit Alias mapping... if there was no match it would "fall back" to the first alias in the list. However this is problematic for a number of reasons so it was changed to require explicit mapping. You can reproduce this problem in the latest version by creating a new hostname in your hosts file that maps to your installation but is not explicitly declared in the Alias table. Then if you browse to the hostname, the framework will not find a match and will return Null. 

I believe your code fix looks fine in terms of trapping the error - however your log message is not helpful as it would simply say "Site Not Found null" - which does not help the user at all. 

```
var hostname = uri.Authority + "/" + uri.LocalPath.Substring(1);
...
_logger.Log(LogLevel.Error, this, LogFunction.Read, "No Matching Alias For Hostname {hostname}", hostname);
```

However this will only be useful for administrators who are actually able to login to the system using a valid alias. I think it would also be useful if we displayed a message visible to the interactive user to alert them to the source of problem. Basically the _host.cshtml could display an alert message:

```
<div class="alert alert-danger" role="alert">No Matching Alias For Hostname {hostname}</div>
```

sbwalker:
fixed in #1174

sbwalker:
@hishamco I ran into a couple of issues with the guidance I provided above, so I took the simple approach in my PR.

1. The logger does not work in _host. This is because the logger relies on multi-tenancy in order to figure out where to save the log events. If the alias is not known then it cannot determine the tenant. So I had to remove the logging for now.
2. The _host does not include Bootstrap. This is on purpose, as the Theme pulls in the CSS resources which are needed. So I could not style an alert box - I opted to just display simple text instead.

hishamco:
No problem, the improtant thing that the issue has been fixed, BTW could you pleade check your Gitter I sent a message about the localization resources stuff

Thanks

Issue #1160, opened 2021-03-05
Issue with menu Icon Spacing.


leigh-pointer:
Using the horizontal menu the child sub menu icons are not spaced away from the menu caption.  
![Capture1](https://user-images.githubusercontent.com/9447612/110117430-20581f80-7db9-11eb-9f3f-ee569567f619.PNG)
![Capture2](https://user-images.githubusercontent.com/9447612/110117449-277f2d80-7db9-11eb-9f58-126e31d45483.PNG)

The vertical menu renders the menu item ok
![Capture3](https://user-images.githubusercontent.com/9447612/110117459-29e18780-7db9-11eb-98f4-9f6bf2bef01f.PNG)

hishamco:
I can verify from the UI bug, the PR on the way ...

sbwalker:
fixed in #1163

PR #1163, opened 2021-03-05
Issue with menu Icon Spacing. #1160

leigh-pointer:


hishamco:
Seems I comment in your issue and start a PR without know that you will submit one ;) but seems my PR is much simpler to fix the issue #1164

leigh-pointer:
Capturem
My solutions fixed the menu item and leave the FontIcon as is.

hishamco:
The problem can be fixed by adding a little padding to icon instead of changing else where

sbwalker:
Some thoughts on #1163 and #1164 - which essentially are trying to solve the same problem..

1. FontIcon is a component which is shared by both the Horizontal and Vertical menus - so modifying its behavior does not only affect the rendering of Horizontal submenus - it affects all menu scenarios - which is not preferable. As a result I would tend to endorse @leigh-pointer's approach as the better option.

2. Based on the performance best practices for Blazor, it would seem to indicate that creating micro-components such as FontIcon can be a potential performance issue: https://docs.microsoft.com/en-us/aspnet/core/blazor/webassembly-performance-best-practices?view=aspnetcore-5.0#avoid-thousands-of-component-instances. Perhaps it may be better to remove FontIcon and embed the HTML markup fragment into the menu components.

leigh-pointer:
@sbwalker I will remove FontIcon and embed the HTML markup fragment into the menu components, test the markup locally and PR when ready.

hishamco:
Closing PR #1164 it's in your plate @leigh-pointer ;)

PR #1171, opened 2021-03-11
Fix Pager component issue which manifested itself in Event Log. This reverts a fix from #1160 and addresses the root problem.

sbwalker:


sbwalker:
@leigh-pointer I reverted your code fix in #1160. The root cause of the issue was due to state being maintained within the Pager component which was not being re-initialized when OnParametersSet life cycle event was called. Your fix seemed to address the symptoms - not the root cause.

PR #1177, opened 2021-03-13
Delete profile should refresh profiles list

### LINKED ISSUE #1175: Profile controller after delete doesn't refresh the list
Removing a profile field does not update the profile list. 
After reloading the page (key F5) the list is displayed correctly.

hishamco:
Fixes #1175 

/cc @Raceeend

### DIFF
--- a/Oqtane.Client/Modules/Admin/Profiles/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Profiles/Index.razor
@@ -32,7 +32,7 @@ else
 
     protected override async Task OnInitializedAsync()
     {
-        _profiles = await ProfileService.GetProfilesAsync(PageState.Site.SiteId);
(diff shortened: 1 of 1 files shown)


hishamco:
@sbwalker seems we have this issue in many places, I can handle this if looks good

sbwalker:
Component should utilize OnParametersSetAsync() rather than OnInitialized(). Please do not use lambdas assignments in this form as it makes the code much harder to read and understand

hishamco:
> Component should utilize OnParametersSetAsync() rather than OnInitialized().

I already used `OnParametersSetAsync()` at the beginning, but doesn't work unless I call the `GetProfilesAsync()`

> Please do not use lambdas assignments in this form as it makes the code much harder to read and understand

I think it makes everything clean and clear if the method contains a single statement

PR #1188, opened 2021-03-17
Remove admin border after edit

erw13n:
After finish Edit, there's a class "container" which is conflict with Bootstrap that cause an issue on Full-width pane.

### DIFF
--- a/Oqtane.Client/UI/Pane.razor
+++ b/Oqtane.Client/UI/Pane.razor
@@ -18,7 +18,7 @@ else
 
 @code {
     private bool _useadminborder = false;
-    private string _paneadminborder = "container";
+    private string _paneadminborder = "app-pane-admin-border";
     private string _panetitle = "";
 
     [CascadingParameter]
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=1188) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=1188) before we can accept your contribution.<br/><br/>:x: erw13n [sign now](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=1188)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/oqtane/oqtane.framework?pullRequest=1188) it.</sub>

hishamco:
Could you please share a GIF to see the UI issue on action?

erw13n:
Indeed, we can just overwrite the CSS. 
But I would suggest to set _useadminborder to false instead of changing the CSS class.

sbwalker:
I will have to take a closer look at this as I am concerned that this change will have an adverse effect on themes which do not use full-width panes.

sbwalker:
After further testing I think your fix is in fact the correct solution. Thank you for sharing!

PR #1198, opened 2021-03-26
Added Keypress event to catch Enter key on Login page

leigh-pointer:
Currently when the [Enter] key is press no action is carried out.  This pull request sets the focus to the User Name field and when the [Enter] key is pressed the Login procedure is executed.


hishamco:
Looks good but I'm not sure why there's 10 commits, can you rebase on `dev`

hishamco:
I think the No. of commits is too much according to what it needed, also I see some formatting has been added, the PR should be focus on what it should to make it easy to review

Thanks

sbwalker:
@leigh-pointer could you resubmit this PR in a cleaner format so that it does not contain a lot of historical commits?

sbwalker:
Closing this PR as it was replaced by #1216

PR #1203, opened 2021-03-30
make module creator templates extensible

sbwalker:


hishamco:
I was planned to do this but I was quite busy with Arabic resources. Finally I finished them ;)

PR #1206, opened 2021-03-30
Fix localizer in Admin pages

hishamco:
Address the issue in https://github.com/oqtane/oqtane.translations/issues/7#issuecomment-810413650


hishamco:
It will be fixed then in `Oqtane.ResxExtractor`

Issue #1225, opened 2021-04-03
Redirect on log out


leigh-pointer:
not sure if this is by design but when logging out of the site the redirection is to the login page.  Not sure, but should it not redirect to the home page?

sbwalker:
Logout is supposed to redirect to whatever page you are currently on.

leigh-pointer:
Just realized what was going on. I was on a page that required permission, so of course when logging out the redirection would redirect me to log in.  So its not a bug, but does it make sense ?

sbwalker:
resolved by #1246

PR #1235, opened 2021-04-11
Redirect on log out #1225 Fix

leigh-pointer:
Currently when logging out on a page that is not visible to an unauthenticated user the login page is displayed, this is correct but seems pointless as the last action was to log out.  

This change handles a more logical redirection when logging out.


sbwalker:
@leigh-pointer your changes to LoginBase.cs are valid... but it is now saying there is a conflict based on my recent PR I merged. I am not sure why there is a conflict as I did not modify that file? Perhaps you get the latest changes and then submit this PR again?

hishamco:
He can rebase on dev branch

leigh-pointer:
@sbwalker got the changes, (nice!) 
Redirect should be ok now?

sbwalker:
The PR now says 84 files changed - when there should only be 1 file that changed?

leigh-pointer:
Cancel that something gone silly!

PR #1244, opened 2021-04-17
User experience improvements

sbwalker:


sbwalker:
- Page Management can now render custom Theme settings UI based on new ThemeSettingsType property in ITheme interface
- Module Settings can now render custom Settings UI based on new SettingsType property in IModule interface
- Module Settings can now render Container settings UI based on new ContainerSettingsType property in ITheme interface
- Module Creator added to Module Management for UX consistency
- Theme Creator added to Theme Management for UX consistency
- TabStrip/TabPanel rendering bug fixed
- AllowFileManagement property added to RichTextEditor
- Html/Text module enhanced with custom module settings for enabling/disabling File Management
- ModuleActions bug fixed to not render Move To menu when no panes exist
- Theme controls organized into subfolders for Theme and Container
- Oqtane Theme enhanced with Theme and Container settings
- default pane name reverted to Admin for backward compatibility
- WebAssembly download of DLLs fixed to handle satellite assemblies in subfolders
- Page model now loads PageSettings automatically
- Localization Manager dynamically loads list of SupportedCultures rather than using appsettings.config
- Upgrade Manager fixed to handle errors deleting Internal module templates
- Pane Layouts deprecated

hishamco:
> WebAssembly download of DLLs fixed to handle satellite assemblies in subfolders

@sbwalker could you please show me a use case for this one?

> Localization Manager dynamically loads list of SupportedCultures rather than using appsettings.config

Is this mean the cultures in `appsettings.json` are no longer needed?

sbwalker:
WebAssembly download of DLLs fixed to handle satellite assemblies in subfolders
- look at Program.cs - the code creates a dictionary of DLLs and it was only using the filename as the key. Therefore if you had multiple cultures installed, there would be multiple DLLs with the same name which the system would try to add to the dictionary - which threw an error. It is now using the path and filename as the dictionary key.

Localization Manager dynamically loads list of SupportedCultures rather than using appsettings.config
- look at LocalizationManager.cs - it now creates the array of cultures based on what is actually installed ( ie. subfolders of /bin which contain *.resources.dll files ). So the SupportedCultures option in appsettings.json is no longer needed.

hishamco:
I will have a look into your changes ..

PR #1245, opened 2021-04-17
Clean Startup Class

hishamco:


ADefWebserver:
This may break the functionality required to register component suits like Syncfusion. See "Register The Syncfusion License" at: https://blazorhelpwebsite.com/ViewBlogPost/47

hishamco:
How this break your functionality while registering  Syncfusion done in the module project?

ADefWebserver:
> How this break your functionality while registering Syncfusion done in the module project?

@hishamco  - I am not certain it *will* break (I just hope not) :)

I was concerned that if Oqtane is not able to call this method signature, this method may not work anymore:

```
        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
        {
            // Register Syncfusion license
            // Get a free license here:
            // https://www.syncfusion.com/products/communitylicense
            var builder = new ConfigurationBuilder()
                .SetBasePath(env.ContentRootPath)
                .AddJsonFile("appsettings.json", optional: false,
                reloadOnChange: true);
            var Configuration = builder.Build();
            var SyncfusionLicense = Configuration.GetSection("SyncfusionLicense");
            if (SyncfusionLicense != null)
            {
                Syncfusion.Licensing.SyncfusionLicenseProvider
                    .RegisterLicense(SyncfusionLicense.Value);
            }
        }
```
The code for the module is attached:

[Syncfusion_Oqtane_Module.zip](https://github.com/oqtane/oqtane.framework/files/6331389/Syncfusion_Oqtane_Module.zip)

hishamco:
No worry the `Configure` method will still called using another extension that loops through all custom startup classes

ADefWebserver:
> No worry the `Configure` method will still called using another extension that loops through all custom startup classes

@hishamco - Thank you for looking at this :)

hishamco:
I need to resolve the conflict after merging the multi database support PR

hishamco:
@sbwalker do you have any feedback about this PR?

sbwalker:
@hishamco my apologies for taking so long to get to this... I have reviewed your PR and it does make startup a lot cleaner and should not have any compatibility issues as the ordering of operations has not changed. Please resolve the latest conflicts and I will merge.

hishamco:
I will try to resolve the conflict, this need a little time, because one mistake may break everything ;)

hishamco:
Done from my side

Oqtane.Client/Program.cs
```diff
@@ -164,6 +110,45 @@ private static async Task LoadClientAssemblies(HttpClient http)
...
+                startup.ConfigureServices(services);
+            }
+        }
+
+        private static async Task SetCultureFromLocalizationCookie(IServiceProvider serviceProvider)
+        {
+            var jsRuntime = serviceProvider.GetRequiredService<IJSRuntime>();
+            var interop = new Interop(jsRuntime);
+            var localizationCookie = await interop.GetCookie(CookieRequestCultureProvider.DefaultCookieName);
+            var culture = CookieRequestCultureProvider.ParseCookieValue(localizationCookie).UICultures[0].Value;
```
bo-bac, line 140: today if I try to initialize Oqtane (first start) and use "Runtime": "WebAssembly" I'll receive something like
image

this is the fix
`var culture = CookieRequestCultureProvider.ParseCookieValue(localizationCookie)?.UICultures?[0].Value;`

would not like to add separate PR in order to avoid merge conflicts
@hishamco could you please modify your PR

hishamco: Openning a new issue is better for tracking

sbwalker:
@hishamco some changes were lost in this PR - specifically see how the HttpClient is constructed here ( https://github.com/oqtane/oqtane.framework/commit/bc720555c4407251bf71f79cbe6b09c8f9223271#diff-2fbf541e290883c7d8ad8f0bf02d667e2a4a89b5485771e05231ee1a91f2e377 ) - it contains newer code than your PR. I feel like I need to roll back this PR.

sbwalker:
@hishamco it looks like the only code mistakenly removed was in the HttpClient creation so I added it back

sbwalker:
There is still an issue in Visual Studio related to AddOqtaneScopedServices() where the tool is not able to resolve the reference - but the code still compiles fine?

image

sbwalker:
the ITenantManager service was also removed from the PR which broke the application.

hishamco:
Let me double check before we rollback

sbwalker:
@hishamco I already dealt with the 2 issues I found - they are linked to this PR

hishamco:
Thanks, BTW there are some changes happen in translations and static resource I need to follow up why this happened especially static resources

sbwalker:
@hishamco I will be adding the default resource files ( *.resx ) to the Oqtane.Framework project this week ( someone is working through them right now before the PR is submitted to:

- ensure all localized text are in the resx file
- lookups use static keys rather than long descriptions
- common terms use the SharedResources concept

Going forward all UI changes will need to have a corresponding update to the resx files.

hishamco:
Why we need to generate the neutral resx while the default localisation APIs shows the missing resources?

sbwalker:
@hishamco see https://github.com/oqtane/oqtane.translations/issues/34 for an explanation of the changes being made

Oqtane.Client/Program.cs
bo-bac: agree, done

PR #1254, opened 2021-04-19
Default Alias

chlupac:
When alias is not found in alias table, Oqtane fails with exception in SiteRouter.razor line 108 - alias is null. This solution allows define default alias "*" for determine a site which will be shown when alias is not found. 
#1156

### DIFF
--- a/Oqtane.Server/Repository/AliasRepository.cs
+++ b/Oqtane.Server/Repository/AliasRepository.cs
@@ -65,8 +65,8 @@ namespace Oqtane.Repository
                     break; // found a matching alias
                 }
             }
-
-            return alias;
+            // return fallback alias
+            return alias ?? aliases.Find(item => item.Name.Equals("*"));
         }
 
         public void DeleteAlias(int aliasId)


sbwalker:
Can you please explain the scenario where the actual alias value is not known and therefore cannot be explicitly configured in the system? I am asking because there is a security risk to allowing random aliases to map to a "default" site. For example I could configure a custom domain name and map it to the IP address of your site and promote the Url publicly and you would have no idea I am hijacking your site.

chlupac:
scenarios:
1. case of configuration error, changing domain name .. here should be fallback to empty or admin site where only host user can log in and repair configuration. Now it ends in exception and repair is possible only with direct database access. I know than this is an edge case, but errors happens..

2. moving database between production, testing and development site,  

Hijacking can be avoided on http server level. With wrong name you are not able open correct site. www.hijackedsite.com will be rejected by iis,

Default alias is optional, not admin has opportunity not to use it.

sbwalker:
thank you for explaining @chlupac... your explanation makes sense. And the fact that this enhancement is "opt-in" means that the system is secure by default and the decision is up to administrators to decide if they wish to use a wildcard alias

PR #1260, opened 2021-04-20
Remove supported cultures entry from appsettings.json

hishamco:


Oqtane.Shared/Shared/Constants.cs
```diff
@@ -71,6 +71,6 @@ public class Constants {
 
         public static readonly string SatelliteAssemblyExtension = ".resources.dll";
 
-        public static readonly string DefaultCulture = CultureInfo.InstalledUICulture.Name;
+        public static readonly string DefaultCulture = "en";
```
hishamco, line 74: It's better to change the default culture into English instead of the current installed one

PR #1261, opened 2021-04-20
Move SqlServer databases into a separate project

hishamco:
Nothing wrong to refactor the SQL Server databases implementations into another project


cnurse:
@hishamco - @sbwalker and I have been discussing how to handle the Databases, and we are leaning towards packaging them - rather than deploying them automatically.  During Install, once the user has chosen their database, the database assemblies will then be deployed into the bin folder prior to executing the install.  In this way only the actual database files being used are deployed.  This PR would be a necessary first step to achieving that goal.

hishamco:
I have some input to what you said earlier

> > During Install, once the user has chosen their database, the database assemblies will then be deployed into the bin folder prior to executing the install

I think it would be nice to load the supported databases on demand, something like supported cultures, so whenever the database provider(s) assemblies are in the binaries folder the Oqtane should list them during the installation process, but there's once case i'm not sure if you think about, what if one tenant using SQL Server and another one is using PostgreSQL, is the current implementaion that you did handle such case?

sbwalker:
I am in agreement with the approach you suggested @hishamco and I believe @cnurse is working on making these changes. The goal is that Oqtane should focus on performance and only install the dependencies which the user needs. So if they only intend on using SQL Server - then only the SQL Server provider and dependencies need to be installed - not SQLite, MySQL, etc... 

In regards to using multiple databases in the same installation, it is intended to support that use case. The database type is specified by the Tenant and the repositories utilize the Tenant to determine how to connect to the database.

hishamco:
>The database type is specified by the Tenant and the repositories utilize the Tenant to determine how to connect to the database

Hope of we can provide a Samples repo for such use cases

sbwalker:
@hishamco would it be possible for you to resolve the conflicts in this PR and then I will merge it

hishamco:
Sure

hishamco:
There's no conflict while the files are removed, I will try to force push

PR #1239, opened 2021-04-12
Implement Database Migrations and add Multi-Database Support

cnurse:
This Pull Request implements 2 related but separate features:

- Converting the Schema changes from DbUp to EF Core Migrations
- Adding support for additional Databases.

It implements the following changes:

- Adding Database Migrations for each of the existing SchemaVersions scripts from the initial version 1.0 release
- Adding a custom script to "migrate" any exisitng SchemaVersion table records to "initialise the EFMigrationsHistory table
- Adding a Migrations Framework so that it is not necessary to build Migrations for every supported Database, but takes advantage of EFCore's providers.
- Add a ne IOqtaneDatabase interface which identifies the Database being used
- Adding an implementation of IOqtaneDatabase for 5 Databases
- - SqlServer and LocalDB implemented in Oqtane.Server as these rpresent the existing databases supported
- - SqlLite, MySQL and PostGreSQL implemented in external projects as these are new databases
- Adding support in both the Installation Wizard and AddSite/Tenant components to dynamically load the appropriate fields for building the correct connection string for each database.
- Adding support to handle the custom naming schemes for each database type (PostGreSQL uses snake-casing, SQlServer uses Pascal casing)

These changes have been tested for new installs on all database types (SqlServer, Sqlite, MySQL, PostgreSQL) on both Windows and Mac (PostgreSQL and Sqlite), and for upgrades on SQL Server, as well as for adding new Tenants.

Enjoy!!

PR #1263, opened 2021-04-21
Supported cultures perf

hishamco:
@sbwalker I pushed a PR that I pointed in https://github.com/oqtane/oqtane.framework/discussions/1258#discussioncomment-637247 frankly I'm not sure if pushed before or something goes wrong during the merge


sbwalker:
@hishamco you appear to be correct as #1247 was included in the 2.0.2 release - but the change does not appear to exist in the latest Dev code. Something went wrong in the merge or conflict resolution for #1239. Perhaps @cnurse might have a clue of what happened.

hishamco:
No problem, luckly I find something lost from what I committed ;)

sbwalker:
Actually, I am mistaken in my last comment. The code change from #1247 is in the Dev branch https://github.com/oqtane/oqtane.framework/blob/dev/Oqtane.Server/Infrastructure/LocalizationManager.cs - so I dont think any changes were lost. That being said, there seems to be something wrong with this PR as the changes seem to reference an older version of Oqtane code: https://github.com/oqtane/oqtane.framework/pull/1263/commits/0b5d1f9f098c2febefdf73655482042b23adf49c. Are you sure your local code is 100% up to date?

On a different note I had already optimized GetSupportedCultures here https://github.com/oqtane/oqtane.framework/blob/dev/Oqtane.Server/Infrastructure/LocalizationManager.cs - the change being that the framework should only consider subfolders under /bin which actually contain an Oqtane.Client.resources.dll satellite assembly to be valid. If there are culture subfolders under /bin and they are empty then those culture should not be included.

hishamco:
> Actually, I am mistaken in my last comment

Seems you are right, seems I forgot to push the PR, thanks anyway, you could merge the PR now after the last commit I added

hishamco:
@sbwalker please merge #1264 after you merge this PR, basically we should remove the English (default language) from lanaguages list

sbwalker:
this PR breaks the behavior I described above - the framework should only consider subfolders under /bin which actually contain an Oqtane.Client.resources.dll satellite assembly. If there are culture subfolders under /bin and they are empty then those cultures should not be included.

sbwalker:
this logic would also be a performance regression, as it is getting a list of all cultures and then doing an individual GetDirectory() on each one... as opposed to doing a single GetDirectory() call with a very narrow filter.

hishamco:
> If there are culture subfolders under /bin and they are empty then those cultures should not be included

I expect you will say that similarlly any folder contains `Oqtane.Client.resources.dll` will be considered as culture even it's name not a proper culture. I think I will add extra condition to ensure that satellite assembly is included

hishamco:
> this logic would also be a performance regression, as it is getting a list of all cultures and then doing an individual GetDirectory() on each one... as opposed to doing a single GetDirectory() call with a very narrow filter.

Assume we have 100 cultures, and 10 supported cultures and each contains 10 modules

This PR will do a check 100 times, while yours will do 10 * 11 = 110, so if this increased dramatically

No. Cultures < Supported Cultures No. x Oqtane Satellite Assemblies No. x Oqtane Modules Satellite Assemblies No.

hishamco:
Furthermore this method is static and it will get call once

hishamco:
Perhaps your PR will win in case the has few cultures, either way we need Oqtane to get better in terms of PERF, I can do a benchmarks if it's necessary

sbwalker:
This PR  changes the intended functionality - it does not restrict the SupportedCultures to only those subfolders which contain Oqtane.Client.resources.dll. I do not see the benefits of the other changes so my preference is to close this PR ( as it does not fix a bug and the performance improvments are highly debatable ).

hishamco:
I will close this PR and start a new one that will add some other benifit

PR #1264, opened 2021-04-21
Exclude English from the languages list

hishamco:


sbwalker:
I have a question about this PR. There is a fairly common scenario when it comes to localization where someone may want to customize the default English resources. For example, they may want add additional clarification to the help text or error messages based on the knowledge level of their specific users. So in this case you would create a customized English translation and you would want to be able to install it and enable it within your site so that the system uses it rather than default static (English) text embedded in the application. Would this PR not make it impossible to do this?

hishamco:
This PR simply removes English from the list only, so whenever the user change the English translation he/she see the changes immediately after run the app

PR #1265, opened 2021-04-21
Add English to the language switcher

hishamco:


hishamco:
Language Management shouldn't list the English, while the LanguageSwitcher should to let the user switch between the languages

PR #1267, opened 2021-04-21
modify nuget package installer to support satellite assemblies in subfolders

sbwalker:


iJungleboy:
This looks like it's related to #1272 - adding dlls in the subfolder `refs`?

sbwalker:
This PR was focused on allowing satellite assemblies which are located in subfolders under the /bin to be installed. The installer logic utilizes the root folder name in the nuget package to make decisions ( for the /bin folder it uses the "lib" convention ). So if your nuget package includes refs in your nuget package in the lib/refs folder then this PR will extract the files to the proper location.

sbwalker:
According to the nuspec definition https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package refs are supposed to be includes with a root folder name of "ref" - so what I described above is not correct. Your nuget package would have to contain entries for ref/... . And the installer would need to copy this under the bin folder ie. /bin/Debug/net5.0/ref/...

sbwalker:
I am trying to understand why you are using refs "Assemblies are added as references only for compile time; So nothing will be copied into project bin folder." Oqtane nuget packages are designed to be deployed into a production environment - not a development environment. So why do you need compile-time support?

iJungleboy:
@sbwalker I'll respond to your question in the refs in #1272

Issue #1282, opened 2021-04-26
Error with Footer on Oqtane Theme


leigh-pointer:
The change to the theme to fix the footer to the bottom covers the bottom of the Main area.

The content is not visible and window scrolled completely to the bottom.
![image](https://user-images.githubusercontent.com/9447612/116062714-74c69f80-a684-11eb-8376-3d74b0312b3c.png)

![image](https://user-images.githubusercontent.com/9447612/116062464-33ce8b00-a684-11eb-8869-d39b141427bd.png)

sbwalker:
please go ahead and submit a PR to fix this styling issue when the footer option is enabled

PR #1285, opened 2021-04-26
Fix Error with Footer on Oqtane Theme #1282

leigh-pointer:
Added clear div only when the option is selected


hishamco:
Note: Again I see a lot of commits for a simple change

leigh-pointer:
>  Did you try it in Mobile or Tablet screens?
yes. tested with the google browser device toolbar.

hishamco:
Just squash the commits if it's possible or start a new PR wth single commit

Thanks

hishamco:
@sbwalker I think we need a code guidance to avoid submitting unnecessary commits into any PR, this is a reminder for us to make the review process easy

PR #1290, opened 2021-04-27
Fix use the contents area #1287

leigh-pointer:


sbwalker:
there are other changes included in this PR

hishamco:
@leigh-pointer please rebase on `dev` or start new PR after you fetch the latest changes from `dev`, no need for unnecessary changes and unnecessary commits

sbwalker:
@leigh-pointer it might be easier to create a new PR to remove the extra files. I review the code change to Container.razor and I think the ModuleInstance div would also need a modification in order for the content to be displayed full width?

```
        <div class="container-fluid">
            <ModuleInstance />
        </div>
```

Issue #1271, opened 2021-04-22
Permissions on Pages and Modules


leigh-pointer:
When creating/adding a page or adding a module to the page, I would expect the permissions to restricted and not public.  This would allow it be to built/configured before the page / module is set to live.

leigh-pointer:
Yes, that is what I am getting at.

sbwalker:
Please go ahead and make this change in the ControlPanel so that when a new module is added to a page it only provides View permissions to those users/roles which have Edit permissions to the Page.

PR #1299, opened 2021-04-29
Fix Permissions on Pages and Modules #1271

leigh-pointer:


sbwalker:
@leigh-pointer this does not follow the guidance I provided in #1271. The module View permissions should be set to the Page Edit permissions. Consider a site with a custom role of "Editor" and a page provides Edit permissions to the "Editor" role and View permissions to "All Users". A user is logged in who is a member of the "Editor" role. When this user adds a new module to the page, the module should have "Editor" permissions by default - as if the module was only given Edit permissions to "Administrators" ( as in this PR ) then the user would be prevented from editing the module.

sbwalker:
Simply use UserSecurity.GetPermissionStrings(PageState.Page.Permissions) to get a List<PermissionStrings> and then remove the item with a PermissionName of PermissionNames.View and then use UserSecurity.SetPermissionStrings() to recreate the string.

leigh-pointer:
@sbwalker  Just doing that now!

leigh-pointer:
PR Permissions for edit are applied from the page permissions. #1301

PR #1301, opened 2021-04-30
Permissions for edit are applied from the page permissions.

leigh-pointer:


sbwalker:
@leigh-pointer I just tested this and I don't think it is working... the module.Permissions property is never set and you should not need to use any string splitting or joining. Here is some code which works:

                    // set module view permissions to page edit permissions
                    List<PermissionString> permissions = UserSecurity.GetPermissionStrings(PageState.Page.Permissions);
                    permissions.Find(p => p.PermissionName == PermissionNames.View).Permissions = permissions.Find(p => p.PermissionName == PermissionNames.Edit).Permissions;
                    module.Permissions = UserSecurity.SetPermissionStrings(permissions);

Do you want to try this yourself and if it looks good submit another PR?

Issue #1305, opened 2021-05-01
Ui Bug (Minor): tab switching needs two clicks after editing a field


iJungleboy:
If I want to add a page I see this:

![image](https://user-images.githubusercontent.com/5321186/116776980-fc892100-aa6b-11eb-9ab7-1131ba4a220e.png)

If I now click on permissions it works, the tabs switch. 

But if I edit the name of the page and click on permissions, nothing happens. I have to click a second time for permissions to show. 

If I leave the field first (tab away) and then click on  permissions, it works. 

Not sure if this is a Blazor issue or Oqtane, but my guess is it has something to do with validation of the field which somehow prevents the tab switch until a validation cycle has completed

sbwalker:
This is actually a more challenging problem than it appears... it does not have anything to do with validation - it is related to Blazor component rendering when the state has changed. @leigh-pointer's solution seems to address the symptoms rather than the cause, but it may be viable solution - I will test various scenarios.

sbwalker:
I found the root cause and addressed in #1310. It was related to the collection of TabPanels being initialized and rebuilt on every re-render of the component. There are only a few scenarios where this is necessary so I introduced a new property specifically for those scenarios. It seems to be working as expected now... please test.

leigh-pointer:
@sbwalker I have test and the issue is resolved.

iJungleboy:
Awesomeness! 🎉

PR #1307, opened 2021-05-01
Fix Ui Bug (Minor): tab switching needs two clicks after editing a field #1305

leigh-pointer:
Because the the focus is not on the Tab the first click sets the focus and the second click will activate the tab.  Setting the focus when the mouse moves over the tab fixes the issue.


sbwalker:
@leigh-pointer I found the root cause of the issue and resolved it

PR #1318, opened 2021-05-05
Allow installation of Language packages through Language Management

sbwalker:


sbwalker:
@leigh-pointer I know you were experimenting with this already... but it would be great if you could test these enhancements. Basically you need to create a Nuget package which contains a satellite assembly... and then go to Language Management and choose Add Language and use the Upload tab to upload the Nuget package and then restart the application to complete the installation ( the satellite assembly should be placed in the subfolder under the /bin ).

hishamco:
This is good and bad from Security perspective, IMHO allow such thing from a TRUSTED source like NuGet.org much better, we can enter the package URL and the system will extract the language package

hishamco:
I think the intend is all the Oqtane translation should be hosted in Nuget to make it easier for developer to distribute the translatio resources

sbwalker:
Yes it supports that as well... it offers exactly the same options as installing a module or theme. However there are no language packages on Nuget.org yet for Oqtane so the easier way to test this is to use the upload option.

hishamco:
> However there are no language packages on Nuget.org yet for Oqtane

You are right, may it is useful for TESTING purpose but it harmful in production

sbwalker:
@hishamco the upload option allows a host user to install a module, theme, or language package in their site at runtime ( usually a package that they developed for their own use and do not want to publish on Nuget ).

leigh-pointer:
> @leigh-pointer I know you were experimenting with this already... but it would be great if you could test these enhancements. Basically you need to create a Nuget package which contains a satellite assembly... and then go to Language Management and choose Add Language and use the Upload tab to upload the Nuget package and then restart the application to complete the installation ( the satellite assembly should be placed in the subfolder under the /bin ).

Ok, I created a satellite assembly package and it installed with no problem. Unable to test the module at the moment due to oqtanes database issue.

hishamco:
I already reported the database issue

PR #1320, opened 2021-05-06
add upgrade support for language packages

sbwalker:


sbwalker:
@hishamco @leigh-pointer what would the package name be for a specific Nuget language package for the Oqtane framework? Would it be Oqtane.Client.{culture}.{version}.nupkg? Will there ever be a need to localize items in Oqtane.Server or other projects shipped as part of Oqtane... and if so should the language package be named Oqtane.Framework.{culture}.{version}.nupkg so that it could contain multiple satellite assemblies?

hishamco:
`Oqtane.Client.{culture}.nupkg` for Oqtane client resources no need to specify the version while NuGet is already taking care about the versioning

leigh-pointer:
@sbwalker this sounds good to me.

sbwalker:
@hishamco - so you are saying that there will never be any resources for the Oqtane framework other than those on the Oqtane.Client project? I am not so sure about this - which is why I asked the question.

hishamco:
> so you are saying that there will never be any resources for the Oqtane framework other than those on the Oqtane.Client project?

May be there will be resources in other projects, if you remember I moved the shared resources to `Oqtane.Client` to make it easier but in can youu want me to make this happen I can move the shared resources to Shared project so that will prove all the resources are working fine

hishamco:
Regarding to the resources we can start test the newly added feature by generating a NuGet packages for `Oqtane.Translations`, 

@leigh-pointer hope you already test this?

sbwalker:
I dont think anything needs to be moved now... I am trying to anticipate future scenarios and make sure the decisions we make now will support those scenarios. I think the safe assumption is that there may be multiple satellite assemblies needed in the future to localize the framework. So the language package name should not be directly associated to a specific assembly - it should be more generic. Oqtane.Framework.{culture} makes the most sense.

hishamco:
> I think the safe assumption is that there may be multiple satellite assemblies needed in the future to localize the framework

Exactly

> Oqtane.Framework.{culture} makes the most sense.

LGTM

sbwalker:
In regards to Oqtane.Translations it would be best to create a language package for each culture/version - not a single language package for all cultures, as we do not want to force people to install resources that they do not require.

hishamco:
Exactly, but we could have a notion of `Oqtane.Framework.all` to include all the supported translations if someone interested rather than download indivisual package

sbwalker:
true... but it would not be supported through the Language Management UI. The Language Management UI displays whether or not a framework language package exists for a specific culture on Nuget.org.

hishamco:
I didn't test the added feature, but I give it a try when I have some time

PR #1395, opened 2021-05-24
fix #1272 - add support for ref folder in package installation

### LINKED ISSUE #1272: Nuspec to include 'refs'
Our oqtane module supports runtime compilation for in runtime created custom cshtml files.
Runtime compile-time references are required to be in folder `.\refs\`.
We need to deploy also this `refs` folder (and its content) as part of module installation.

I have a hard time finding the right way to specify in nuspec that the content of `refs` folder is extracted in `refs` folder in oqtane. 

In nuspec I used this (as it is simmilar to wwwroot).
```xml
<file src="..\ToSic.Sxc.Oqt.Server\bin\Release\net5.0\refs\**\*.*" target="refs" />
```

The content of `refs` folders ends in its `refs` folder in nupkg, but I got worrnings.

```
Attempting to build package from 'ToSic.Sxc.nuspec'.
Successfully created package 'C:\Projects\2sxc\2sxc\Src\Oqtane\ToSic.Sxc.Oqt.Package\ToSic.Sxc.11.90.0.nupkg'.
WARNING: NU5100: The assembly 'refs\Microsoft.AspNetCore.Antiforgery.dll' is not inside the 'lib' folder and hence it won't be added as a reference when the package is installed into a project. Move it into the 'lib' folder if it needs to be referenced.
WARNING: NU5100: The assembly 'refs\Microsoft.AspNetCore.Authentication.Abstractions.dll' is not inside the 'lib' folder and hence it won't be added as a reference when the package is installed into a project. Move it into the 'lib' folder if it needs to be referenced.
WARNING: NU5100: The assembly 'refs\Microsoft.AspNetCore.Authentication.Cookies.dll' is not inside the 'lib' folder and hence it won't be added as a reference when the package is installed into a project. Move it into the 'lib' folder if it needs to be referenced.
WARNING: NU5100: The assembly 'refs\Microsoft.AspNetCore.Authentication.Core.dll' is not inside the 'lib' folder and hence it won't be added as a reference when the package is installed into a project. Move it into the 'lib' folder if it needs to be referenced.
```

image

Issue it that after installing the module, all these dll's from `refs` folder are extracted in root-folder as normal dll's but that does not work in this case (for razor runtime compilation references).

On oqtane installation hosted in IIS it is expected to have `refs` folder as visible in screenshots.

### root-folder

image

### `refs` folder

image

sbwalker:


sbwalker:
This PR contains a variety of fixes/improvements:

- fix #1272 - add support for ref folder in package installation
- consolidate package installation to wwwroot/Packages folder ( Modules and Themes folders are still supported for backward compatibility )
- installation manifest now stored in Packages folder using packagename.log file
- added UnnstallPackage() method to consolidate logic for processing manifest file
- fixed issue in module and theme uninstall where the item was still displayed in the grid after uninstall
- fixed issue in module uninstall which would sometimes throw an error "Database operation expected to affect 1 row(s) but actually affected 0 row(s)"

PR #1397, opened 2021-05-24
Fix Upgrade issue with new componentized Database projects

cnurse:
This PR addresses the issue of Upgrades by making 3 major changes
1. IsInstalled was invoked before the Db components have been deployed.  This has been changed, so that we ensure the components are deployed first
2. In the Migration that adds the DBType field to the Tenant object, the logic was incorrect so the value was wrong.  This has been fixed
3. During the analysis of the above it was realised tat there is no fundamental difference between the LocalDb and Sql Server providers (the connection string is now created in the Client razor components, so the LocalDB database type has been removed.


cnurse:
Took a bit to resolve the conflicts - but it should be good to go now

PR #1460, opened 2021-06-09
fix issue: WebAssembly with IDM app

albahadly:
when set Runtime as WebAssembly, IDM app recognize  oqtane.zip and download this file and this make an issue and oqtane does not work correctly and for this I set different extention that IDM can't recognize it

### DIFF
--- a/Oqtane.Server/Controllers/InstallationController.cs
+++ b/Oqtane.Server/Controllers/InstallationController.cs
@@ -82,7 +82,7 @@ namespace Oqtane.Controllers
         {
             if (_config.GetSection("Runtime").Value == "WebAssembly")
             {
-                return File(GetAssemblies(), System.Net.Mime.MediaTypeNames.Application.Octet, "oqtane.zip");
+                return File(GetAssemblies(), System.Net.Mime.MediaTypeNames.Application.Octet, "oqtane.oqbz");
             }
             else
             {


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=1460) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=1460) before we can accept your contribution.<br/><br/>:x: albahadly [sign now](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=1460)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/oqtane/oqtane.framework?pullRequest=1460) it.</sub>

sbwalker:
@albahadly by IDM I assume you mean Internet Download Manager? Rather than using an obscure file extension like *.oqbz wouldn't it make more sense to use *.gz which is what Blazor uses internally for compressing DLLs?

sbwalker:
@albahadly do you have any problem changing this to *.gz?

albahadly:
@sbwalker yes still Internet Download Manager recognizes this file


image

sbwalker:
@albahadly so just so I understand - you are confirming we should use *.gz? Since Blazor itself uses *..gz it seems to be the most appropriate option.

albahadly:
@sbwalker we need to choose an unknown extension to Internet Download Manager, and  *.gz does not work fine still framework is broken after change extension  to *.gz

albahadly:
@sbwalker this is file types Internet Download Manager can recognize it and (GZ) one of them
3GP 7Z AAC ACE AIF ARJ ASF AVI BIN BZ2 EXE GZ GZIP IMG ISO LZH M4A M4V MKV MOV MP3 MP4 MPA MPE MPEG MPG MSI MSU OGG OGV PDF PLJ PPS PPT QT R0* R1* RA RAR RM RMVB SEA SIT SITX TAR TIF TIFF WAV WMA WMV Z ZIP

sbwalker:
@albahadly Blazor WebAssembly needs to download assets to the browser - normally these are DLLs but lately Microsoft has been making efforts to reduce the size of the payload so they have been using Brotli compression and Gzip. So in order to support Blazor WebAssembly, Internet Download Manager would need to be configured to allow these assets to be delivered to the browser. So my question for you is how Internet Download Manager handles these assets currently ( ie. DLLs )

sbwalker:
@albahadly what about renaming it to oqtane.dll - the reality is that this "file" is simply a stream of data sent from the server to the browser - it is not actually saved as a file - so it could be named anything.

albahadly:
@sbwalker yes oqtane.dll work fine

albahadly:
> @albahadly Blazor WebAssembly needs to download assets to the browser - normally these are DLLs but lately Microsoft has been making efforts to reduce the size of the payload so they have been using Brotli compression and Gzip. So in order to support Blazor WebAssembly, Internet Download Manager would need to be configured to allow these assets to be delivered to the browser. So my question for you is how Internet Download Manager handles these assets currently ( ie. DLLs )

PR #1461, opened 2021-06-10
refactoring, enhancements, and some fixes

sbwalker:


sbwalker:
- relocated client tenant resolution from router to app component for better performance
- added support for AntiForgeryToken usage in APIs
- added support for HttpClient headers and added custom Authorization Policy support via headers
- implemented antiforgery and auth policy via headers in HtmlText module
- fixed WebAssembly issue where DatabaseService was not being registered
- decoupled syncevents from alias
- fixed UX refresh issue when moving modules in panes
- introduced new ServiceBase constructor which simplifies service development
- adjusted alias management in Site Settings so that it is only available to host users

PR #1466, opened 2021-06-11
back out auth policy header support as Blazor HttpClient is registered as Scoped and can not support variable headers

sbwalker:


sbwalker:
Originally these changes appeared to be working however some peculiarities were detected, which is explained by the Microsoft docs:

"HttpClient is intended to be instantiated once and re-used throughout the life of an application. Instantiating an HttpClient class for every request will exhaust the number of sockets available under heavy loads. This will result in SocketException errors."

The fact that HttpClient is registered as a Scoped service means that and headers which are added for a specific HttpClient invocation will be retained for all subsequent HttpClient calls for the life of the application. This could lead to undesirable side effects if a header is only intended to be used for some API calls and not others, or if the value of the header is not static and is expected to vary - which could lead to issues under load due to race conditions.

PR #1483, opened 2021-06-16
Introduce SharedResource Localisation to ControlPanel

leigh-pointer:
Updated the ControlPanel to use Shared resources for local strings ;

 **Add**, **Edit**, **Delete**, **Cancel**

Corrected the SharedResources.cs namespace.

### DIFF
--- a/Oqtane.Client/SharedResources.cs
+++ b/Oqtane.Client/SharedResources.cs
@@ -1,4 +1,4 @@
-namespace Oqtane.Client
+namespace Oqtane
 {
     /// <summary>
     /// Dummy class used to collect shared resource strings for this application
--- a/Oqtane.Client/Themes/Controls/Theme/ControlPanel.razor
+++ b/Oqtane.Client/Themes/Controls/Theme/ControlPanel.razor
@@ -10,6 +10,7 @@
 @inject ILogService logger
 @inject ISettingService SettingService
 @inject IStringLocalizer<ControlPanel> Localizer
+@inject IStringLocalizer<SharedResources> SharedLocalizer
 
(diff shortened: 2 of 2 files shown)


Oqtane.Client/SharedResources.cs
```diff
@@ -1,4 +1,4 @@
-namespace Oqtane.Client
+namespace Oqtane
```
hishamco, line 1: Why?

hishamco (commented):
Just a comment for almost your PRs, why I see extract commits that are not related. Please make your local dev branch up to date before you create a PR

Thanks

hishamco:
Could you apply the change in all the places in the control panel

Oqtane.Client/SharedResources.cs
hishamco: > Corrected the SharedResources.cs namespace.

I know ;) but why, while the class in `Oqtane.Client`

sbwalker:
@leigh-pointer there is someone who is going through the entire Oqtane.Client project this week to ensure all localizations are up to date ( see item 2 on https://github.com/oqtane/oqtane.translations/issues/34 ) - it would be helpful to not introduce PRs at this moment as it will cause conflicts when the large PR is submitted.

sbwalker:
@hishamco there are not any other items of static text in the control panel which belong in the SharedResources. SharedResources should only be used for common terms which are used in multiple areas of the user interface - all unique terms to the razor component should be localized in its own dedicated resx file.

hishamco:
> there is someone who is going through the entire Oqtane.Client project this week to ensure all localizations are up to date

Who is ;)

> SharedResources should only be used for common terms which are used in multiple areas of the user interface

But those terms can be found in many places

PR #1485, opened 2021-06-16
Add missing Oqtane.Server into Databases solution

hishamco:
I notice that Databases doesn't build becasue the fact `Oqtane.Server` was missing


sbwalker:
@hishamco good catch on the Oqtane.Databases solution reference. However the modification to the .gitignore needs to be removed as we actually need to be able to commit the nupkgs that are created in the wwwroot/Packages folder. If you do not want to check in these packages from your local environment then just delete them rather than modifying the .gitignore.

hishamco:
But why the need to be tracked in Git?

sbwalker:
Because these packages are a required part of the overall solution: https://github.com/oqtane/oqtane.framework/tree/dev/Oqtane.Server/wwwroot/Packages

hishamco:
I will revert the last commit

PR #1490, opened 2021-06-17
add extension method for Localization which allows specification of key and value

sbwalker:


Oqtane.Client/Extensions/OqtaneLocalizationExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+    public static class OqtaneLocalizationExtensions
+    {
+        /// <summary>
+        /// Gets the string resource for the specified key and returns the value if the resource does not exist
+        /// </summary>
+        /// <param name="localizer"></param>
+        /// <param name="key">the static key used to identify the string resource</param>
+        /// <param name="value">the default value if the resource for the static key does not exist</param>
+        /// <returns></returns>
+        public static string GetString(this IStringLocalizer localizer, string key, string value)
```
hishamco, line 12: @sbwalker the localization APIs is already return the missing resource, in case you need to set a default value what the scenario or use case that this method can be used for?

sbwalker: @hishamco the problem is documented here https://github.com/oqtane/oqtane.translations/issues/34 and was earlier raised here https://github.com/oqtane/oqtane.translations/issues/17:

3. Localization implemented within components will use static keys. This change is being made to ensure that language specific translations are not broken when static text messages are modified in UI components. The standard method for .NET Core localization was theoretically optimized to make software development easier; however it is a nightmare for translators because the key values are not static and often change - breaking the key lookup linkage. 

For example, logic such as Localizer["Please Enter All Fields And Ensure Passwords Match And Are Greater Than 5 Characters In Length"] may require a change to the message - which would break the lookup for all existing resource files for other languages. This will be changed to use an extension method Localizer.GetString("Message.PasswordValidation", "Please Enter All Fields And Ensure Passwords Match And Are Greater Than 5 Characters In Length") where "Message.PasswordValidation" is a static key that will never change and therefore can be relied upon by translators.

hishamco:
So why we didn't use the key `Message.PasswordValidation` something similar to what we did in resource names for components

sbwalker:
Localizer["Please Enter All Fields And Ensure Passwords Match And Are Greater Than 5 Characters In Length"] is not a component - it is static text within a component. I suppose we could introduce a "Message" component with a ResourceKey property ( similar to Label ) but the general guidance from Microsoft in regards to Blazor application performance is to try and avoid "micro-component" usage as components have a life cycle and consume system resources.

hishamco:
Could you please let me know what the resx will contains for the above example?

sbwalker:
file.razor:

Localizer.GetString("Message.PasswordValidation", "Please Enter All Fields And Ensure Passwords Match And Are Greater Than 5 Characters In Length") 

file.resx:

```
  <data name="Message.PasswordValidation" xml:space="preserve">
    <value>Please Enter All Fields And Ensure Passwords Match And Are Greater Than 5 Characters In Length</value>
  </data>

```

Static key "Message.PasswordValidation" NEVER changes.

hishamco:
I see, but after you decide to support neutral resx the static keys and their messages will always be there. Am I right?

PR #1499, opened 2021-06-20
OnInitialized is not the right method to do this, Content is not set yet

ijaz-saeed:


sbwalker:
@ijaz-saeed there does not appear to be any details on the problem this code change is intended to fix, how to reproduce the problem, or validate the solution. Please supply this info so that this PR can be reviewed.

ijaz-saeed:
hi @sbwalker sure, here is the explanation of problem with steps to reproduce.
**Steps to Reproduce** 

1. Install Oqtane.Blogs module and add it to any page
2. create first blog and it is added successfully 
3. now Edit that new blog, hit save WITHOUT making changes to the content ( mean, only go to edit page, hit save WITHOUT updated the contents )
4.  Contents disappear

**Explanation**
we set content property of RichTextEditor control twice in blogs Edit page. once it is set to empty and then an Async call is made to fetch the contents, once this call is completed contents are set again. 
On the second set the OnInitialized event is not called again and _contents remain empty.  so, the line https://github.com/oqtane/oqtane.framework/blob/dev/Oqtane.Client/Modules/Controls/RichTextEditor.razor#L185 is empty, if you hit save WITHOUT making any change to Contents.

Issue #1543, opened 2021-07-07
Localization Resources are not being loaded for external Modules


leigh-pointer:
In the Constants class the SatelliteAssemblyExtension is not correct ( `= ".resources.dll"; ` ) the Wildcard is missing and is needed.

I will raise a PL for this issue.

sbwalker:
fixed in #1545

PR #1544, opened 2021-07-07
Fix for #1543 Localization Resources not being loaded for ext Modules

leigh-pointer:
The EnumerateFiles was not using the correct searchPattern, the Wildcard (*) was missing which in turn will not find any local Resources.

### DIFF
--- a/Oqtane.Shared/Shared/Constants.cs
+++ b/Oqtane.Shared/Shared/Constants.cs
@@ -72,7 +72,7 @@ namespace Oqtane.Shared {
         };
         public static readonly string[] InvalidFileNameEndingChars = { ".", " " };
 
-        public static readonly string SatelliteAssemblyExtension = ".resources.dll";
+        public static readonly string SatelliteAssemblyExtension = "*.resources.dll";
 
         public static readonly string DefaultCulture = "en";
 


Oqtane.Shared/Shared/Constants.cs
```diff
@@ -72,7 +72,7 @@ public class Constants {
         };
         public static readonly string[] InvalidFileNameEndingChars = { ".", " " };
 
-        public static readonly string SatelliteAssemblyExtension = ".resources.dll";
+        public static readonly string SatelliteAssemblyExtension = "*.resources.dll";
```
hishamco, line 75: I presume this is an extension which is fine, if there's a bug in search we could fix in the place where search occurs

leigh-pointer: It occurs in `static class OqtaneServiceCollectionExtensions` 
on `static void LoadSatelliteAssemblies(string[] supportedCultures)`

The code in foreach would **never** execute without a wildcard
`foreach (var assemblyFile in assembliesFolder.EnumerateFiles(Constants.SatelliteAssemblyExtension))`

hishamco: I created #1544 which don't break the satellite extension

sbwalker:
@leigh-pointer I merged #1545 because it preserves the original intent of the SatelliteAssemblyExtension constant - which is to identify the file extension for satellite assemblies. The wildcard makes more sense to be implemented in the method where there the constant is used.

PR #1554, opened 2021-07-14
ConltolPanel to use Bootstrap Offcanvas component

leigh-pointer:
Offcanvas is a sidebar component that can be toggled on of off


sbwalker:
I do think it is a good idea to use the new native BS5 offcanvas component for the Control Panel because it is much more appropriate than the card component which is currently being used. 

I do believe that as per the standard approach for all Oqtane controls, it still needs to have a selector "app-controlpanel" so that it is possible to target style overrides without affecting all offcanvas components in a site. 

This means the current styles in app.css for app-controlpanel probably need to be modified or removed.

sbwalker:
@leigh-pointer the user experience of the control panel has changed as a result of this PR. I would prefer if the previous behavior was retained where the control panel would be hidden if a user clicked on the page outside of the control panel - and also would be hidden when a user selected a button on the control panel such as Admin Dashboard.

PR #1573, opened 2021-07-28
additions to the required inputs and validations

gjwalk:


sbwalker:
This PR contains a lot of extraneous files which should not be added to the repo. In addition, it has modifications to so many files that it is next to impossible to review throughly. My preference would be to break it into smaller more manageable PRs.

PR #1639, opened 2021-09-02
Replacing dependency on System.Drawing with SixLabors.ImageSharp based on cross platform guidance from Microsoft https://github.com/terrajobst/designs/blob/9b4520703cfa11ccfbd6e4389ab62730e0cb0de4/accepted/2021/system-drawing-win-only/system-drawing-win-only.md

sbwalker:


sbwalker:
https://github.com/terrajobst/designs/blob/9b4520703cfa11ccfbd6e4389ab62730e0cb0de4/accepted/2021/system-drawing-win-only/system-drawing-win-only.md

Issue #1645, opened 2021-09-05
After Delete Page, Cant create page of same name


rchibnik:
I created a page named "CMS" and instead of turning off navigation, I deleted it.

When I try to re-create a page named CMS i get the error:
A page with path CMS already exists for the selected parent page. The page path needs to be unique for the selected parent.
but I can't find the page in my list of pages,  So the name CMS is no longer available to use as a page.

A minor defect - hopefully an easy fix.

I love Oqtane. Thanks for creating such an excellent framework. I look forward to using it for years to come.

leigh-pointer:
The CMS page is not truely deleted until you clean out the recycle bin. This is found in the manager settings.

hishamco:
> The CMS page is not truely deleted until you clean out the recycle bin

Good point, nothing but I suggest to be able to create a new page with the same name, similar to what Windows is doing while deleting a file

rchibnik:
Thanks - found the recycle bin and confirmed that it works as designed.

PR #1646, opened 2021-09-08
Page create - Recycle Bin Check

leigh-pointer:
After Delete Page, Cant create page of same name #1645 issue. Added check and message if the page is in the recycle bin.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Pages/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Pages/Add.razor
@@ -322,6 +322,12 @@
                         }
                     }
 
+                    if(!PagePathIsDeleted(page.Path, page.SiteId, _pageList))
+                    {
+                        AddModuleMessage(string.Format(Localizer["Message.Page.Deleted"], _path), MessageType.Warning);
+                        return;
+                    }
+
                     if (!PagePathIsUnique(page.Path, page.SiteId, _pageList))
                     {
                         AddModuleMessage(string.Format(Localizer["Message.Page.Exists"], _path), MessageType.Warning);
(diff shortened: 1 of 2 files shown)


Oqtane.Client/Modules/Admin/Pages/Add.razor
```diff
@@ -412,4 +418,9 @@
     {
         return !existingPages.Any(page => page.SiteId == siteId && page.Path == pagePath);
     }
+
+    private static bool PagePathIsDeleted(string pagePath, int siteId, List<Page> existingPages)
```
hishamco, line 422: I suggest to move the LINQ query before the condition

Oqtane.Client/Modules/Admin/Pages/Add.razor
```diff
@@ -322,6 +322,12 @@
                         }
                     }
 
+                    if(!PagePathIsDeleted(page.Path, page.SiteId, _pageList))
+                    {
+                        AddModuleMessage(string.Format(Localizer["Message.Page.Deleted"], _path), MessageType.Warning);
+                        return;
```
hishamco, line 328: Add line break before 'return`

hishamco (changes requested):
This check may be a workaround, the good solution is allow to create a page with the same name

leigh-pointer:
> This check may be a workaround, the good solution is allow to create a page with the same name

you can't create a page with the same name because the database table Page has a UNIQUE NONCLUSTERED INDEX on it consisting of fields [SiteId],[Path],[UserId].

hishamco:
You're right, I know that, but we may change this to make it an optimal solution

/cc @sbwalker

sbwalker:
@leigh-pointer are you sure the logic below is correct? The condition "!PagePathIsDeleted()" seems to be saying "if the page is NOT deleted" then show the warning message - which does not seem accurate?

```
if(!PagePathIsDeleted(page.Path, page.SiteId, _pageList))
{
     AddModuleMessage(string.Format(Localizer["Message.Page.Deleted"], _path), MessageType.Warning);
    return;
}

PR #1698, opened 2021-09-28
1690 User Management Tab needs clicking to render UI when language is not default.

leigh-pointer:
Added the Heading property to all uses of the TabPanel component.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Languages/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Languages/Add.razor
@@ -16,7 +16,7 @@
 else
 {
     <TabStrip>
-        <TabPanel Name="Manage" ResourceKey="Manage">
+        <TabPanel Name="Manage" Heading="Manage" ResourceKey="Manage">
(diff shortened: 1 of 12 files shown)


sbwalker:
I don't think this is the optimal solution. Rather than forcing developers to have to use the Header property - with such widespread impacts - can we not concentrate the changes to the TabPanel component alone? The current behavior is that the component will use the Name value as the Header if the Header is not provided. I believe the DisplayHeading() method in TabPanel.razor could be modified to display the localized values - without changing the values of the Name or Header properties themselves.

leigh-pointer:
I have closed this and have got a better solution along your path.  I will send a new Pull Request.

`    
protected override void OnParametersSet()
    {
        base.OnParametersSet();

        Parent.AddTabPanel((TabPanel)this);

        if (string.IsNullOrEmpty(Heading))
        {
            Heading = Localize(nameof(Name), Name);
        }
    }

    public string DisplayHeading()
    {
        return (string.IsNullOrEmpty(Heading)) ? Name : Heading;
    }
`

Issue #1736, opened 2021-10-27
Blazor Theme not responding correctly


leigh-pointer:
![image](https://user-images.githubusercontent.com/9447612/139030150-bcb30739-74dd-4c51-9ef5-b56a6fefc975.png)

sbwalker:
resolved in #1750

PR #1737, opened 2021-10-28
Fix for #1736 Blazor theme not rendering correctly

leigh-pointer:
Fix to the Default theme and container


leigh-pointer:
@sbwalker all checked and fully responsive

Issue #1749, opened 2021-10-31
unable to navigate to sub sub pages


leigh-pointer:
navigation to sub sub pages is difficult as the current menu components do not support sub menus,

sbwalker:
added in #1752

PR #1752, opened 2021-11-02
Fix for #1749 navigate to sub sub pages

leigh-pointer:
Added Open button that will navigate to sub pages


sbwalker:
@leigh-pointer it would be better if this followed the pattern in Site Management where the button is named "Browse" and is displayed in a column after Edit and Delete

leigh-pointer:
Ok, i will change it.

PR #1810, opened 2021-11-15
Page management buttons resizing

leigh-pointer:
When the language is changed, in this instance Dutch the buttons are not resized to fit the caption.  This small fix rectifies this.

**before**
image

**after**
image


hishamco:
I agree with the change, but can you provide a screenshot for RTL cultures to set the expectation

PR #1812, opened 2021-11-15
Modification to Page Management component 

leigh-pointer:
Add a new tabpane that lists all the module on that page.  From here you are able to modify the module settings and or delete the module from a page.  Delete will send the module to the recycle bin.

image


sbwalker:
@leigh-pointer this is a good enhancement. I will merge it but I will probably make some minor modifications to the labels, etc... for consistency in terminology

leigh-pointer:
@sbwalker cool.

PR #1813, opened 2021-11-15
Modification to Module Management

leigh-pointer:
The component now reports back if the module is in use.  This will assist in housekeeping and removal of unused modules.

image


curlyfro:
this is a nice feature!

Issue #1814, opened 2021-11-17
Implementing ErrorBoundary in ModuleInstance component


chlupac:
This helps to catch unhandled module exception on module instance level. Nice new NET6 feature.
sample module
![image](https://user-images.githubusercontent.com/7838468/142179565-21780885-7f75-4519-96a3-ea7538c45da7.png)

```c#
@inherits ModuleBase

<button type="button" class="btn btn-primary" @onclick="OnClick">Throw exception </button>
@code {
    protected void OnClick()
    {
        throw new Exception("Simulated exception");
    }
}
```

output for users 
![image](https://user-images.githubusercontent.com/7838468/142179973-2f933489-e0a9-428c-9250-6bc36577f8fc.png)

output for host
![image](https://user-images.githubusercontent.com/7838468/142180075-31291b18-88bb-40fc-997e-8f46fa723795.png)

hishamco:
I think this should be closed now

PR #1815, opened 2021-11-17
Implementing ErrorBoundary in ModuleInstance component 

chlupac:
#1814

### DIFF
--- a/Oqtane.Client/Resources/UI/ModuleInstance.resx
+++ b/Oqtane.Client/Resources/UI/ModuleInstance.resx
@@ -123,4 +123,7 @@
   <data name="Error.Module.InvalidType" xml:space="preserve">
     <value>Module Type Is Invalid For {0}</value>
   </data>
+  <data name="Error.Module.Exception" xml:space="preserve">
+      <value>Program error in module {0}</value>
+  </data>
 </root>
--- a/Oqtane.Client/UI/ModuleInstance.razor
+++ b/Oqtane.Client/UI/ModuleInstance.razor
@@ -1,17 +1,30 @@
 @namespace Oqtane.UI
 @inject IStringLocalizer<ModuleInstance> Localizer
 
-<ModuleMessage Message="@_message" Type="@_messagetype" />
-@DynamicComponent
-@if (_progressindicator)
-{
-    <div class="app-progress-indicator"></div>
-}
+<ErrorBoundary>
+    <ErrorContent>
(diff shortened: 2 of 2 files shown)


sbwalker:
@chlupac this is a good enhancement... as it reduces the amount of code that a developer needs to write and it makes the application more resilient to unhandled exceptions. It is also using the new DynamicComponent introduced in .NET 6 which eliminates the need for more primitive methods such as OpenComponent() CloseComponent() etc...

If we consider a typical example of how error handling is currently being done in module razor components:

```
    {
        try
        {
            await Something();
        }
        catch (Exception ex)
        {
            await logger.LogError(ex, "Error {Error}", ex.Message);
            AddModuleMessage(Localizer["InitializeError"], MessageType.Error);
        }
    }
```

You will notice that errors are logged to the event log ( using logger ). This is very important so that there is a permanent record of the issue. I assume it would not be very difficult to add this. 

In addition, the module message usually provides a user friendly message in regards to the event where it occurred ( and this message can be localized to another language ). For example it might say "Error Loading Announcements" or "Error Saving Document". This seems to be a bit more challenging to accomplish as the ErrorBoundary catches any error regardless of where it was thrown within a razor component and there does not appear to be a way to provide any friendly message.  Instead it just says "Program error in module ClassName, AssemblyName" - which is not user friendly and does not provide any info about the specific event which caused it. At the very least the general message would have to be modified to not expose internal system information to end users and instead display the friendly module name instead ie. "Program error in {ModuleState.ModuleDefinition.Name}". However this still does not solve the event granularity info issue. Perhaps the only way to achieve event granularity info is to still include try...catch in each event?

chlupac:
1. ErrorBoundary is the last instance to catch unhanded exceptions,  This is a big step forward, as these errors used to be caught at the application level and a reload of the entire page was required for recovery. This was very confusing for users.
2. Logging is very important for developers in this case, but I haven't fully tested it yet, so I haven't published it yet.
3. The two previous errors of this module, however, are also not logged, this should be changed.
4. For the average user, all that is needed is information that something has happened and that the module is not working as it should. For the time being, I have chosen to list the entire exception for the administrator. If it is logged, a standard error message will suffice.
5. I don't think it's a substitute for a standard try-catch pattern, but in many cases it may be enough for simpler modules.

sbwalker:
I will merge this PR and make some modifications to it to add logging, etc...

PR #1824, opened 2021-11-20
New Component ActionButton with EventCallback

leigh-pointer:
The ActionLink only accepts url where as this new component uses EventCallback allowing hooking up to a OnButtonClick event.


leigh-pointer:
@sbwalker Yes this seems to satisfy, thanks.

sbwalker:
ok excellent! I will close this PR then.

PR #1832, opened 2021-11-24
Settings for ModuleDefinitions #1829

leigh-pointer:
Add Update settings for the ModuleDefinition

### DIFF
--- a/Oqtane.Client/Services/Interfaces/ISettingService.cs
+++ b/Oqtane.Client/Services/Interfaces/ISettingService.cs
@@ -70,7 +70,7 @@ namespace Oqtane.Services
         /// <summary>
         /// Returns a key-value dictionary of all module settings for the given module
         /// </summary>
-        /// <param name="pageId"></param>
+        /// <param name="moduleId"></param>
         /// <returns></returns>
         Task<Dictionary<string, string>> GetModuleSettingsAsync(int moduleId);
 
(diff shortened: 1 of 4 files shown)


sbwalker:
@leigh-pointer after reviewing this I just realized that there may an issue. ModuleDefinitions are stored in the Master database (ie. MasterDBContext) whereas Settings are stored in each Tenant database (ie. TenantDBContext). This PR may appear to work on a default install however as soon as you create another site using its own database you will realize that the settings are not shared across tenants - which essentially breaks your original goal. This is the reason why the SettingService API does not currently have Host Settings - it only has Tenant Settings. In order to implement this feature would require more work.

leigh-pointer:
Ok understand but for me i see it as moddef is set when the mod is first used, therefore if in a new tenant the mod is not used there is no reason to know about thes setup data! Or am I losing the plot!
Are all moduledefs stored in the master? If so, then i agree.
________________________________

sbwalker:
ModuleDefinitions are stored only in the Master - they are global to the installation

leigh-pointer:
Ok the would be good to master something tenant related, or does that break the overall plan?
________________________________

leigh-pointer:
I have a simple solution for now.  I will use the Get/UpdateSiteSettings then I can store the settings when the table is created the firsts time and then the settings are in the correct Tenant.

sbwalker:
I think this enhancement is valuable and I believe I have a solution... so I will merge your PR and make the necessary adjustments

Issue #1837, opened 2021-12-01
RichTextEditor Content not re-Rendering when changed.


leigh-pointer:
I am using the RTE and changing the content via the Content parameter.  The updated content is not being rendered in the RTE control.

In the control itself "RichTextEditor.razor" the Content property does not update the underlying _content string.

`   
    private string _content = string.Empty
`
`    [Parameter]
    public string Content { get; set; }
`

```
RichTextEditorHtml
{Oqtane.Modules.Controls.RichTextEditor}
    Actions: ""
    AllowFileManagement: false
    Content: "<p>New Content added</p>"
    DebugLevel: "info"
    JSRuntime: {Microsoft.AspNetCore.Components.Server.Circuits.RemoteJSRuntime}
    Localizer: {Microsoft.Extensions.Localization.StringLocalizer<Oqtane.Modules.Controls.RichTextEditor>}
    LoggingService: {Oqtane.Services.LogService}
    ModuleInstance: null
    ModuleState: {Oqtane.Models.Module}
    PageState: {Oqtane.UI.PageState}
    Placeholder: "Enter Your Content..."
    ReadOnly: false
    Resources: Count = 3
    SecurityAccessLevel: View
    Theme: "snow"
    Title: ""
    ToolbarContent: null
    UseAdminContainer: true
    _content: "<p><strong>Old Content</strong></p>"
    _editorElement: {Microsoft.AspNetCore.Components.ElementReference}
    _fileManager: null
    _filemanagervisible: false
    _logger: {Oqtane.Modules.ModuleBase.Logger}
    _message: ""
    _original: "<p><strong>Old Content</strong></p>"
    _toolBar: {Microsoft.AspNetCore.Components.ElementReference}
    logger: {Oqtane.Modules.ModuleBase.Logger}
```

sbwalker:
Currently the Content parameter is only being populated in OnInitialized - which means it will only be populated when the component is first created:

```
    protected override void OnInitialized()
    {
        _content = Content; // raw HTML
    }
```

Perhaps it should be using the OnParametersSet event instead.

PR #1841, opened 2021-12-01
Fix RichTextEditor Content not re-Rendering when changed. #1837

leigh-pointer:
Fix Content property not updating when Value changed.
Updating the field on  async Task OnParametersSetAsync() instead of  void OnInitialized()
This fix required some additional code to be executed to keep it in line with the current way of working.


sbwalker:
JavaScript interop should never be used in any event except for OnAfterRenderAsync.

leigh-pointer:
@sbwalker I will create a new PR for this.

PR #1860, opened 2021-12-08
Updated Module and Theme templates to use Code-behind

leigh-pointer:
Module and Theme templates use the Code-behind. This allows a clean separation of the HTML from your presentation logic.


sbwalker:
@leigh-pointer this has come up before and the decision was to not use the code-behind approach

PR #1879, opened 2021-12-15
1877 Module data not being delete when recycle bin is purged

leigh-pointer:
Fixed the permissions validation and added functionality to remove all the settings for the deleted module.


sbwalker:
@leigh-pointer I am going to relocate the setting deletion logic to the repository and follow a pattern consistent with permissions

Issue #1880, opened 2021-12-15
Issue with new SettingService.GetModuleDefinitionSettingsAsync


leigh-pointer:
I am unable to retrieve the ModuleDefinition settings when not logged in.  I can understand it being secure for Adding and Updating but ReadOnly should be available so the ModuleDefinition  settings can be used for the Module instance.

sbwalker:
@leigh-pointer the ModuleDefinition settings are intended to be available to all users - however there is another dimension to consider as well. Similar to Site settings, there is an IsPublic property that must be set to True on a setting if the setting is safe to expose through the API. This allows settings to be properly filtered to avoid sensitive information disclosure.

leigh-pointer:
Ok, will look into it and raise a PR

sbwalker:
It should not require a PR. When you create your ModuleDefinitionSetting you should use the method which allows you to specify the isPublic property:

       public Dictionary<string, string> SetSetting(Dictionary<string, string> settings, string settingName, string settingValue, bool isPublic)

leigh-pointer:
@sbwalker I actually need to raise a PR as the IsPublic was not being set on update.

sbwalker:
It seems to be working in Site Settings? The collection of settings is loaded, the SetSetting() method is used to ensure IsPublic is set, and UpdateSettingsAsync() saves the values.

PR #1882, opened 2021-12-15
#1880 Issue with new SettingService

leigh-pointer:
Needed to update the SettingService to apply the IsPublic field when updating settings for the ModuleDefinition.

### DIFF
--- a/Oqtane.Client/Services/SettingService.cs
+++ b/Oqtane.Client/Services/SettingService.cs
@@ -134,12 +134,19 @@ namespace Oqtane.Services
                 bool ispublic = false;
                 if (value.StartsWith("[Public]"))
                 {
-                    if (entityName == EntityNames.Site)
+                    switch (entityName)
                     {
-                        ispublic = true;
+                        case EntityNames.Site:
+                        case EntityNames.ModuleDefinition:
+                            ispublic = true;
+                            break;
+                        default:
+                            ispublic = false;
+                            break;
                     }
-                    value = value.Substring(8); // remove [Public]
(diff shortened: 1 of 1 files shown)


sbwalker:
Thank you @leigh-pointer ... to provide a bit more background, the SettingService relies on a standard Dictionary type (ie. name/value pairs) to make the API as simple as possible for developers. However this makes it difficult to support additional metadata for settings without breaking the API. So the IsPublic property is "supported" by embedding the metadata into the Dictionary value and then parsing/interpreting the value during load/save. This makes the internal code of the SettingService less intuitive and more complex - but it does shield the consumer of the API from these internal complexities. 

The bug you ran into is caused by logic which is trying to be very specific about the specific types of settings which support the IsPublic property. Originally it was introduced for Site Settings - and the use case was that the SMTP details for a site are stored in Site Settings - however that information is sensitive and should not be disclosed to non-Administrators - so there needed to be a mechanism to be able to filter Site Settings for standard users. The IsPublic concept accomplishes this... however there is currently a drawback to the implementation:

The IsPublic concept is currently only enabled for some setting types and not others. This can cause issues if a developer tries to use the SetSetting IsPublic overload method for types other than Site - which is the scenario you ran into for ModuleDefinition. The problem is that the supported types are static - but the SettingService is intended to be dynamic.

One possible solution is to make the IsPublic concept applicable to ALL setting types. This would eliminate the hard-coded logic for checking if a specific type supports it or not. What are your thoughts?

leigh-pointer:
@sbwalker I think the approach you are outlining sounds more intuitive and that is always a good thing.

PR #1848, opened 2021-12-03
Fix #1837 RichTextEditor Content not re-Rendering

leigh-pointer:
Change to the OnAfterRenderAsync method and changed OnInitialized to OnParametersSet

PR #1842, opened 2021-12-02
Quill Security related bug fixes.

leigh-pointer:
Upgraded Quill references to 1.3.7
Tabnabbing vulnerability in snow theme #2438
https://github.com/quilljs/quill/issues/2438

https://github.com/quilljs/quill/releases/tag/v1.3.7

PR #1883, opened 2021-12-16
Rework to #1848 RawHTML not being saved

leigh-pointer:
Restructured the execution of code.
RawHTML now works as it did in previous versions as well as the new functionality.

### DIFF
--- a/Oqtane.Client/Modules/Controls/RichTextEditor.razor
+++ b/Oqtane.Client/Modules/Controls/RichTextEditor.razor
@@ -120,11 +120,6 @@
         new Resource { ResourceType = ResourceType.Script, Bundle = "Quill", Url = "js/quill-interop.js" }
     };
 
-    protected override void OnInitialized()
-    {
-        _content = Content; // raw HTML
-    }
-
     protected override async Task OnParametersSetAsync()
     {
         _content = Content; // raw HTML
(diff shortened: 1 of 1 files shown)


sbwalker:
@leigh-pointer the RichTextEditor still has issues with these latest changes. To reproduce the problem, simply install the Oqtane.Blogs module and create a blog entry and then click the Edit button to edit the entry. The following line of code throws an unhandled exception:

            // preserve a copy of the rich text content ( Quill sanitizes content so we need to retrieve it from the editor )
            _original = await interop.GetHtml(_editorElement);

The problem is that _editorElement is null - which happens on a re-render because the logic to set the _editorElement variable is located inside the if firstrender block. I moved the logic above inside the if block (which is where it was located previously) and it resolved the problem.

My concern however is that I was never able to reproduce the original problem which your previous PR for #1837 was attempting to resolve... so I cannot determine if it resolves all scenarios.

leigh-pointer:
@sbwalker I have had a quick look and moving the code seems to continue to work with my original PR
However the Blog module references
```
 public override List<Resource> Resources => new List<Resource>()
    {
        new Resource { ResourceType = ResourceType.Stylesheet, Url = "css/quill/quill1.3.6.bubble.css" },
        new Resource { ResourceType = ResourceType.Stylesheet, Url = "css/quill/quill1.3.6.snow.css" }
    };
```

These resources were removed when Quill was updated to 1.3.7 Update Quill component to latest version #1842 and remove quill 1.3.6 assets #1866

sbwalker:
Yes I noticed that as well - renaming the CSS resources was a breaking change. The solution going forward is to rename the Quill CSS and JS files so they do not identify a specific version number.

Issue #1914, opened 2022-01-08
Label Control appending Class to LabelClass and SpanClass


leigh-pointer:
When modifying the Label control at runtime the classes are appended to the Class field which in turn does not render the correct class. 

```
    private string _spanclass = "app-tooltip";
    private string _labelclass = "form-label";

 _spanclass += (!string.IsNullOrEmpty(Class)) ? " " + Class : "";
_labelclass += (!string.IsNullOrEmpty(Class)) ? " " + Class : "";
```

I have toggled the Label from Bold to Normal text and the results are as follows:

`<label for="Text3" class="form-label fw-normal fw-bold fw-normal fw-bold fw-normal fw-bold fw-normal">Text3</label>`

I will submit a PR for this issue.

sbwalker:
The classes on the label field are not intended to be modified at runtime - they are intended to be specified statically in your razor code.

sbwalker:
My apologies... after looking at the PR I now understand the problem. The PR looks good - I will merge it.

PR #1917, opened 2022-01-08
Fix for  #1914 Label Control appending Class to LabelClass

leigh-pointer:
Modified so that the Class parameter is not constantly appended when a new Class is applied.


sbwalker:
technically you do not need to declare a local var in the method...

```
            var spanclass = (!string.IsNullOrEmpty(Class)) ? " " + Class : "";
            _spanclass = "app-tooltip" + spanclass;
```

could be rewritten as:

```
            _spanclass = "app-tooltip" + (!string.IsNullOrEmpty(Class)) ? " " + Class : "";
```

which would save some memory allocation.

PR #1969, opened 2022-01-29
fix #1272 - add support for refs folder in package installation

### LINKED ISSUE #1272: Nuspec to include 'refs'

tvatavuk:
Please add support also for `refs` folder in package installation.

My understanding is that Microsoft AspNetCore dependencies that are stored in the "refs" folder are an essential part of the default compilation context to support run-time compilation scenarios in https://github.com/2sic/2sxc/ or any other module or app that targets standard Microsoft.NET.Sdk.Web.

In oqtane source development environment, "refs" are deployed by default when `Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation` nuget is included as project reference and property `PreserveCompilationContext` is set in `csproj`.

For oqtane release standalone installations, we need to manually deploy "refs" and that is the reason why it will be great to include it as part of 2sxc oqtane module nuget, and that will make life easier for anyone installing https://github.com/2sic/2sxc/.

### DIFF
--- a/Oqtane.Server/Infrastructure/InstallationManager.cs
+++ b/Oqtane.Server/Infrastructure/InstallationManager.cs
@@ -125,6 +125,9 @@ namespace Oqtane.Infrastructure
                                 case "ref": // ref/net*/...
                                     filename = ExtractFile(entry, Path.Combine(binPath, "ref"), 2);
                                     break;
+                                case "refs": // refs/net*/...
+                                    filename = ExtractFile(entry, Path.Combine(binPath, "refs"), 2);
+                                    break;
                             }
 
                             if (filename != "")


sbwalker:
@tvatavuk thank you for submitting this PR.... before I merge it I want to understand the difference between the /ref folder (which is supported in the Nuget spec) and the /refs folder (which is created during a .NET publish operation). The fact that they are similarly named and have a similar purpose is very confusing.

tvatavuk:
Thx @sbwalker. Same _situation_ was here. I been _there_ before some time. I will try to share what I find...

Both folders "ref" and "refs" contains "_[reference assemblies](https://docs.microsoft.com/en-us/dotnet/standard/assembly/reference-assemblies)_" (dll's with only public interface signature, necessary for compiler, while full concrete implementation are in _real_ dll's_ that are somewhere else).

"ref" in nupkg will contain reference assemblies for dll's that are distributed by that nupkg. Goal is that third party code that has that nupkg as dependency can be compiled, without including full dll's from dependent nupkg.

"refs" is special case, not expected to be used in nupkg, but with Oqtane we expands a lots of boundaries :-), that contains [Microsoft.AspNetCore.App.Ref](https://www.nuget.org/packages/Microsoft.AspNetCore.App.Ref/) reference assemblies that are needed to [compile razor pages (or webapi, cs's) only in run-time](https://docs.microsoft.com/en-us/aspnet/core/razor-pages/sdk?toc=%2Fdotnet%2Ffundamentals%2Ftoc.json&bc=%2Fdotnet%2Fbreadcrumb%2Ftoc.json&view=aspnetcore-6.0#runtime-compilation-of-razor-views), because end user (razor or cs developer) expect to have whole aspnetcore web/razor framework for reuse while running Oqtaned based solution.

PR #1973, opened 2022-01-31
fix #1972 Missing Preprocessor Directives during Runtime Compile

### LINKED ISSUE #1972: Missing Preprocessor Directives during Runtime Compile
.net 6 has a breaking change which drops preprocessor values in the `.deps` JSON file

https://docs.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/6.0/preservecompilationcontext-not-set-by-default

This causes trouble for runtime compile, because any code that runs on multiple platforms needs things like this: 

```cs
#if NETCOREAPP
 // .net core code
#else
  // .net Framework code
#endif
```

In Oqtane 2 this code did what you would expect, but in Oqtane 3 it doesn't because the runtime compiler doesn't set `NETCOREAPP`.

We believe that this is essential, and ideally would also define additional keys like `OQTANE` and `OQTANE3` so that code could be different if built for Oqtane vs. NopCommerce or version differences in Oqtane. That would then also allow code like:

```cs
#if OQTANE3
 // oqtane 3 code
#else
  // oqtane 2 code
#endif
```

Our recommendation

1. Enable the default .net keys to be added in the `.deps` file according to the [list by Microsoft](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives)
2. Also add an `OQTANE` key to make development easier for cross-platform components
3. Also add an `OQTANE3` key to make development easier for cross-version components, which may have to survive breaking changes in the platform
4. Optionally also add an `OQTANE3_0_OR_GREATER` which would be consistent with MS recommendations

@tvatavuk will create a pull-request for these changes to help you review if this makes sense to you.

tvatavuk:
fix https://github.com/oqtane/oqtane.framework/issues/1972, enables again missing preprocessor directives for runtime compile

### DIFF
--- a/Oqtane.Server/Oqtane.Server.csproj
+++ b/Oqtane.Server/Oqtane.Server.csproj
@@ -16,6 +16,8 @@
     <RepositoryType>Git</RepositoryType>
     <RootNamespace>Oqtane</RootNamespace>
     <IsPackable>true</IsPackable>
+    <DefineConstants>$(DefineConstants);OQTANE;OQTANE3;OQTANE3_0_OR_GREATER </DefineConstants>
+    <PreserveCompilationContext>true</PreserveCompilationContext>
   </PropertyGroup>
   <ItemGroup>
     <Compile Remove="wwwroot\Modules\Templates\**" />


sbwalker:
The framework does not require preprocessor directives, so I am not sure how they are missing? I do not understand the rationale behind the constants... and I also do not understand how to test this PR.

tvatavuk:
Your are right, the Oqtane Framework does not require preprocessor directives so it can not be missing 👍. 

Actually Oqtane Framework 👌 is one of the best projects in open source community, and for me personally in TOP 5 🦄 in .NET eco system, so I follow each new version with un-hidden excitement and looking forward to see it everywhere 🤩. Huge respect 👍. 

I see that my comment was imprecise and vague. I will try to improve it. As always no bad intentions 😀.

In Oqtane 3.x we are missing precompiled symbols like `NETCOREAPP` that we are using in 2sxc [hybrid razor](https://docs.2sxc.org/net-code/razor/hybrid/index.html#the-preprocessor-directives-in-cs-files) files. Oqtane Framework is not the reason, but breaking change in .NET 6 [PreserveCompilationContext not configured by default](https://docs.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/6.0/preservecompilationcontext-not-set-by-default).

We where looking what is needed to enable again precompiled symbols in run-time for 2sxc hybrid razor solutions in Oqtane Source, and Oqtane release installations.

For Oqtane Source, the most simple is change in `Oqtane.Server.csproj` ...
```xml
<PropertyGroup>
   <PreserveCompilationContext>true</PreserveCompilationContext>
</PropertyGroup>
```

For Oqtane release install, we need to deploy:
- Oqtane.Server.deps.json (and related json files) that contain "compilationOptions" with "defines" like...
```json
  "compilationOptions": {
    "defines": [
      "TRACE",
      "DEBUG",
      "NET",
      "NET6_0",
      "NETCOREAPP",
      "NET5_0_OR_GREATER",
      "NET6_0_OR_GREATER",
      "NETCOREAPP1_0_OR_GREATER",
      "NETCOREAPP1_1_OR_GREATER",
      "NETCOREAPP2_0_OR_GREATER",
      "NETCOREAPP2_1_OR_GREATER",
      "NETCOREAPP2_2_OR_GREATER",
      "NETCOREAPP3_0_OR_GREATER",
      "NETCOREAPP3_1_OR_GREATER"
    ],
    "languageVersion": "10.0",
    "platform": "",
    "allowUnsafe": false,
    "warningsAsErrors": false,
    "optimize": false,
    "keyFile": "",
    "emitEntryPoint": true,
    "xmlDoc": false,
    "debugType": "portable"
  },
```

In same PR we proposed defining precompile symbols like: OQTANE, OQTANE3, OQTANE3_0_OR_GREATER because while developing 2sxc we noticed that we need different versions of 2sxc code if we would like to support uncompatilbe parts of Oqtane2 and Oqtane3 (this are special cases that can not be supported with reference to single version of Oqtane framework nugets, or runtime checking oqtane version in oqtane share). The most simple way we see to include the both versions of code in same source (without compilation errors) is with using preprocessor directives and precompile symbols like: OQTANE, OQTANE3. It is not the nicest code, but it is serving the purpose.

Building of Oqtane.Server.csproj will result with `Oqtane.Server.deps.json` near Oqtane.Server.dll that has "compilationOptions" with "defines" like above (possible including also new precompile symbols like: OQTANE, OQTANE3, OQTANE3_0_OR_GREATER).

I would like to help with testing this PR, in any way I can.

PR #2005, opened 2022-02-18
Theme Template Updated

leigh-pointer:
Bootstrap reference updated to 5.1.3


leigh-pointer:
The CDN for Bootstrap is the same as what is used in the Oqtane Theme, or have i missed something?

leigh-pointer:
I see it now. I will fix.

PR #2130, opened 2022-04-13
Update to User Management 

leigh-pointer:
Updated user management to display more of the User ; Last Login Last IP and Is Authenticated.

image

### DIFF
--- a/Oqtane.Client/Modules/Admin/Users/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Users/Index.razor
@@ -38,6 +38,9 @@ else
 				<th style="width: 1px;">&nbsp;</th>
 				<th>@SharedLocalizer["Name"]</th>
 				<th>@SharedLocalizer["Username"]</th>
+				<th>@Localizer["LastLoginOn"]</th>
+				<th>@Localizer["LastIPAddress"]</th>
+				<th>@Localizer["IsAuthenticated"]</th>
 				</Header>
 				<Row>
 					<td>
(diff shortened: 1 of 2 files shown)


sbwalker:
@leigh-pointer I will merge this but I will remove the IsAuthenticated column because it does not represent what you likely assume it is... it does not indicate if a user is currently logged in... it is a transient property which is only relevant for the current user.

leigh-pointer:
@sbwalker I thought the IsAuthenticated was that the account was authenticated as the email had been verified.  I think having this information is handy also with the option to resend confirmation email ( something I was going to work on).  Thoughts?

sbwalker:
No, IsAuthenticated does not have anything to do with email validation... it is property on the User model which is not persisted anywhere except for in memory and is only populated for the current user based on whether they are authenticated (ie. signed in). 

Email validation is managed by .NET Core Identity. In the case where a site has enabled User Registration (which is disabled by default) and a new user registers on the site, a notification is sent to the user asking them to validate their email address. The validation link uses .NET Core Identity to construct a token and this token is passes to the User API. Once a user clicks the link, their EmailConfirmed field is updated in the AspNetUsers table. This EmailConfirmed field is used during login by .NET Core Identity to verify that a user already validated their email address. The EmailConfirmed property is not included in Oqtane's User model - it is only managed server-side. I would prefer to keep AspNetUsers decoupled from Users for performance and abstraction reasons. It would still be possible to include a Resend option for email verification - but it would not be dependent upon on EmailConfirmed property (it can sort of be assumed that if a user has never logged in it may be because they have not verified their email). However I would not include the Resend option in the main Users grid view - I would include it in the detailed modal view for a specific user. And I would only make the option visible if the User Registration option is enabled - as it makes no sense to display it if User Registration is disabled.

The Is Authenticated column should be removed from the main User Management view. However I would suggest adding the Users CreatedOn date so that the Admin can easily see how long the user has been a member of the site?

PR #2152, opened 2022-04-22
completed antiforgery implementation, improved external login claim mapping, principal construction, and user experience

sbwalker:


leigh-pointer:
Nice, got some translation work :) i see.

PR #2161, opened 2022-04-27
Fix #2160 - upgrade module option

### LINKED ISSUE #2160: Upgrade module option in Module Management is downloading an already installed version of module
There is a small issue with the UPGRADE module option in Module Management that makes this option unusable.

image

When there is a newer version of the oqtane module available for upgrade on oqtane.net/nuget.org user see [Upgrade] button.
If user clicks on [Upgrade] button it is expected that the newer/latest version of the module package is downloaded and installed.
Instead, the already installed version of the module is downloaded again and installed. 
The reason is that the currently installed module package version is provided to download code instead newer/latest version.
The fix should be simple.

tvatavuk:
provide newer/latest module package version (instead of already installed version) to download code

### DIFF
--- a/Oqtane.Client/Modules/Admin/ModuleDefinitions/Index.razor
+++ b/Oqtane.Client/Modules/Admin/ModuleDefinitions/Index.razor
@@ -50,9 +50,9 @@ else
                 @((MarkupString)PurchaseLink(context.PackageName))
             </td>
             <td>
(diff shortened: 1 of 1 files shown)


sbwalker:
@tvatavuk thank you for reporting this issue. I am going to merge your PR but I am going to refactor it as I try to avoid using "out" parameters as a general rule in the framework.

PR #2178, opened 2022-05-05
Added the User Email field to the List

leigh-pointer:
Added the formatted email address of the user to the list view.


sbwalker:
@leigh-pointer I like this improvement - however the more columns that are added to the main User view, the more the mobile experience gets degraded as there is too much information to display horizontally without a lot of scrolling. This is the reason why most of the main admin view components in Oqtane have minimal columns -  and Edit will display the details in a vertical orientation where it is easier to navigate. This will become even more important once Oqtane runs in MAUI and there will be mobile users. If we add email to the main view perhaps some of the other fields should be removed? What information is absolutely needed 90% of the time... and what information is better to include in the Edit view?

PR #2187, opened 2022-05-09
fix #2182 - modifications to address MySQL compatibility issues

### LINKED ISSUE #2182: Installation with MySQL fails.
A fresh installation of oqtane 3.1.1 (from May 5, 2022) with mysql database fails.

Error on the screen: An Error Occurred Executing Upgrade Logic. Unknown column 's.CaptureBrokenUrls' in 'field list'

Contents of error.log
[2022-05-06 10:03:43+00:00] [Error] [Oqtane.Infrastructure.DatabaseManager] An Error Occurred Migrating A Tenant Database. This Is Usually Related To A Tenant Database Not Being In A Supported State. Cannot drop index 'IX_Folder': needed in a foreign key constraint
[2022-05-06 10:03:43+00:00] [Error] [Oqtane.Infrastructure.DatabaseManager] An Error Occurred Executing Upgrade Logic. Unknown column 's.CaptureBrokenUrls' in 'field list'

Prior to the installation, I created an empty database in MySQL. This database is filled with tables during the installation process, but not as much as with a successful installation in SqlServer.
It seems that the installation fails when processing migration 2.01.01 (removing IX_Folder index).

Database used:
MySQL version 8.0.15 and 8.0.29 in a docker container.
The installation with SQLite and SqlServer does work.

Question is, what am I doing wrong.
tables_mysql
tables_sqlserver

sbwalker:


Raceeend:
Thanks, it works like a charm ;-)

PR #2188, opened 2022-05-09
GetTenant{} returns null in some Blazor server scenario cases 

chlupac:
Because SiteState.Alias is null.

### DIFF
--- a/Oqtane.Server/Infrastructure/TenantManager.cs
+++ b/Oqtane.Server/Infrastructure/TenantManager.cs
@@ -62,7 +62,7 @@ namespace Oqtane.Infrastructure
 
         public Tenant GetTenant()
         {
-            var alias = _siteState?.Alias;
+            var alias = _siteState?.Alias ?? GetAlias();
             if (alias != null)
             {
                 // return tenant details


sbwalker:
@chlupac to provide some more clarification... 

TenantManager.GetAlias() should be used in the vast majority of cases when dealing with tenancy requirements. Note that it relies on HttpContext - however it can also be set manually if required. Note that the Alias object contains a TenantId property which can be used to identify the specific tenant.

TenantManager.GetTenant() should not be used - the only scenario where it is used within the framework is within DBContextBase to get the ConnectionString for the Tenant - and in this case it must rely on the SiteState service.

chlupac:
This happens during DBContext initialization when you try to open DBContext directly on server-side blazor component. 
For example when I'm using repositories directly from server-side components. It works well in previous version OF.

Stack trace:
System.InvalidOperationException: No database provider has been configured for this DbContext. A provider can be configured by overriding the 'DbContext.OnConfiguring' method or by using 'AddDbContext' on the application service provider. If 'AddDbContext' is used, then also ensure that your DbContext type accepts a DbContextOptions<TContext> object in its constructor and passes it to the base constructor for DbContext.
   at Microsoft.EntityFrameworkCore.Internal.DbContextServices.Initialize(IServiceProvider scopedProvider, DbContextOptions contextOptions, DbContext context)
   at Microsoft.EntityFrameworkCore.DbContext.get_ContextServices()
   at Microsoft.EntityFrameworkCore.DbContext.get_Model()
   at Microsoft.EntityFrameworkCore.Internal.InternalDbSet`1.get_EntityType()
   at Microsoft.EntityFrameworkCore.Internal.InternalDbSet`1.CheckState()
   at Microsoft.EntityFrameworkCore.Internal.InternalDbSet`1.get_EntityQueryable()
   at Microsoft.EntityFrameworkCore.Internal.InternalDbSet`1.System.Linq.IQueryable.get_Provider()
   at System.Linq.Queryable.Where[TSource](IQueryable`1 source, Expression`1 predicate)
   **at Oqtane.Repository.FolderRepository.GetFolder(Int32 siteId, String path) in W:\projects\mfl\lib\oqtane.framework\Oqtane.Server\Repository\FolderRepository.cs:line 79**
   at Mfl.Modules.MyModuleBase.GetFolder(String path, Boolean forceCreate) in W:\projects\mfl\src\Mfl.Server\Modules\MyModuleBase.cs:line 114
   at Mfl.Modules.Events.EventDetailBase.OnInitialized() in W:\projects\mfl\src\Mfl.Server\Modules\Events\EventDetailBase.cs:line 120

sbwalker:
@chlupac I would need to see a source code example of what you are referring to, as none of the admin modules in the core framework have this issue, nor do custom modules created from the Module Creator template... which means your custom modules are implemented differently. This is not necessarily a problem - the issue is that unless I am able to reproduce the issue and validate the root cause, it is risky to merge this PR.

chlupac:
Solved with service replacement.

Issue #2209, opened 2022-05-22
Localization Pager component (Page @_page of @_pages)


InfDev:
In the Pager component (qtane.Client\Modules\Controls file) for
Page @_page of @_pages
does not apply resource for localization capability

leigh-pointer:
Creating a PR now.

InfDev:
Oqtane.Client\Resources\Modules\Controls\Pager.resx (add resource with key "PageOfPages")
In Pager.razor:
1) @Inject IStringLocalizer<Pager> Localizer
2)  Localizer["PageOfPages", _page, _pages]

leigh-pointer:
@InfDev 
That does not work and errors with

```
An unhandled exception occurred while processing the request.
InvalidOperationException: Cannot provide a value for property 'Localizer' on type 'Oqtane.Modules.Controls.Pager`1[[Oqtane.Models.Page, Oqtane.Shared, Version=3.1.2.0, Culture=neutral, PublicKeyToken=null]]'. There is no registered service of type 'Microsoft.Extensions.Localization.IStringLocalizer'.
Microsoft.AspNetCore.Components.ComponentFactory+<>c__DisplayClass7_0.<CreateInitializer>g__Initialize|1(IServiceProvider serviceProvider, IComponent component)
```

sbwalker:
When you inject a localizer  you need to specify the name of the razor component:

```
@inject IStringLocalizer<Pager> Localizer

```

leigh-pointer:
@sbwalker correct but the component wont load because of the  class needs to specified as IStringLocalizer<Pager<TableItem>> which does not work.

sbwalker:
Why doesn't it work? Other controls like PermissionGrid use this approach and it works fine

leigh-pointer:
@sbwalker 
`@inject IStringLocalizer<PermissionGrid> Localizer`
`@inject IStringLocalizer<Pager> Localizer` is not valid as it requires the Pager to be initialized with the `@typeparam TableItem`

So this means the inject would look like
`@inject IStringLocalizer<Pager<TableItem>>
`
When doing this the IStringLocalizer is not injected.

InfDev:
I bypassed the localization problem for the class Pager<TableItem>.
The code [here](https://github.com/InfDev/oqtane.translations/tree/main/Oqtane.Client.OnlyPagerComponent)

P.S.

[...]

PR #2210, opened 2022-05-22
Fix for #2209 Localization Pager component (Page @_page of @_pages) 

leigh-pointer:
Added fixe for issue.  Added the resource "PageOfPages" to SharedResources as trying to inject IStringLocalizer<Pager<TableItem>> and adding Pager.resx the resource failed to load.

### DIFF
--- a/Oqtane.Client/Modules/Controls/Pager.razor
+++ b/Oqtane.Client/Modules/Controls/Pager.razor
@@ -1,5 +1,6 @@
 @namespace Oqtane.Modules.Controls
 @inherits ModuleControlBase
+@inject IStringLocalizer<SharedResources> Localizer
 @typeparam TableItem
 
 @if (ItemList != null)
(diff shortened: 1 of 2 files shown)


hishamco (approved):
LGTM

sbwalker:
@hishamco I am going to merge this PR but there is a localization issue which @leigh-pointer was trying to describe above... which is why he ended up implementing this via SharedResources rather than the standard method. Let me explain the problem and you can let us know if this should be logged as a .NET issue (since you are heavily involved in .NET localization)...

If you have a Razor component with a TypeParam (Blazor components can actually have multiple TypeParam declarations but a single declaration is most common):

@typeparam TableItem

And then you would like to use Localization within that same component. So you need to declare IStringLocalizer - however in this case you cannot simply specify the component name - you also need to specify the TypeParam:

@inject IStringLocalizer<Pager<TableItem>> Localizer

This code compiles fine. Assume that a resx file has been created in the appropriate location and wired up correctly (ie. \Resources\Modules\Controls\Pager.resx). Howeverm when you try to reference resource strings by name within the resx file from within your component, it does not load the translated value correctly. 

Localizer["KeyName"]

I believe this is because the TypeParam causes StringLocalizer some confusion in identifying the resx file based on convention. Do you have any idea if this should be considered a bug?

hishamco:
Thanks a lot for you explaination Shaun, please don't merge this PR. Regarding the above issue it remindes me with something relates especially in Blazor.

> Note that an Oqtane community member @InfDev came up with a custom solution for this using reflection and IStringLocalizerFactory:

The solution may it works fine, but I'm still think something goes wrong here, no need to create another localizer through a factory while there's already one registed in the DI

So, I suggest give me some time to investigate on this and come up with a solution

hishamco:
Sure

PR #2223, opened 2022-05-27
Avoid toggle password & confirm password as same time when one button is clicked

hishamco:


sbwalker:
I do not understand the actual problem you are trying to fix with this PR? The current user experience is that when a user chooses the Show Password button it displays the values of the Password and Confirm fields. This is by design as the reason why a user would use the Show Password option in this specific scenario is because the passwords they entered do not match - which is a requirement for installation. So it makes no sense to make them click 2 separate buttons.

hishamco:
> So it makes no sense to make them click 2 separate buttons.

If it's like that there's no need for two buttons, one button is enough. If you have a look to some browsers each password field has it's icon to hide / show the password

The actual issue is UI/UX why I should click any of the buttons to hide / show both password and confirm password

sbwalker:
Ok I understand your point... I will merge this PR

PR #2228, opened 2022-05-31
Log notification improvement

chlupac:
Recently my inbox is spammed with messages like:

Subject: Site Error Notification
Please visit https://localhost:44356/admin/log for more information

Url opens whole log.

Improvement opens concrete log item

Subject: Site Error Notification
Please visit https://localhost:44356/admin/log?id=30231 for more information

### DIFF
--- a/Oqtane.Server/Infrastructure/LogManager.cs
+++ b/Oqtane.Server/Infrastructure/LogManager.cs
@@ -207,7 +207,7 @@ namespace Oqtane.Infrastructure
                 {
                     if (userrole.Role.Name == RoleNames.Host)
                     {
-                        var url = _accessor.HttpContext.Request.Scheme + "://" + _tenantManager.GetAlias().Name + "/admin/log";
+                        var url = $"{_accessor.HttpContext.Request.Scheme}://{_tenantManager.GetAlias().Name}/admin/log?id={log.LogId}";
                         var notification = new Notification(log.SiteId.Value, userrole.User, "Site " + log.Level + " Notification", "Please visit " + url + " for more information");
                         _notifications.AddNotification(notification);
                     }
--- a/Oqtane.Client/Modules/Admin/Logs/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Logs/Index.razor
@@ -5,7 +5,11 @@
 @inject IStringLocalizer<Index> Localizer
 @inject IStringLocalizer<SharedResources> SharedLocalizer
 
-@if (_logs == null)
+@if (_id > 0)
+{
+    <Detail/>
+}
+else if (_logs == null)
 {
     <p><em>@SharedLocalizer["Loading"]</em></p>
 }
@@ -91,6 +95,7 @@ else
 }
 
 @code {
+    private int _id = 0;
     private string _level = "-";
     private string _function = "-";
     private string _rows = "10";
@@ -104,6 +109,10 @@ else
     {
         try
         {
+            if (PageState.QueryString.ContainsKey("id") && int.TryParse(PageState.QueryString["id"], out int id))
+            {
+                _id = id;
+            }
 			if (PageState.QueryString.ContainsKey("level"))
 			{
 				_level = PageState.QueryString["level"];


Oqtane.Server/Infrastructure/LogManager.cs
```diff
@@ -207,7 +207,7 @@ private void SendNotification(Log log)
                 {
                     if (userrole.Role.Name == RoleNames.Host)
                     {
-                        var url = _accessor.HttpContext.Request.Scheme + "://" + _tenantManager.GetAlias().Name + "/admin/log";
+                        var url = $"{_accessor.HttpContext.Request.Scheme}://{_tenantManager.GetAlias().Name}/admin/log?id={log.LogId}";
```
hishamco, line 210: If you specifiy the id here how the log will show all if the id not specified?

chlupac: why? you want to see concrete item from notification email.

sbwalker:
@chlupac when you made the request to add notifications for event log items (#2014) I indicated that one of the risks is that your inbox may get flooded with messages. By default the framework is set to only notify on "Error" conditions - and it does have a setting to disable the notifications for event log notifications if necessary. If you are getting flooded with messages you should probably try to identity why the error is happening (which is ultimately the purpose of this feature - to let you know when errors are occurring in production). I think this PR is a good enhancement - including the ID of the event log item in the Url sent in the notification allows you to immediately view the specific detailed error message.

Oqtane.Server/Infrastructure/LogManager.cs
hishamco: Oops, I didn't notice that the change inside `SendNotifications`

chlupac:
@sbwalker I believe that it is necessary to determine what and at what level it will be recorded in the log. Many errors that are logged can be caused by the user modifying the URL parameters in the browser or by incorrect input. I would either ignore these errors or record them at a lower level. 
On the other hand, I am particularly interested in exceptions, I need to know about them immediately, in best case in notification body.

hishamco (changes requested):
LGTM except a format issue

chlupac:
The most common errors in my log that are not so critical that I have to be informed by email and which can be used in attack:

User Login Failed .......
File Does Not Exist .....
Page Does Not Exist Or User Is Not Authorized To View Page ....
Unauthorized File Get Attempt  .....


I like idea to include details of log item in email.

sbwalker:
@chlupac I reviewed the PR and I have a few observations and requests....

We are trying to make it simple and convenient for a host user to be alerted to potential issues in their installation. The current system sends email notifications however it does not provide any details in the email. Instead it simply includes a link and forces the user to navigate to the Event Log. However it does not redirect them to the specific item which triggered the notification - it shows them all event log entries.

I believe there should be 2 improvements:

1. LogManager - should include the Event Log "Message" field value in the notification. It should also include a link to the specific Event Log item (using "id=" as you have already done) as there is a variety of additional context which is available in the Event Log which will not be included in the email notification (on purpose).

2. Admin/Logs/Index - should support a QueryString parameter of "id" - however the implementation should follow the framework conventions and use EditUrl(). Basically if the Url contains an "id" QueryString parameter it should call:

```
if (PageState.QueryString.ContainsKey("id") && int.TryParse(PageState.QueryString["id"], out int id))
{
    NavigationManager.NavigateTo(EditUrl(PageState.Page.Path, ModuleState.ModuleId, "Detail", "id={id}"))
}
```

chlupac:
Meanwhile I made some improvements in topic.

1. Include alias and log severity in subject. If exception field is not null severity is displayed as exception - for email sorting rules
2. Include log item serialized to formatted json

For my comfort I like to see these items in notification:
{
  "LogDate": "2022-06-03T10:11:18.9996378Z",
  "Url": "https://qqqqqq.qq/aaaaaaa",
  "Category": "Mfl.Modules.Register.Index, Mfl.Server.oqtane, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null",
  "Message": "Error Adding User ......",
  "Exception": "Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred  ..........." 
}

sbwalker:
I will merge and make the changes I requested above

PR #2232, opened 2022-06-06
missing translation keys for module names

ijaz-saeed:
without these keys, following module names are not translating, 

image

### DIFF
--- a/Oqtane.Client/Resources/SharedResources.resx
+++ b/Oqtane.Client/Resources/SharedResources.resx
@@ -330,4 +330,10 @@
   <data name="PageOfPages" xml:space="preserve">
     <value>Page {0} of {1}</value>
   </data>
+  <data name="Url Mappings" xml:space="preserve">
+    <value>Url Mappings</value>
+  </data>
+  <data name="Visitor Management" xml:space="preserve">
+    <value>Visitor Management</value>
+  </data>
 </root>


hishamco:
Why you put them in shared resources?

hishamco:
I'm expecting `ShareResources` for that can be used in multiple places

ijaz-saeed:
> Why you put them in shared resources?

as @leigh-pointer explained, all the other items in the image above are coming from ShareResources, so I put these there. But please let me know any other place and I will move it there. thanks @hishamco

hishamco:
It's appropriate if we are using it in different places like Cancel, Delete ... etc, otherwise I suggest to move it to particular resource where it lives

sbwalker:
Each site contains a default set of Administrative pages. The names of these Administrative pages as well as titles of the modules on these pages use standard names such as "Site Settings" or "Page Management". When a user is interacting with the framework, they expect that these page names and module titles are displayed in their preferred language. The page names are displayed in the Admin Dashboard (ie. \Modules\Admin\Dashboard\Index.razor ). The module titles are displayed within a Container (ie. \Themes\Controls\Container\ModuleTitle.razor ). So the same localized strings are used in 2 different locations in the framework - so they belong in SharedResources.

hishamco:
Make sense

hishamco (approved):

PR #2243, opened 2022-06-15
added extension method for creating a LocalizerFactory using a type name, refactored Pager and LocalizableComponent to use LocalizerFactory

sbwalker:


sbwalker:
@hishamco this refactoring introduces an extension method to IStringLocalizerFactory so that it is possible to create an IStringLocalizer based on a type name. The method is able to handle generic type parameters as well as root namespaces. I then refactored Pager to use this method, as I believe the standard approach recommended by Microsoft is to use IStringLocalizerFactory in those scenarios where IStringLocalizer is not capable of meeting a developer's needs. I also refactored LocalizableComponent to use this new extension method (as it was a good fit for this case as well) and replaced the ServiceFactory logic with the IStringLocalizerFactory service which is already available.

Oqtane.Client/Modules/Controls/Pager.razor
```diff
@@ -1,6 +1,6 @@
 @namespace Oqtane.Modules.Controls
-@inherits LocalizableComponent
-
+@inherits ModuleControlBase
+@inject IStringLocalizerFactory LocalizerFactory
```
hishamco, line 3: This is the first time ever I see someone inject the `IStringLocalizerFactory` instead of `IStringLocalizer`

hishamco:
While you already merged this PR, no need to add comments ;), but the solution should works fine

PR #2272, opened 2022-07-06
Support for module editors by exposing Edit Mode in the Control Panel

sbwalker:


leigh-pointer:
@sbwalker brilliant :)

Issue #2277, opened 2022-07-10
Environment Variable


Sen-Gupta:
Hello Oqtane community,

We containerized and tried to pass connectionString from environment variables. 

The observation is that environment  variables are picked by containers but not being honoured by Oqtane.

It seems it is becuase of ConfigManager which just tries to read values from appsettings.

Please let me know if, I got it correctly.

Is there a reasoning behind not honouring environment variables?

Rodien:
Please provide more information.

How are you storing and passing the environment variables?

Are you using azure? 

Show a code snippet. 

Provide more information and you may get help more quickly.

sbwalker:
@Sen-Gupta are you referring to the ability to have appsettings.json files for each environment? ie. appsettings.development.json?

sbwalker:
resolved by #2295

PR #2295, opened 2022-07-18
Fixed loading of alternative appsettings "appsettings.{env.EnvironmentName}.json"

chlupac:
Allows to have different configs in different enviroments.

### DIFF
--- a/Oqtane.Server/Startup.cs
+++ b/Oqtane.Server/Startup.cs
@@ -31,7 +31,8 @@ namespace Oqtane
         {
             var builder = new ConfigurationBuilder()
                 .SetBasePath(env.ContentRootPath)
-                .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true);
+                .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true)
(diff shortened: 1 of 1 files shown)


sbwalker:
I believe this may also resolve #2277. One question which I have is how this will affect the functionality in the framework which writes to the appsettings.json file... there is logic which adds configuration items to appsettings on upgrade as well as when a host user modifies settings in the System Info UI. I am concerned that the code change in this PR is not a complete solution and will break the scenarios I just described. Any thoughts?

chlupac:
I use this because I have a different connectionstring in development than in production. Also for set of logging levels. It's on host or developer responsibility how to use this possibility, appsettings.{env.EnvironmentName}.json is optional.
It overrides values from appsettings.json, overrided value change will be not reflected. But it overrides only presented values, not whole file.

sbwalker:
So basically you are saying that it is fine that the framework contains logic to modify appsettings.json - and if a developer optionally decides to use an environment-specific appsettings file, it would be up to them to modify it manually as necessary. This makes sense.

PR #2296, opened 2022-07-18
FileService fix

chlupac:
Return empty list when directory does not exists  (controller returns null).

### DIFF
--- a/Oqtane.Client/Services/FileService.cs
+++ b/Oqtane.Client/Services/FileService.cs
@@ -46,7 +46,7 @@ namespace Oqtane.Services
 
             var path = WebUtility.UrlEncode(folderPath);
 
-            List<File> files = await GetJsonAsync<List<File>>($"{Apiurl}/{siteId}/{path}");
+            List<File> files = await GetJsonAsync<List<File>>($"{Apiurl}/{siteId}/{path}") ?? new List<File>();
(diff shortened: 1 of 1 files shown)


sbwalker:
@chlupac Oqtane uses a specific approach in its API when retrieving collections to communicate status to the client. It returns an empty collection only under the scenario where the parameters passed to the method are valid but result in a filter operation which returns no records. In the event that the parameters passed to a method are invalid or the user does not have permission to access the resource, the method returns null. It is important to be able to differentiate between these 2 scenarios, and it is up to the client to handle them appropriately. Therefore, modifying the FileService to convert a null into an empty collection is not a valid solution as it changes the meaning of the API response and is not consistent with the rest of the API behavior.

chlupac:
Ok, second solution is ` return files?.OrderBy(item => item.Name).ToList();` to be safe and return null
Did you agree with this solution?

sbwalker:
Realistically it would probably be a better practice to implement the default ordering in the Controller method so that it avoids these types of exceptions in the client service layer and also leverages the server processing power. However the framework has been inconsistent in this area, so the solution you suggested above would also be fine.

chlupac:
Why is output ordered? I think it is not necessary.  Third soution is `return files;`

PR #2299, opened 2022-07-19
Fix - InstallationManager crash when package folders are missing

chlupac:


sbwalker:
@chlupac validating that the folders exist makes sense... as well as creating them if they are missing... so I will merge this PR. However, it is not clear to me how an installation could be created which are missing these folders... as these folders in the wwwroot contain assets which are fairly critical to Oqtane (ie. the Module and Theme folder contain the Template assets, and the Packages folder contains the database provider Nuget packages). Whether you installed Oqtane from the latest code in this repo or used an official release, it would always contain these folders.

chlupac:
It happens when I publish my application to new, clean place. I'ts not problem of installer.

sbwalker:
@chlupac you should use the release.cmd in the Oqtane.Package folder as it includes all of the instructions for publishing official release packages (which includes all of the folders)

PR #2305, opened 2022-07-22
Content root for Oqtane Private folders is configurable in appsettings.json 

chlupac:
with variable "ContentRoot", path can be absolute or relative.

If null or empty, defaults to WebHostEnvironment.ContentRootPath.

If relative, is relative to WebHostEnvironment.ContentRootPath

### DIFF
--- a/Oqtane.Server/Controllers/FileController.cs
+++ b/Oqtane.Server/Controllers/FileController.cs
@@ -28,16 +28,16 @@ namespace Oqtane.Controllers
     [Route(ControllerRoutes.ApiRoute)]
     public class FileController : Controller
     {
-        private readonly IWebHostEnvironment _environment;
+        private readonly IContentManager _contentManager;
         private readonly IFileRepository _files;
         private readonly IFolderRepository _folders;
         private readonly IUserPermissions _userPermissions;
         private readonly ILogManager _logger;
         private readonly Alias _alias;
 
-        public FileController(IWebHostEnvironment environment, IFileRepository files, IFolderRepository folders, IUserPermissions userPermissions, ILogManager logger, ITenantManager tenantManager)
+        public FileController(IContentManager contentManager, IFileRepository files, IFolderRepository folders, IUserPermissions userPermissions, ILogManager logger, ITenantManager tenantManager)
         {
-            _environment = environment;
+            _contentManager = contentManager;
             _files = files;
             _folders = folders;
             _userPermissions = userPermissions;
@@ -611,7 +611,7 @@ namespace Oqtane.Controllers
 
         private string GetFolderPath(string folder)
         {
-            return Utilities.PathCombine(_environment.ContentRootPath, folder);
+            return _contentManager.GetContentPath(folder);
         }
 
         private void CreateDirectory(string folderpath)
--- a/Oqtane.Server/Controllers/FolderController.cs
+++ b/Oqtane.Server/Controllers/FolderController.cs
@@ -18,15 +18,15 @@ namespace Oqtane.Controllers
     [Route(ControllerRoutes.ApiRoute)]
     public class FolderController : Controller
     {
-        private readonly IWebHostEnvironment _environment;
+        private readonly IContentManager _contentManager;
         private readonly IFolderRepository _folders;
         private readonly IUserPermissions _userPermissions;
         private readonly ILogManager _logger;
         private readonly Alias _alias;
 
-        public FolderController(IWebHostEnvironment environment, IFolderRepository folders, IUserPermissions userPermissions, ILogManager logger, ITenantManager tenantManager)
+        public FolderController(IContentManager contentManager, IFolderRepository folders, IUserPermissions userPermissions, ILogManager logger, ITenantManager tenantManager)
         {
-            _environment = environment;
+            _contentManager = contentManager;
             _folders = folders;
             _userPermissions = userPermissions;
             _logger = logger;
@@ -242,7 +242,7 @@ namespace Oqtane.Controllers
 
         private string GetFolderPath(Folder folder)
         {
-            return Utilities.PathCombine(_environment.ContentRootPath, "Content", "Tenants", _alias.TenantId.ToString(), "Sites", folder.SiteId.ToString(), folder.Path);
+            return _contentManager.GetContentPath("Content", "Tenants", _alias.TenantId.ToString(), "Sites", folder.SiteId.ToString(), folder.Path);
         }
     }
 }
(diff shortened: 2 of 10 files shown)


sbwalker:
Currently the framework supports 2 types of folders - Private and Public. The difference between these 2 folder types is confusing to most people as they are both managed by the same Permissions options. However Permissions only manage access to these assets within the Oqtane application user interface itself - they are not able to control access to the physical assets that exist in the underlying file system. 

Private folders are subfolders beneath the Content folder in the application root - which means that there is no way to access the files in this location unless they are explicitly served by an Oqtane API (ie. they are secure). This is because the native .NET StaticFileHandler does not provide access to files outside of the wwwroot.

Public folders are subfolders beneath the Content folder in the wwwroot (ie. wwwroot/Content) - which means they can be accessed by any web client who knows the specific Url (ie. they are not secure). 

This PR appears to be to allow a developer to override the wwwroot location which obviously changes the physical location of where the Public folder assets will be stored. The main question which comes to mind is whether the StaticFileHandler will still be able to serve these assets regardless of the location which the developer specifies - as if it cannot, then it breaks the whole concept of Public folders.

sbwalker:
It is not clear to me what problem you are trying to solve with these changes... but if you are interested in being able to store files in other physical locations then perhaps a better approach would be to make the folder type concept extensible?

chlupac:
This PR allows to move PRIVATE root to independent location, not to be mixed with application files. I do not override wwwroot, I allow to move location, where private files are stored. 

My goal is allow to move user files OUTSIDE of application folder structure.  
It can be done in two steps:
1. Move  PRIVATE root - this PR, which replaces _environment.ContentRootPath with _contentManager.ContentRootPath
2. Move Public root - can be archieved in IIS with virtual directory and web.config file, it's under investigation

```c#   
FolderRepository.cs:
 public string GetFolderPath(Folder folder)
        {
            string path = "";
            switch (folder.Type)
            {
                case FolderTypes.Private:
                    path = _contentManager.GetContentPath("Content", "Tenants", _tenants.GetTenant().TenantId.ToString(), "Sites", folder.SiteId.ToString(), folder.Path);
                    break;
                case FolderTypes.Public:
                    path = Utilities.PathCombine(_environment.WebRootPath, "Content", "Tenants", _tenants.GetTenant().TenantId.ToString(), "Sites", folder.SiteId.ToString(), folder.Path);
                    break;
            }
            return path;
        } 


// and here is path resolution, from ContentPathManager.cs:

    private string GetContentRoot()
    {
        var root = _config.GetValue(SettingKeys.ContentRootKey, String.Empty);
        if (root.IsNullOrEmpty()) return _environment.ContentRootPath; // if no key do nothing, same as current version
        if (Path.IsPathRooted(root)) return root;  // if absolute path, use it                     
        return Path.GetFullPath(Path.Combine(_environment.ContentRootPath, root!)); // if relative path, count it in relation to current path ... root == "../" moves folder one level down in structure
    }
`

sbwalker:
The request to be able to store files in different physical locations has been raised previously. And in fact the framework was developed with an abstraction for files and folders for precisely this reason. 

The abstraction initially included the ability to define virtual hierarchical folders, support for permissions, UI components, etc... And for simplicity it included 2 variations of local storage which were tightly coupled to the framework. However the goal was always to include the ability to fully abstract the physical storage of files.

The plan was to replace the Folder Type concept with a true plug-in model where developers could provide alternate implementations which conform to an core interface. This would provide the flexibility to allow files to be stored on disk, in a database, or in the cloud. 

So although I understand the intent of the changes in this PR, they do not align with the longer term goals for the framework in this area. Rather than modifying existing framework behavior with configuration options it would be preferable to use a more well defined abstraction. 

And specifically related to your use case I would expect that a LocalFolderProvider plug-in would have ability to configure the root path - stored in FolderSettings in the database rather than appsettings.json.

chlupac:
I understand that this PR is just a light abstraction of the file location. I think that the concept of "Public" and "Private" files was created in a similar way and you don't want to expand this solution any further.. I believe that this implementation is not optimal. Likewise, it is complicated for content creators to access files via fileId.

I suggest

1. implement a generic IOqtaneFileProvider : IFileProvider {https://docs.microsoft.com/en-us/aspnet/core/fundamentals/file-providers?view=aspnetcore-6.0} which would map the /content or /files web folder, to the content/tenants/{TenantId}/Site/{SiteID} folder and would allow access to files with permissions respected. This will make the Public/Private folder flag obsolete, as all files will be accessed via their path/name.
2. extend the IOqtane provider with the StoreFile function, which will enable writing to the given provider, if it is not Read-only
3. Expand the Folder table with the Provider column. The provider defined here will be used to work with the given folder.

questions:
1. Inheritance. Will child folders inherit provider? Or the provider will be set to Inherit as default.
2. Another possibility is that the root directory is mapped for each provider? (//provider/path/path/file)
3. Is it necessary for all folders and files to be mapped to the database? I think not, the provider used for the given folder will be able to read information from the repository.

sbwalker:
The high level API which module developers use to interact with Folders and Files (ie. FolderService, FileService) need to remain the same (avoid breaking changes and avoid duplicate APIs which can cause confusion). The new abstraction would be purely on the backend - allowing folders to be mapped to different physical locations through FolderType providers.

sbwalker:
Yes all files and folders need to be mapped through the database as it provides the higher level abstraction for permissions and additional metadata.

sbwalker:
Please note that the concept of a folder provider needs to support more than just arbitrary locations on disk - it also needs to be able to accommodate database blob storage and storage in cloud providers such as S3, Azure blobs, etc...

sbwalker:
If I created a diagram it will probably be more clear, as I have a very specific vision for this area - it is conceptually similar to the Folder Provider concept which exists in DNN, which was extremely successful in allowing developers to extend the framework capabilities.

chlupac:
> Please note that the concept of a folder provider needs to support more than just arbitrary locations on disk - it also needs to be able to accommodate database blob storage and storage in cloud providers such as S3, Azure blobs, etc... so Microsoft's fileprovider is not a solution.

IFileProvider can be implemented in many ways ...
webdav implementation
https://github.com/patagonaa/FileProviders

database implementation
https://www.mikesdotnetting.com/article/301/loading-asp-net-core-mvc-views-from-a-database-or-other-location

azure blob
https://github.com/endintiers/ASPNetCore-FileProvider-WMBlobContainerProvider

sbwalker:
I will take a closer look at IFileProvider... I was not aware it had these capabilities. The main challenge with most .NET services is that they do not naturally support multi-tenancy and they store their configuration in appsettings which is not optimal in a framework like Oqtane for a variety of reasons. So a wrapper would likely be required as you suggested.

chlupac:
Some try-fail .... You was right, IFileProvider is not applicable in my way, because is implemented as singleton service. I'm examining DNN FolderProviders.

PR #2322, opened 2022-07-28
Updated Bootstrap to 5.2

leigh-pointer:
Replaced Bootstrap cloudflare versions and Integrity keys to match 5.2.0

### DIFF
--- a/Oqtane.Client/Installer/Installer.razor
+++ b/Oqtane.Client/Installer/Installer.razor
@@ -186,8 +186,8 @@
 		if (firstRender)
 		{
 			var interop = new Interop(JSRuntime);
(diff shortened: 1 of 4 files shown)


sbwalker:
@leigh-pointer each time we upgrade to a newer version of Bootstrap it impacts the appearance of the framework (which is actually surprising to me as I would expect backward compatibility when upgrading to newer versions within the same major release). Some examples:

1. The Module Action dropdown indicator lost some margin spacing so it appears to be much closer to the module title now:
 
image

2. The Admin Dashboard icons are now missing some margin spacing - making the rows appear squashed together:

image

leigh-pointer:
I will action

PR #2327, opened 2022-08-02
Added Progress Indicator 

leigh-pointer:
When deleting large blocks of Pages, Modules or Notifications there was currently no visual feedback so added the ModuleInstance.ShowProgressIndicator() and ModuleInstance.HideProgressIndicator() calls to these processes.


sbwalker:
@leigh-pointer I will merge this but I will note that you do not need to include the ModuleInstance reference when calling the ShowProgressIndicator of HideProgressIndicator method - the methods are exposed through ModuleBase.

PR #2431, opened 2022-09-24
Unattended installation fix

chlupac:
When connection string is specified, result.success == true prevents automatic install to run.
result.Success = true should be set only when everything is ok.


sbwalker:
Thank you @chlupac... the change appears to be safe but since there are so many different scenarios handled by this code I want to verify them all (ie. new install, unattended install, new site creation, framework upgrade ).

Issue #2584, opened 2023-02-07
SQLite database init fails with 'An Error Occurred Creating Site. An error occurred while saving the entity changes. See the inner exception for details.'


markdav-is:
When pulling master for v3.3.1 and running the startup screen in the SQLite configuration it stops after a few moments with the error "An Error Occurred Creating Site. An error occurred while saving the entity changes. See the inner exception for details."  no additional details are captured in the logs.

![image](https://user-images.githubusercontent.com/311063/217147394-00c3cd7d-c004-4923-8642-53e9010f5206.png)

sbwalker:
Thank you for reporting this issue... it turns out that this is a bit tricky. In version 3.2.2 migrations were introduced to remove the DeletedBy, DeletedOn, and IsDeleted columns from the Folder and File tables (as these columns were never used). However SQLite does not support the ability to drop columns - so the migration executes fine but does not remove the columns from the tables in the database. And then the system tries to create some folder records during install and it fails because the IsDeleted column has a NOT NULL constraint. I am going to have to investigate if there is a way to remove the constraint so that this works on SQLite. Multi-database support is a real challenge based on the fact that not all database engines support the same basic operations.

sbwalker:
There is no ability to alter a column in SQLite either... so no ability to change the constraint. Which basically only allows for 2 options:

1. Add the IsDeleted properties back to the Folder and File entities and identify them as deprecated. The IsDeleted property will still need to be explicitly set to avoid the SQLite error during insert - even though the property is meaningless to the rest of the framework.
2. Remove the deprecated properties from the EntityBuilders. This will ensure they are never added on a new install. However this will not resolve the issue for users that are already running on SQLite and want to upgrade to a later version of Oqtane.

Option 1 is the better approach so that Oqtane can meet its upgradeability philosophy. In the future this essentially means that properties should not be physically removed from database entities if they are not nullable.

markdav-is:
thanks for addressing this.

PR #2605, opened 2023-02-19
add AddByteColumn to add tinyint to the database table.

Behnam-Emamian:
Added AddByteColumn method to BaseEntityBuilder class to generate tinyint field in the database table.

### DIFF
--- a/Oqtane.Server/Migrations/EntityBuilders/BaseEntityBuilder.cs
+++ b/Oqtane.Server/Migrations/EntityBuilders/BaseEntityBuilder.cs
@@ -122,6 +122,26 @@ namespace Oqtane.Migrations.EntityBuilders
             return table.Column<DateTimeOffset>(name: RewriteName(name), nullable: nullable, defaultValue: defaultValue);
         }
 
+        public void AddByteColumn(string name, bool nullable = false)
+        {
+            _migrationBuilder.AddColumn<byte>(RewriteName(name), RewriteName(EntityTableName), nullable: nullable, schema: Schema);
+        }
+
+        public void AddByteColumn(string name, bool nullable, int defaultValue)
+        {
+            _migrationBuilder.AddColumn<byte>(RewriteName(name), RewriteName(EntityTableName), nullable: nullable, defaultValue: defaultValue, schema: Schema);
+        }
+
+        protected OperationBuilder<AddColumnOperation> AddByteColumn(ColumnsBuilder table, string name, bool nullable = false)
+        {
+            return table.Column<byte>(name: RewriteName(name), nullable: nullable);
+        }
+
+        protected OperationBuilder<AddColumnOperation> AddByteColumn(ColumnsBuilder table, string name, bool nullable, int defaultValue)
+        {
+            return table.Column<byte>(name: RewriteName(name), nullable: nullable, defaultValue: defaultValue);
+        }
+
         public void AddIntegerColumn(string name, bool nullable = false)
         {
             _migrationBuilder.AddColumn<int>(RewriteName(name), RewriteName(EntityTableName), nullable: nullable, schema: Schema);


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=2605) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=2605) before we can accept your contribution.<br/><br/>:x: Behnam-Emamian [sign now](https://cla.dotnetfoundation.org/oqtane/oqtane.framework?pullRequest=2605)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/oqtane/oqtane.framework?pullRequest=2605) it.</sub>

sbwalker:
Oqtane supports 4 database engines - SQL Server, SQLite, mySQL, PostgreSQL. The BaseEntityBuilder only included data types which were supported by all of these databases. Do you know if tinyint is supported by all of these databases?

Behnam-Emamian:
EF Core SQL Server Provider (Microsoft.EntityFrameworkCore.SqlServer) is using Microsoft.Data.SqlClient to connect to sql server database and supports byte as tinyint field.

EF Core Sqlite Provider (Microsoft.EntityFrameworkCore.Sqlite) is using Microsoft.Data.Sqlite to connect to sqlite database and supports byte as INTEGER field.
https://learn.microsoft.com/en-us/dotnet/standard/data/sqlite/types

EF Core MySQL Provider (MySql.EntityFrameworkCore) is using MySQL Connector/NET to connect to MySql database and supports byte as tinyint unsigned

EF Core PostgreSQL Provider (Npgsql.EntityFrameworkCore.PostgreSQL) is using Npgsql to connect to PostgreSql database and supports byte as smallint
https://www.npgsql.org/doc/types/basic.html

Behnam-Emamian:
I have tested sql server and it is ok, I can see the tinyint field in the table.

I am testing sqlite but I cannot install oqtane 3.3.1 on sqlite, it shows an error and stops in installation wizard.

I will test MySql and PostgreSQL after finding the sqlite bug.

Behnam-Emamian:
It cannot create site when the db is sqlite

image

Behnam-Emamian:
Ok, found the bug

I can see in the folder migration, logical delete fields has been added to the table:
image

but the entity does not have them:
image

then we receive a null exception error in inserting a folder record in the sqlite folder table.

@sbwalker may I know what your advice is, 
Should I add the IsDeleted, DeletedOn and DeletedBy to the entity?
OR
Do I need to remove AddDeletableColumns from migration?

and it is strange how it is working for SQL Server!!! wow, Folder table in sql server does not have IDeletable columns!?!?!?
image

Behnam-Emamian:
I have found this migration RemoveFolderFileDeletableColumns which I can see has been run in sql server and sqlite in __EFMigrationsHistory

but the IDeletable columns are still there in sqlite folder table

Behnam-Emamian:
nice, DropColumn is not working in sqlite :-D
https://github.com/dotnet/efcore/issues/329

Solution:
https://stackoverflow.com/questions/41902905/ef-core-dropping-a-column-workaround-sqlite

I can fix it if you are ok!

sbwalker:
@Behnam-Emamian the issue with SQLite was already reported and fixed in https://github.com/oqtane/oqtane.framework/issues/2584

Behnam-Emamian:
> @Behnam-Emamian the issue with SQLite was already reported and fixed in #2584

It can be the option 3, like copy from old table to a temporary table without those columns that we need to delete and then delete the old table and renew the temporary table to old table.

sbwalker:
I previously investigated the StackOverflow link you provided but the solution is rather ugly and difficult to integrate. The goal in the Oqtane database abstraction is to keep the core API simple and generic and try to avoid conditional logic as much as possible. Any differences in database engine implementations should be in the providers - which is why the DropColumn method in the SQLite provider is overridden (as well as a few others). In order to include logic similar to the StackOverflow article in the SQLite provider, the DropColumn method would need a LOT more context (since it needs to recreate the table, etc...) but this context would only be applicable to SQLite. As a result I opted for a simpler approach.

sbwalker:
In case my explanation is not clear, a developer using Oqtane should be able to use simple/generic methods when interacting with the database. Specifically in their own code they should be able to specify DropColumn("ColumnName") and it should "work" on all supported databases. They should not not have to write DropColumn("ColumnName", "Create New Table Logic", "Migrate Data Logic", etc...) as the extra context is only applicable to a specific database (SQLite).

sbwalker:
Based on your research it appears that all of the databases support tinyint (or at least map it to some other internal data type). So I can merge this PR.

Behnam-Emamian:
> In case my explanation is not clear, a developer using Oqtane should be able to use simple/generic methods when interacting with the database. Specifically in their own code they should be able to specify DropColumn("ColumnName") and it should "work" on all supported databases. They should not not have to write DropColumn("ColumnName", "Create New Table Logic", "Migrate Data Logic", etc...) as the extra context parameters are only applicable to a specific database (SQLite).

I am completly agree but this is the sqlite limitation and if user really need to delete column and save the current data, there is no simple solution like DropColumn("ColumnName") unfortunately. 

I have checked the Microsoft.Data.Sqlite solution for this limitation, it is "rebuilding the table" which I believe it is not a good solution as the data will be lost.

https://learn.microsoft.com/en-us/ef/core/providers/sqlite/limitations?toc=%2Fdotnet%2Fnavigate%2Fdata-access%2Ftoc.json&bc=%2Fdotnet%2Fbreadcrumb%2Ftoc.json

sbwalker:
> I am completly agree but this is the sqlite limitation and if user really need to delete column and save the current data, there is no simple solution like DropColumn("ColumnName") unfortunately.

Correct - there is no simple solution at the database level due to differences between database engines - so the simple solution is to orphan the column in the SQLite database (this approach is actually very common in the NoSQL database community).

PR #2609, opened 2023-02-24
extends watching *.dll files

Behnam-Emamian:
extends watching group to include *.dll files and exclude the ones cause an infinite loop.

@sbwalker I need your advice about where I should put the documentation about how to use it.

Hey @vzdesic, could you please help me to finish this as you initiated the issue and provided solution for this.


vzdesic:
@Behnam-Emamian Sure, I will help if needed. Is something needs to be clarified related to mentioned steps?

Behnam-Emamian:
@vzdesic 

You prepaid a good documentation for debugging external module, I believe we need to put it somewhere in https://docs.oqtane.org/

Issue #2625, opened 2023-03-04
When Clearing modules from the Recycle Bin Records are Orphaned


leigh-pointer:
When the Modules are cleared from the Recycled bin only the PageModule record is deleted thus leaving orphaned records in tables;
Module, Permissions, Settings

sbwalker:
The expected behavior is that if you delete a module from the recycle bin, the framework removes the PageModule instance. It then checks if there are any other PageModule instances for that ModuleId in the Site. If none are found then it deletes the Module record as well.

leigh-pointer:
@sbwalker I will close #2625 and recreate as described.

PR #2626, opened 2023-03-04
Fix Clearing modules from the Recycle Bin Records are Orphaned #2625

leigh-pointer:
There was  bug in the Recycle Bin that stopped the module from being deleted.
That is fixed. Also moved the deletion logic to the ModuleController.  Now when a module is deleted the related records are also deleted ; PageModules, Permissions, Settings and finally the Module.


leigh-pointer:
Closing this PR adding functionality to Recycle Bin

PR #2642, opened 2023-03-06
Fix Clearing modules from the Recycle Bin

leigh-pointer:
Modules and all related records now correctly deleted from the database.

### DIFF
--- a/Oqtane.Client/Modules/Admin/RecycleBin/Index.razor
+++ b/Oqtane.Client/Modules/Admin/RecycleBin/Index.razor
@@ -185,7 +185,7 @@ else
             await PageModuleService.DeletePageModuleAsync(module.PageModuleId);
 
             // check if there are any remaining module instances in the site
-            if (!_modules.Exists(item => item.ModuleId == module.ModuleId))
+            if (_modules.Exists(item => item.ModuleId == module.ModuleId))
             {
                 await ModuleService.DeleteModuleAsync(module.ModuleId);
             }
@@ -211,7 +211,7 @@ else
                 await PageModuleService.DeletePageModuleAsync(module.PageModuleId);
 
                 // check if there are any remaining module instances in the site
-                if (!_modules.Exists(item => item.ModuleId == module.ModuleId))
+                if (_modules.Exists(item => item.ModuleId == module.ModuleId))
                 {
                     await ModuleService.DeleteModuleAsync(module.ModuleId);
                 }
--- a/Oqtane.Client/Modules/Admin/RecycleBin/Index.razor
+++ b/Oqtane.Client/Modules/Admin/RecycleBin/Index.razor
@@ -76,8 +76,8 @@ else
 }
 
 @code {
-	private List<Page> _pages;
-	private List<Module> _modules;
+    private List<Page> _pages;
+    private List<Module> _modules;
 
     public override SecurityAccessLevel SecurityAccessLevel => SecurityAccessLevel.Admin;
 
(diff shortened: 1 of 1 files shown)


leigh-pointer:
Running these queries verify the deletion

```
--Permissions Without Modules
SELECT        *
FROM            Permission LEFT OUTER JOIN
                         Module ON Module.ModuleId = Permission.EntityId
WHERE        (Permission.EntityName = N'Module' AND Module.ModuleId Is NULL)

--Permissions Without PageModule
SELECT        *
FROM            Permission LEFT OUTER JOIN
                         PageModule ON PageModule.PageId = Permission.EntityId
WHERE        (Permission.EntityName = N'Page' AND PageModule.PageId Is NULL)

--Settings Without Modules
SELECT        *
FROM            Module RIGHT OUTER JOIN
                         Setting ON Module.ModuleId = Setting.EntityId
WHERE        (Module.ModuleId IS NULL) AND (Setting.EntityName = N'Module')
```

These will delete any orphaned records

```
---DELETE
DELETE FROM Permission
FROM            Permission LEFT OUTER JOIN
                         Module ON Module.ModuleId = Permission.EntityId
WHERE        (Permission.EntityName = N'Module') AND (Module.ModuleId IS NULL)

DELETE FROM Permission
FROM            Permission LEFT OUTER JOIN
                         PageModule ON PageModule.PageId = Permission.EntityId
WHERE        (Permission.EntityName = N'Page' AND PageModule.PageId Is NULL)

DELETE FROM Setting
FROM            Module RIGHT OUTER JOIN
                         Setting ON Module.ModuleId = Setting.EntityId
WHERE        (Module.ModuleId IS NULL) AND (Setting.EntityName = N'Module')
```

sbwalker:
@leigh-pointer this logic still appears to be incorrect. It is checking if the module exists in the list of modules for the site, and if so, it is deleting the module. The logic should be opposite - only if the module does NOT exist in the site then it should be deleted.

leigh-pointer:
@sbwalker on Task Load() the modules are loaded these are all modules with no distinction between deleted of not. So the module needs to be removed from the collection.  Currently using the **!**_modules.Exists leaves the module and settings in the database.
As a sanity check I have just executed the procedure with the current codebase and the DeleteModule leaves orphaned records.

sbwalker:
Allow me to provide some more clarity... Oqtane has the concept of Module and PageModule. So if you create an instance of the Html/Text module on Page1 of your site, there is both a Page and PageModule record created. Then if you navigate to Page2, you can use the "Add Existing Module" feature in the Control Panel to add an instance of the same Html/Text module from Page 1 onto Page 2 - and in this case only a PageModule record is created. So now you decide that you want to delete the Html/Text module on Page 1. What should happen is that the PageModule record for Page 1 is deleted - but the Module record should NOT be deleted (because it is still associated to the PageModule record on Page 2). Later, you might decide to delete the Html/Text module instance on Page 2. So the PageModule record should be deleted. In this case the system should also be able to determine that there are no more module instances (PageModules) for that ModuleId which exist - and therefore it should also delete the permissions and settings.

leigh-pointer:
Ok got it, will look at that Case now.

leigh-pointer:
Currently there is now way in the API to collect all PageModules by ModuleId as this would be the only way for this check to be done. Or am I in the wrong bus?

sbwalker:
ModuleService.GetModulesAsync(siteId) gets all PageModule instances in the site. So to check if the ModuleId still exists in the site you simply need to check:

```
            await PageModuleService.DeletePageModuleAsync(module.PageModuleId);

            // check if there are any remaining module instances in the site
	    if (!_modules.Exists(item => item.ModuleId == module.ModuleId && item.PageModuleId != module.PageModuleId))
```
the "&& item.PageModuleId != module.PageModuleId" is the missing logic which is needed to ignore the PageModuleId you just deleted

PR #2656, opened 2023-03-09
Updated all Pager controls to handle the OnPageChange event

leigh-pointer:


sbwalker:
@leigh-pointer the logic you added is valid for tracking the current page number within the current component. However unless it is combined with explicitly persisting the state across component renderings, it is not useful to the end user.

For example... if I am in Page Management (Index.razor) and I go to Page 2 and then click the Edit button for a specific item, the Edit Page component (Edit.razor) will be displayed. Then if I then click Cancel, the main Page Management component (Index.razor) will again be displayed but I will be back on page 1. This is because the state of the "page" variable is local to the Index.razor component and was lost in the transition from Index.razor to Edit.razor (and again).

In order to use the OnPageChange to benefit the user, you need to explicitly pass the page value to other components being loaded. A good example of this is the Event Log which utilizes Url Parameters to pass the state of various filters (including page number) from the Index.razor component to the Details.razor component and back again.

Please let me know if this PR provides something that I do not understand.

leigh-pointer:
@sbwalker probably was over zealous with this PR.  There are certainly a few valid changes.  I will cherry pick through and update this PR.

sbwalker:
I already merged #2649 but after closer inspection I am trying to understand why saving the page number state for the 2 pagers was necessary? Was the component somehow losing state and resetting the pagers to the first page?

leigh-pointer:
@sbwalker 

> Was the component somehow losing state and resetting the pagers to the first page?
Yes, both Pages and Modules.

Other admin pages that have item Delete functionality will need to have the CurrentPage parameter set, I believe.

PR #2685, opened 2023-03-15
Cosmetic change - Rider recognize 'required' keyword in middle as syntax error

chlupac:


sbwalker:
@chlupac I appreciate this PR but I cannot merge it in its current form, as it contains a wide variety of modifications which are unrelated to the title "Rider recognize 'required' keyword in middle as syntax error". If the modifications could be grouped according to the problem they solve and submitted separately with a description of the actual problem it would allow the changes to be reviewed more easily. 

The general process I follow for problems that are reported is that I need to be able to reproduce them myself locally in order to validate the issue - and then I can review the PR to determine if it effectively and safely resolves the problem.

PR #2691, opened 2023-03-20
Code Cleanups

Behnam-Emamian:


Oqtane.Server/Controllers/PageController.cs
```diff
@@ -345,7 +345,7 @@ private List<Permission> GetPermissionsDifferences(List<Permission> permissions1
         [Authorize(Roles = RoleNames.Registered)]
         public void Put(int siteid, int pageid, int? parentid)
         {
-            if (siteid == _alias.SiteId && siteid == _alias.SiteId && _pages.GetPage(pageid, false) != null && _userPermissions.IsAuthorized(User, siteid, EntityNames.Page, pageid, PermissionNames.Edit))
```
Behnam-Emamian, line 348: I believe second condition is not necessarily.

Oqtane.Server/Extensions/OqtaneSiteAuthenticationBuilderExtensions.cs
```diff
@@ -358,13 +358,13 @@ private static async Task<ClaimsIdentity> ValidateUser(string email, string id,
                                         else
                                         {
                                             identity.Label = ExternalLoginStatus.UserNotCreated;
-                                            _logger.Log(user.SiteId, LogLevel.Error, "ExternalLogin", Enums.LogFunction.Create, "Unable To Add User {Email}", email);
```
Behnam-Emamian, line 361: user is null here.

Oqtane.Server/Infrastructure/Jobs/NotificationJob.cs
```diff
@@ -138,7 +138,7 @@ public override string ExecuteJob(IServiceProvider provider)
                             try
                             {
                                 client.Send(mailMessage);
-                                sent = sent++;
```
Behnam-Emamian, line 141: it should be a mistake I believe.

sbwalker:
@Behnam-Emamian I appreciate the PR however I am having a difficult time validating these changes. This is because they are related to many different areas of the application... and the fact that there is no description of the issue they address or any steps to reproduce makes it very difficult to determine if they are valid. In general, I am not in favor of making changes to code because it "does not look correct" - I have been burned by this approach too many times in the past, only to find out that I did not understand the run-time characteristics of the logic and introduced a bug that did not exist previously. Are you able to provide more context for these changes?

Behnam-Emamian:
I have setup the SonarQube for my project and it found these bugs which I believe they are positive alarm. 

I have added a short description to each of them, 3 of them are accessing to a null reference and 2 of them are related to mistake code.

PR #2697, opened 2023-03-24
Modified Registration to display the Password requirments

leigh-pointer:
Generating a password can generate a lot of errors and leading to frustration when trying to Register on an Oqtane site.  This is manly due to the visitor not knowing the Password Requirements.  This has been addressed in this PR.

The Password requirements are now displayed to the visitor.

image

### DIFF
--- a/Oqtane.Client/Modules/Admin/Register/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Register/Index.razor
@@ -1,6 +1,5 @@
 @namespace Oqtane.Modules.Admin.Register
 @inherits ModuleBase
-@using System.Resources;
 @inject NavigationManager NavigationManager
 @inject IUserService UserService
 @inject IStringLocalizer<Index> Localizer
--- a/Oqtane.Client/Modules/Admin/Register/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Register/Index.razor
@@ -16,7 +16,7 @@
             <ModuleMessage Message="@Localizer["Info.Registration.Exists"]" Type="MessageType.Info" />
         </Authorized>
         <NotAuthorized>
-            <ModuleMessage Message="@Localizer["Info.Registration.InvalidEmail"]" Type="MessageType.Info" />
+            <ModuleMessage Message="@_passwordconstruction" Type="MessageType.Info" />
             <form @ref="form" class="@(validated ? "was-validated" : "needs-validation")" novalidate>
                 <div class="container">
                     <div class="row mb-1 align-items-center">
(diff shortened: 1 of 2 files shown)


sbwalker:
@leigh-pointer I agree that it is helpful to provide the user with more information related to password complexity requirements. My preference would be for this information to be included in the HelpText of the Password label - however I am not sure of the technical feasibility of including dynamic text in this property. As an alternative I am wondering if the password information could be included in the same ModuleMessage which includes the "Please Note That Registrations Requires..." text?

leigh-pointer:
Im on it! Wait for new or updated PR

leigh-pointer:
@sbwalker is this better?

image

thabaum:
As discussed Tool-tips and Validation error messages generally fits all registration form requirement which it redundant?

I think improving some nicer modal popups for the entire framework help tool-tips would something to consider a focus if this is not adequate.  In areas like this form to help explain what is needed to register keeping a completely uniform look and feel with direct messages maybe there are some things that can be enhanced adding some dynamic engineering to Oqtane.

Usually this stuff is pretty straight forward now days.  And you might even create a step-by-step process to enhance the application for on-boarding directing users to entirely different page of components.

If username is email address then the instructions can change.

Username requirements need to be in that message as well.  This will make it bigger on loading for small screen devices.

Can we enable/disable this feature in the Registration form as well?  Disable it for screens smaller than certain size?  Maybe a BIG question mark at the top a user can over or click for a modal?  So many possibilities ;) 

Screen space is what I take into consideration.  I think the form should not be any wider than few characters longer than the longest input you can enter to fix how wide the entire form is to keep it looking.  I would call that a plus on all edit forms.  I dont care for ultra wide forms for a password I think it looks horrible.   The longest thing is the email... forms fields i think should not be wider than this.

To be more direct with your message if you did create it maybe something like:

Requirements:
Email Verification Process
Username must be:  [ option 1 ], [ option 2 ],etc... and [ last option if more than one ]
Password must be: [ option 1 ], [ option 2 ],etc... and [ last option if more than one ]

If a an email verification process next page "We sent you a code to verify your [ email or phone number ] please enter it below:  Resend Code (please wait [ Wait Time ] minutes."

Again if options are enabled otherwise keep it clean yet function with industry standard best practices and let developers go from there.  Options are cool to display how you can make certain features enhance the core features for ideas towards the framework you maybe using Oqtane for as a base.

It would be nice to see a class of modules supported and sold on Oqtane network that is check (pays to be checked) to have all code comply with Oqtane standards.  Then we would know if we are dealing with a vendor that actually keeping the project or just throwing it out there as something to buy and use.

Anyhow...

With form validation messages shown when typed wrong along with tool tips makes it hard to want to add more so I would like to see a switch added for this as I can see how people would maybe like this for their users.   Help Modals improved always a great idea if uniform across the framework.

Is this something community suggested to fix others issues getting users registered I am not aware of?

thabaum:
As another idea, could we maybe put a way to put tool-tip modal popups into the title with a larger question mark to click that pops up all instructions of a module maybe it can even hold FAQs, make this available for any module to be placed in title or anywhere you wish for it to be.  Module Modal Popup feature you can set the information, instructions or FAQ's without taking up another line of the screen?

PR #2717, opened 2023-04-04
Added Sorting to User- could be extended to other Pagers.

leigh-pointer:
Added the ability to sort the Pager data by clicking the Pager Heder Column.

Oqtane-PagerSorting

### DIFF
--- a/Oqtane.Client/Modules/Admin/RecycleBin/Index.razor
+++ b/Oqtane.Client/Modules/Admin/RecycleBin/Index.razor
@@ -26,9 +26,9 @@ else
 					<Header>
 						<th style="width: 1px;">&nbsp;</th>
 						<th style="width: 1px;">&nbsp;</th>
-						<th>@SharedLocalizer["Name"]</th>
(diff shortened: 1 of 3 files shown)


leigh-pointer:
Updated Recycle Bin with column Sorting

sbwalker:
I have not merged this PR for a number of reasons. From a business perspective, sorting can be a useful feature for users when dealing with certain types of data (ie. amounts, etc...), however as a general feature for the types of information managed in the Admin UI I am not sure it is very useful. From a technical perspective I have concerns on how it would scale with higher volume and I do not like how much duplicate/custom code needs to be added (and maintained) in each razor component to support it. Overall, the main question is related to the amount of value it adds to the product.

leigh-pointer:
@sbwalker I have now been running sort on one of my modules with over 25 column and 2500 records.  The performance is unnoticeable.  However I would want to update the Oqtane PR.

leigh-pointer:
I will need to do some refactoring but really happy to do it.

leigh-pointer:
replaced with #3016

Issue #2721, opened 2023-04-04
Edit site is not working when site is not working


tvatavuk:
Issue is related to `Admin / Site Management`.

It is visible when sub-site is not working as demonstrated in #2720. Where main site is working but sub-site `s1` is not working.

## Steps

1. On main site (https://oqtane-poc212.oqtane.me/) open `Admin / Site Management`.

![image](https://user-images.githubusercontent.com/2724850/229807447-2ba01074-a2ae-4eba-a404-24f0e263320d.png)

2. Click on **Edit** for sub-site `s1`.

![image](https://user-images.githubusercontent.com/2724850/229807499-36d7583e-c26d-4e2f-844f-9b11a6d5ede4.png)

Edit site is not working, because site edit is executed with in sub-site  (https://oqtane-poc212.oqtane.me/s1) that is not working because of #2720.

It will be great if we can edit `s1` site options, on main site (https://oqtane-poc212.oqtane.me/).

sbwalker:
The expected behavior of the Edit button in Site Management is that it navigates you to the other site (where you can edit if you have the appropriate rights). The security architecture prevents you from accessing APIs or performing any type of management for a site with a different alias. This is by design.

Issue #2720, opened 2023-04-04
Sub-site breaks after restart application


tvatavuk:
New created sub-site breaks after restart application. It is not working anymore. This issue is show-stopper for sub-site functionality in Oqtane v3.4.2.

## Steps:
1. Clean install release `Oqtane.Framework.3.4.2.Install.zip` on **IIS** (https://oqtane-poc212.oqtane.me/) with local **MS SQL** server.

![image](https://user-images.githubusercontent.com/2724850/229800872-d0ae94b0-d00d-4e2d-b071-4c73a14da7e8.png)

2. In  `Admin / Site Management` create new sub-site `s1` (https://oqtane-poc212.oqtane.me/s1)

![image](https://user-images.githubusercontent.com/2724850/229801708-5e828a97-da44-4f83-acdc-1f09daea8ce4.png)

Sub-site is created and working as expected.

![image](https://user-images.githubusercontent.com/2724850/229801816-ba8f9baa-a2b9-4754-9cfe-9f415382b2c3.png)

3. In `Admin / System Info` **Restart Application**.

![image](https://user-images.githubusercontent.com/2724850/229802386-7ceb4685-2e03-4444-8b51-0cb57f509d49.png)

**Sub-site breaks after restart.**

Administration is not working.

![image](https://user-images.githubusercontent.com/2724850/229802728-ebb153c0-3872-4fe4-9a30-b4de681d16e9.png)

Sub-site S1 home page is not working.

![image](https://user-images.githubusercontent.com/2724850/229802779-2ea77863-c011-48a1-8ba9-66e35a74a812.png)

Main site is still working.

![image](https://user-images.githubusercontent.com/2724850/229802981-77dd7199-3743-488b-9e8d-4abe10690d2d.png)

### Test again

So we can create new sub-site `s2` (https://oqtane-poc212.oqtane.me/s2)

![image](https://user-images.githubusercontent.com/2724850/229803626-380d1835-4354-4bb3-907a-8ec65405e3e3.png)

And new sub-site `s2` is working.

![image](https://user-images.githubusercontent.com/2724850/229803824-71b29c10-8acd-4b14-9bc2-bf0b81d8171a.png)

But old sub-site `s1` is not working.

![image](https://user-images.githubusercontent.com/2724850/229804077-dcf86a99-a331-4114-b748-93158dc48c87.png)

And if we **Restart Application** again, it will break sub-site 's2' also.

[...]

PR #2725, opened 2023-04-04
Fixes null reference permissions issue: Fixes #2724 

### LINKED ISSUE #2724: Permissions for Role Name is null causing child sites to fail to load creating an exception error.
It seems the issue was in the Get Permissions.

When you browse from within the framework to a site created by a host administrator other than the master site, the sites would error.

I have a fix for the issue I will share in a minute that helps with the check.

I also experimented with emptying cache permissions and as a by-product I created logic to empty the cache when a site loads to help with this error, but did not cure it.

I can present that as another PR as I added  "ClearPermissionsCache" functionality to the project.   Again maybe something for review I can share as another PR if desired.

thabaum:
Permissions for Role Name is null causing child sites to fail to load creating an exception error. #2724 
 
 I tried to clear cache in other areas like at the top of GetPermissions and LoadModuleDefnitions however did not resolve the issue of trying to access sites created from the master site.
 
 This fix presented in this PR should... allow Oqtane to continue to run while accessing alias sites.
 
 This is probably not the fix as I seen errors now in the event log for the site:
 
 Unauthorized Folder Get Attempt {Path} For Site {SiteId}
 
 however I logged out and back in and no error.
 
But then after browsing user profile I would get notification errors popping up 
 
`Log Message: Unauthorized Folder Get Attempt Users\1\ For Site 2`
 
 So need further testing.  This probably should not be returning a null but some reason it is.

### DIFF
--- a/Oqtane.Server/Repository/PermissionRepository.cs
+++ b/Oqtane.Server/Repository/PermissionRepository.cs
@@ -36,7 +36,7 @@ namespace Oqtane.Repository
                     {
                         if (permission.RoleId != null && string.IsNullOrEmpty(permission.RoleName))
                         {
-                            permission.RoleName = roles.Find(item => item.RoleId == permission.RoleId).Name;
+                            permission.RoleName = roles.Find(item => item.RoleId == permission.RoleId)?.Name;
                         }
                     }
                     entry.SlidingExpiration = TimeSpan.FromMinutes(30);


thabaum:
https://github.com/oqtane/oqtane.framework/issues/2721

It may relate to this issue as well I am hoping...

thabaum:
https://github.com/oqtane/oqtane.framework/issues/2720

looking down the list looks like this one may fall into this fix as well.

thabaum:
Now if I log into the master site (try to) with the account created on Site 2 it acts like the user is logged in no warnings or errors and just takes you to home page.  The user should be told this user does not exist.  

If I make a new users on the master site and log in the user is logged into their profile.

Now for sites that want users to be able to log into all the selected sites and sub domain sites along with sharing components and data how are we going to accomplish this?  I will test this out as needed.  We should just need a wildcard SSL, a domain name and all domains/sub domains working with authentication passing between each for some sites.  Other sites on the same installation may be scoped to only the users that are registered.

Not sure if this falls into this PR but while I am in this area testing things I want to be sure things are as expected.   I am currently testing Oqtane.Server not running IIS Express.

I have been able to load a basic Blazor template site, however not an Oqtane successfully with ease as one discussion points out.  I am hoping we can resolve that as well.

sbwalker:
This fix does not address the root cause of the issue.. it will merely stop an error from being thrown in this specific location - but errors will continue to be thrown in other areas. I located the root cause and I am investigating further.

thabaum:
I have created a few different formulas that keep the system going, I just run into the host folder permission issue when trying to access the Profile after.  Trying to understand the logic for why it would be null.  I see in the database and from the roles it includes the bool to include all the roles for SiteId == null.
```

                return _cache.GetOrCreate($"permissions:{alias.TenantId}:{siteId}:{entityName}", entry =>
                {
                    var roles = _roles.GetRoles(siteId, true).ToList();
                    var permissions = _db.Permission.Where(item => item.SiteId == siteId).Where(item => item.EntityName == entityName).ToList();
                    foreach (var permission in permissions)
                    {
                        if (permission.RoleId != null && string.IsNullOrEmpty(permission.RoleName))
                        {
                            var role = roles.Find(item => item.RoleId == permission.RoleId);
                            permission.RoleName = role?.Name ?? "";
                        }
                    }
                    entry.SlidingExpiration = TimeSpan.FromMinutes(30);
                    return permissions;
                });
```

And 

```
        return _cache.GetOrCreate($"permissions:{alias.TenantId}:{siteId}:{entityName}", entry =>
        {
            var roles = _roles.GetRoles(siteId, true).ToList();
            var permissions = _db.Permission.Where(item => item.SiteId == siteId).Where(item => item.EntityName == entityName).ToList();
            foreach (var permission in permissions)
            {
                if (permission.RoleId != null)
                {
                    var role = roles.Find(item => item.RoleId == permission.RoleId);
                    if (role != null)
                    {
                        permission.RoleName = role.Name;
                    }
                }
            }
            entry.SlidingExpiration = TimeSpan.FromMinutes(30);
            return permissions;
        });
```

also got me to the same place I am with the issue with Host not being able to access profile folder for profile picture access from Site 2.

From the master site 1 the host can edit things no problem.

So issue leads me to the Profile logic or File Manager could be fixed for Host and really things are working as normal as far as I can tell.  I have not seen or recall exactly how it was acting in the past to compare past behaviors in this area.

But it seems the only bug I am finding is with the Host logging into other sites.  I have not tried child sites yet for administrators and users going in and out of portals that should allow a cookie to pass between sites.

But this is what I came up with so far if any of it helps.

sbwalker:
Reopening as although #2731 fixes the root cause of #2724, there is potential for Permission records to be orphaned and not associated to a role.

Issue #2734, opened 2023-04-06
Module Data Deletion


leigh-pointer:
Scenario :
I have a page with a module on it the module has data.  I delete the page go to recycle bin and delete All Page, which inturn deletes the PageModule and the Page.  the module data is not deleted for that instance also the Module Settings.

sbwalker:
When you say that the "module has data" are you saying that it has its own database tables where it stores its data? If so, Oqtane does not know anything about how a module stores its data. In DNN this was typically handled with a cascading foreign key relationship to the Module framework table. Another approach would be to introduce an IDeletable interface which a module can implement.

leigh-pointer:
@sbwalker 
> "module has data" are you saying that it has its own database tables where it stores its data?
yes

I think it can be achieved by the framework as we are currently deleting modules and removing the Module Record executes the cascading delete.

The work around is to Delete every module on the Page then empty All modules in the Reecycle Bin

```
    public class ModDefaultTemplateEntityBuilder : AuditableBaseEntityBuilder<ModDefaultTemplateEntityBuilder>
    {
        private const string _entityTableName = "StudioElfModDefaultTemplate";
        private readonly PrimaryKey<ModDefaultTemplateEntityBuilder> _primaryKey = new("PK_StudioElfModDefaultTemplate", x => x.ModDefaultTemplateId);
        private readonly ForeignKey<ModDefaultTemplateEntityBuilder> _moduleForeignKey = new("FK_StudioElfModDefaultTemplate_Module", x => x.ModuleId, "Module", "ModuleId", ReferentialAction.Cascade);
```

I will produce a fix and add a PR

sbwalker:
Perhaps I don't understand but I don't think this a framework issue. If you delete a module from a page it goes to the recycle bin. Nothing should be deleted at this point because you may want to restore the module. Later if you go to the recycle bin and decide you want to remove the module completely, you choose delete. This will remove the pagemodule instance. Then it needs to check if that module exists on other pages.  If it doesn't then it deletes the module record. However this will not automatically clean up the module's own data in its tables - unless the module uses a cascading foreign key.

leigh-pointer:
@sbwalker 
The  issue doen't exist on the Delete Module or Delete All Modules the isse is with Deleting a Page or Deleting All Pages.  The Delete Page should delete the Page, PageModule and Module records.  Currently it is not deleting the Module Record thus orphaned Module Record and orphaned External Module Data.

[...]

PR #2735, opened 2023-04-06
Fix for Module Data Deletion #2734

leigh-pointer:
Updated the framework to delete 
Page -> PageModule -> Module -> 
[ externa Module data + externa Module Settings ]

### DIFF
--- a/Oqtane.Server/Repository/PageModuleRepository.cs
+++ b/Oqtane.Server/Repository/PageModuleRepository.cs
@@ -4,6 +4,7 @@ using Microsoft.EntityFrameworkCore;
 using Oqtane.Extensions;
 using Oqtane.Models;
 using Oqtane.Shared;
+using Oqtane.UI;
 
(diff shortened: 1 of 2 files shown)


sbwalker:
@leigh-pointer This fix does not seem to be correct - see discussion in #2642 - a module record should only be deleted if there are no other pagemodule records which reference that module on any other pages in a site.

leigh-pointer:
@sbwalker your correct, I keep forgetting that feature.  I will fix it.

sbwalker:
@leigh-pointer this PR still has issues. The PageRepostory.cs (which is server-side) is referencing the IModuleService (which is client-side). I am not even sure how this works - but I am guessing it is because you only tested it on Blazor Server. Either way, this breaks the separation of concerns for client and server and cannot implemented in this way.

sbwalker:
#2872 resolves this issue

PR #2754, opened 2023-04-19
Add ability to use environment variables in order to set them as app settings in Azure

pepsinio:


pepsinio:
@dotnet-policy-service agree

leigh-pointer:
I Agree with this PR

pepsinio:
Not sure whether there are other guides available on how to do CD pipeline into Azure, but happy to share mine approach with this PR in mind.

PR #2765, opened 2023-05-01
Updates Page.ModifiedOn in PageModuleController methods

### LINKED ISSUE #2764: Sitemap LastMod property needs to be set properly
When adding, updating, moving it, or deleting a page module the Page.ModifiedOn property needs to update to reflect the date of changes in the Sitemap.

I have a PR for this at the finish line.

thabaum:
Fixes #2764 

Fix this up as needed if it looks like it will work.  Using` _pages.UpdatePage(page)` was able to update the sitemap and modified on date for pages when a module is moved/deleted/updated/modified(settings).

I am curious however if we should do this as well in the ModuleController.cs file as well for when modules are deleted, updating all pages `ModifiedOn` property to reflect the changes which I had fixed up but found a better path down the road that may be a solution.  I was not sure if this should be logged... so adjust it as you like :)  I hope this helps!

For updating module content I used another approach in the HtmlText updating it in the repository, however I can play with that more as well see if I can work a better solution once I get more feedback here.

The logging and things not related to the `UpdatePage();` function I was not 100% sure was needed so this can be removed as desired I will let you make adjustments if you find this somewhat acceptable.

Thank you!


sbwalker:
Closing this PR based on the comments here https://github.com/oqtane/oqtane.framework/issues/2764#issuecomment-1530144392

PR #2770, opened 2023-05-02
Remove Admin Page/Module Registered User View

thabaum:


sbwalker:
"If you do not like this default behavior you can always adjust the page permissions in your own installation to remove the Registered Users role" - this will result in a 404 page

thabaum:
"If you do not like this default behavior you can always adjust the page permissions in your own installation to remove the Registered Users role" - this will result in a 404 page

I did this, but I guess I am still trying to understand why it is by default allowing registered users access to the admin page.  Shouldn't it be the other way around, if you want to allow a user to perform User/Role/Profile management you grant them access to these areas.

Can you show me or explain the use case in a bit more detail for this being default setting? 

"User/Role/Profile management can be delegated to non-Administrators" but should it be by default?  A user can update their profile still.  I would love to understand this fully.  Is there documentation I can read up on?

sbwalker:
After further review, the delegated admin capability relies on an Administrator setting the permissions for the specific pages/modules (ie. User, Role, Profile) which they would like to delegate - so it makes sense that they would need to set the permissions for the /admin page as well. So I will merge this PR. There is no official documentation for delegated permissions - there is just a blog which attempts to describe the enhancement https://www.oqtane.org/blog/!/55/permission-enhancements and there is a thread in the repo which contains more info as well.

sbwalker:
@thabaum here is the link to information about delegated administration:

@thabaum these are the details for delegated administration: https://github.com/oqtane/oqtane.framework/discussions/2481#discussioncomment-4636995 

You will notice that it mentions that you need to adjust the permissions for the various admin pages/modules you wish to delegate ie. User Management, Role Management, etc... Based on this PR, the instructions will need to be modified as the Admin Dashboard page/module will also need to have its permissions adjusted to include delegated access.

Note that in order for a non-Admin to be able to see the Admin Dashboard, the Index.razor SecurityAccessLevel cannot be set to Admin - or else it would restrict access and there would be no way to override. This is why PR #2776 is not valid.  I do believe the SecurityAccessLevel could be set to View rather than Anonymous without breaking any functionality.

thabaum:
OK I agree with view, now that I read this I think we are on the same page as far as concern goes.  My next concern is changing this has no affect on the page.

PR #2771, opened 2023-05-02
Invalid template when I try and use the Deploy to Azure button

vnetonline:
1. removed depends on connection strings
2. Moved sku outside properties for type Microsoft.Web/serverfarms and added various app plans
3. Removed worker size and replaced with instance capacity
4. Replaced database and sqlserver to use various editions
5. Updated ApiVersions for various types

### DIFF
--- a/azuredeploy.json
+++ b/azuredeploy.json
@@ -2,13 +2,40 @@
   "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
   "contentVersion": "1.0.0.1",
   "parameters": {
+    "sqlDatabaseEditionTierDtuCapacity": {
+      "type": "string",
+      "defaultValue": "Standard-S1-20-250",
+      "allowedValues": [
+        "Basic-Basic-5-2",
+        "Standard-S0-10-250",
+        "Standard-S1-20-250",
+        "Standard-S2-50-250",
+        "Standard-S3-100-250",
+        "Standard-S4-200-250",
+        "Standard-S6-400-250",
+        "Standard-S7-800-250",
+        "Standard-S9-1600-250",
+        "Standard-S12-3000-250",
+        "Premium-P1-125-500",
+        "Premium-P2-250-500",
+        "Premium-P4-500-500" ,
+        "Premium-P6-1000-500",
+        "Premium-P11-1750-500-1024",
+        "Premium-P15-4000-1024",
+        "GeneralPurpose-GP_Gen5_2-2-250",
+        "GeneralPurpose-GP_S_Gen5_2-2-250"
+      ],
+      "metadata": {
+        "description": "Describes the database Edition, Tier, Dtu, Gigabytes (Edition-Tier-Dtu-Gigabytes)"
+      }
+    },
     "sqlServerName": {
       "type": "string",
       "metadata": {
         "description": "The name of the sql server. It has to be unique."
       }
     },
-    "databaseName": {
+    "sqlDatabaseName": {
       "type": "string",
       "metadata": {
         "description": "The name of the sql databaseName. It has to be unique."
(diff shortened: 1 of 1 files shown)


vnetonline:
@dotnet-policy-service agree

sbwalker:
I am looking forward to testing these changes - thank you for investigating this long standing issue!

vnetonline:
Here is a quick guide

Here is the input screen 

image

The `Review & Create` screen 

image

Deployment Complete screen 

image

Here u can see that the database base has been created as a Standard with 20 DTU and 250 GB as selected in the template via the  `Sql Database Edition Tier Dtu Capacity dropdown` as `Standard-S1-20-250`(which is budget friendly :) )

image

Here u can see that B1 was selected as the App Service Plan 

image

In this instance, I selected 2 as the `Blazor SKU Capacity`so the instance count is 2

image

When we navigate to the site we can see, be sure to select Security custom and use the SQL Administrator login and password with Encryption set to True and Verifiable

image

Once you click Install all should be good and you will see `Welcome to Oqtane`

image

sbwalker:
Thank you... this is excellent! Quick question on Blazor SKU Capacity - by setting it to a value greater than 1, are you essentially saying that you want to use a web farm?

vnetonline:
> Thank you... this is excellent! Quick question on Blazor SKU Capacity - by setting it to a value greater than 1, are you essentially saying that you want to use a web farm?

The quick answer is not exactly 

here is what Microsoft have to say

`Autoscale is a built-in feature that helps applications perform their best when demand changes. You can choose to scale your resource manually to a specific instance count, or via a custom Autoscale policy that scales based on metric(s) thresholds, or schedule instance count which scales during designated time windows. Autoscale enables your resource to be performant and cost effective by adding and removing instances based on demand. [Learn more about Azure Autoscale](https://docs.microsoft.com/azure/azure-monitor/platform/autoscale-get-started?WT.mc_id=Portal-Microsoft_Azure_Monitoring) or [view the how-to video](https://www.microsoft.com/videoplayer/embed/RE4u7ts).`

My understanding is ... when you use `App Service Plan` you can Scale Out instances which are automatically managed by azure during busy periods of time if you see in the azuredeploy.json the section which create the App Service Plan as follows 

    {
      "name": "[variables('hostingPlanName')]",
      "type": "Microsoft.Web/serverfarms",
      "location": "[resourceGroup().location]",
      "apiVersion": "2022-09-01",
      "dependsOn": [],
      "tags": {
        "displayName": "Blazor"
      },
      "sku": {
        "name": "[parameters('BlazorSKU')]",
        "capacity": "[parameters('BlazorSKUCapacity')]"
      },
      "properties": {
        "name": "[variables('hostingPlanName')]",
        "numberOfWorkers": 1
      }
    },

Refers to `Microsoft.Web/serverfarms`

But you could create your own VM and use the `Azure Load Balancer Service` (which can get very expensive quickly ) and create a server farm, but why go to that effort when the App Service Plan has Auto Salcing when Higher Plans have Auto Scaling included. I guess an enterprise may want to set this up if they want more control.

sbwalker:
@vnetonline I was aware of the Autoscale configuration in Azure... and that you can set rules to scale out your infrastructure based actual usage. Essentially when an Autoscale threshold is triggered, it creates a new "instance" of your application. The challenge is that it assumes applications can support multiple instances... and this is not true for all applications. For example if an application utilizes standard IMemoryCache it is only valid for a single instance - it does not coordinate cache invalidation across multiple instances. So if your application is dynamic in nature you may notice inconsistent behavior depending on which instance you are connected to. In addition, IHostedService will run on all instances which may not be the desired behavior as it could lead to repetitive work being done in the background.

sbwalker:
@vnetonline I will add Azure Autoscale support to the roadmap for 4.x (there are a number of modifications which would be needed to support this)

PR #2772, opened 2023-05-02
improve module/theme/translation upload user experience to be consistent with download

sbwalker:


sbwalker:
this completes the changes from #2730

PR #2776, opened 2023-05-03
Updated Dashboard Index

thabaum:
This pr updates dashboard index SecurityAccessLevel to match other module pages.  Clarity here maybe needed to understand this override.

I set this to Admin and tested all scenarios where you may have another role with a registered user manage admin/roles and everything worked as expected.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Dashboard/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Dashboard/Index.razor
@@ -22,7 +22,7 @@
 @code {
 	private List<Page> _pages;
 
-	public override SecurityAccessLevel SecurityAccessLevel => SecurityAccessLevel.Anonymous;
+    public override SecurityAccessLevel SecurityAccessLevel => SecurityAccessLevel.Admin;
 
 	protected override void OnInitialized()
 	{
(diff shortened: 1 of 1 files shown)


sbwalker:
Can you please explain the specific problem you are fixing with this PR? It is critical for others to be able to reproduce the problem so that the solution can be validated.

thabaum:
SecurityAccessLevel.Admin is set on all other dashboard admin files except the main dashboard file which is set to anonymous.  Shouldn't this be set to Admin as well?  I am trying to understand why it would not be set the same as other admin dashboard files.  


SecurityAccessLevel.View: Allows users to view a resource, but not to modify it.

SecurityAccessLevel.Edit: Allows users to view and modify a resource.

SecurityAccessLevel.Admin: Allows users to view, modify, and manage the system's administrative settings.

SecurityAccessLevel.Anonymous: Allows anonymous users (i.e., users who are not authenticated) to access a resource.

SecurityAccessLevel.Unauthorized: A user is not authorized to access a resource.
    
    I will close this as not having an issue to reproduce other than a concern.  Should this be set to View at least?  I am trying to understand how this works in the admin dashboard and related files.

thabaum:
I see your comment on #2770, I adjusted this PR to View which is more acceptable to me as well, however the issue changing this having no change on the file behavior maybe a concern?

As in you can set this override setting to whatever you want, the page is really only using the PermissionList to dictate who can view/edit this page.  At least from my testing it had no change, maybe it works on your end and something was not right on mine?

PR #2800, opened 2023-05-09
Unhandled error when creating module instance on page due to missing ModuleDefinition in Module table 

chlupac:
When a module is defined in the Module table but does not have a corresponding definition in the ModuleDefinition table, the program will cause an unhandled error. The module.ModuleDefinition is initialized with default values in the controller and it is not possible to create an instance of the type. This error occurs during the creation of module instances on the page.

### DIFF
--- a/Oqtane.Client/UI/SiteRouter.razor
+++ b/Oqtane.Client/UI/SiteRouter.razor
@@ -403,7 +403,9 @@
 			if ((module.PageId == page.PageId || module.ModuleId == moduleid))
 			{
 				var typename = Constants.ErrorModule;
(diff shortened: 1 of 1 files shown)


sbwalker:
@chlupac thank you for identifying this issue. This was introduced as a result of the recent changes to filter some of the ModuleDefinition information sent from the server to the client by the API controllers. Rather than adding some defensive logic to the client, I believe the better solution would be to modify the FilterModuleDefinition() method so that if the object passed to the method is null, it returns null (returning an initialized object in this case is not a valid behavior).

chlupac:
@sbwalker,
I agree with you. It's better to modify the FilterModuleDefinition() method to return null if the object passed to the method is null instead of returning an initialized object. Closing this PR.

Issue #2508, opened 2022-11-24
Feat: Service to programmatically add custom Headers such as Open-Graph on Prerender


iJungleboy:
This is a continuation of #1967, in which this was already discussed. 

The solution for it was #1984 which added a field in the page settings to add meta-headers manually. But that's not exactly what we had in mind. 

Our need is that a software developed on Oqtane can add headers to the prerender. Example use case:

1. A blog which adds open-graph headers which are different for each post
2. A public Recipe DB or Real-Estate listing, which needs to add Open-Graph headers
3. A system which can add no-index headers to certain pages which shouldn't be available in any crawler

In many such cases there is not a real Oqtane page per visible page, and the module on the page would need to be able to add these headers. I'm specifically concerned about pre-render, but of course a service that would work in all cases would be even better.

Your thoughts?

sbwalker:
If you need to inject information into the page during pre-render, the logic needs to be executed server-side (ie. in _Host.cshtml). However it sounds like you want to be able to include this logic in your module components.  The problem is that modules are only executed client-side - they are razor components which are dynamically rendered and depend on state which is generated by the client-side router. So it is not possible to instantiate modules server-side and execute logic. The only reason why the Resources concept works is because it is a stateless static property. More research will be required for this feature.

iJungleboy:
So I may have the wrong idea here, but this is what I understand:

* The client can run in the browser or in a **very isolated** thread on the server in something that I'll just call the **Client Process** for now. 
* The initial HTML is generated in the **Server Process** (or please correct me if this actually happens in the Client Process)
* Pre-Render probably augments this, by letting the _Server Process_ talk to the _Client Process_ and inserting the final HTML into the initial HTML, simulating what the browser would do

So this opens these questions:

1. Is the process I think is happening correct, or does the Server Process also render the client stuff (is it the same process, in which case it would know more and also be able to share services - in which case it would be easy to have a service transfer the necessary headers)
2. If the process is completely isolated, how could the client pass not just the HTML, but additional data to the prerenderer...

sbwalker:
[...]

PR #2801, opened 2023-05-10
Feat: Ability to set the Google Analytics MeasurementID in Site Settings

vnetonline:
Ability to set Google Analytics MesurementID in `Site Settings` Screen

1. Added the GoogleAnalystics Section to the Site Settings Screen 
2. Modified the _Hosts.cshtml to embed the Google Analytics Script between the header and body tags


sbwalker:
@vnetonline thank you for submitting this PR. I have a few questions/comments below...

In general when it comes to adding new functionality to the framework, the main questions are:

1. Could this feature be implemented as a custom extension? 
2. Is this feature applicable to the majority of users?

In regards to #1 the framework does not currently include support for injecting arbitrary elements into the head of a page - which is why you submitted this PR. However there are many types of functionality which need to be declared in the head of a page, and as a framework, Oqtane really should provide support for these types of integrations in a generic manner (rather including specific implementations for each feature). However, there are a lot of considerations to consider. Depending on the integration, the feature may need to be declared at either the Site or Page level (and sometimes with a requirement for overrides). There are also a variety of mechanisms which developers will need to leverage the functionality - statically where the information is stored in a database (which is exemplified by this PR), declaratively where it is embedded in code during compilation, and dynamically where it is generated at run-time. And obviously the solution would need to work consistently across Blazor hosting models. 

In regards to #2 you need to keep in mind that Oqtane is used in variety of different ways in a variety of different environments. For example, if someone had built an internal HR application using Oqtane and it was only intended of running in an internal environment then there would never be any need for Google Analytics functionality.

So my preference in this case would be to include the foundational capability so that developers can inject elements into the head of the page. This would provide the flexibility for creating a custom module to accomplish your goal. However this is technically much more challenging - especially the requirement for supporting dynamic content (which is logged under issue #2508)

vnetonline:
@sbwalker I agree with both #1 and #2 because Oqtane doesn't support the ability to inject arbitrary elements into the head of the page is why I submitted this PR

Also, Google documentation suggests that the best place to have the Google Analytics Script is between the Head of the Body of the page, so even if Oqtane had the ability to inject arbitrary elements into the head of the page in my opinion it would still need to be implemented in this way. In terms of overriding then, we could introduce a setting `isGAPageTracking ENabled` so we can control if it should be disabled or enabled on specific pages.

I also understand that Oqtane will be used in various different environments as per your example of an internal HR app. That's why if you don't provide a value in the MesurementID or setting or we introduce an `isEnabled` setting at the site level. Google Analytics script will not appear.

As part of Oqtane .Net 7.0 support are you going to plan to use the new [Head Outlet Component](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/control-head-content?view=aspnetcore-7.0) so thet developers can inject arbitrary elements into the head of the page?

Also in order for developers such as myself who have something like 22 sites and want to uptake Oqtane, Oqtane really needs to provide support for both Application and Framework extensibility so that developers can extend the functionality of the Oqtane for different use cases, either as a custom module or custom framework extension.

sbwalker:
HeadOutlet was discussed in #2508 - it was intended to solve this problem but the technical implementation is unfortunately flawed which means that it does not behave consistently on all hosting models. Basically the implementation on Server is totally different than WebAssembly (including additional dependencies) and because Oqtane supports both hosting models from a single code base it is not a viable solution.

sbwalker:
"In order for developers such as myself who have something like 22 sites/apps and want to uptake Oqtane, Oqtane really needs to provide support for both Application and Framework extensibility so that developers can extend the functionality of the Oqtane for different use cases, either as a custom module or framework module." - this is the primary goal of Oqtane... but it obviously takes significant time/engineering to enable all of the use cases.

sbwalker:
@vnetonline I believe I have found a viable solution to inject arbitrary elements into the page head. It works on Blazor Server and Blazor WebAssembly and will provide a number of other framework benefits as well. I will explain more this coming week.

vnetonline:
> @vnetonline I believe I have found a viable solution to inject arbitrary elements into the page head. It works on Blazor Server and Blazor WebAssembly and will provide a number of other framework benefits as well. I will explain more this coming week.

That's excellent please let me know how I can me of assistance

PR #2807, opened 2023-05-15
initial changes to upgrade to .NET 7

sbwalker:


sbwalker:
I also installed the Blog module using the standard Download method in Module Management and it installs/runs fine (note that the Blog module is a .NET 6 module at this time).

PR #2812, opened 2023-05-16
Enhance - User Profile Page to smooth up to the top of the page when …

vnetonline:
…saved to see the notification


sbwalker:
@vnetonline I added validation support to user profile fields in #2843 and added the smooth scroll so I will close this item

PR #2832, opened 2023-05-23
format head content, remove scripts, and filter duplicate elements

sbwalker:


sbwalker:
@leigh-pointer thank you!

PR #2895, opened 2023-06-14
Fixed targeting .NET 7.0 in ThemeController and ModuleDefinitionContr…

vnetonline:
…oller missed in #2888 and #2890

### DIFF
--- a/Oqtane.Server/Controllers/ModuleDefinitionController.cs
+++ b/Oqtane.Server/Controllers/ModuleDefinitionController.cs
@@ -330,9 +330,9 @@ namespace Oqtane.Controllers
                     if (moduleDefinition.Version == "local")
                     {
                         text = text.Replace("[FrameworkVersion]", Constants.Version);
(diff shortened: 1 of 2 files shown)


vnetonline:
Fixing the Libraries needed for declaring theme resources in new way

sbwalker:
I don't believe the framework should have used ".Themes" - it should have used ".Theme" but it's too much work to change the names of default themes at this point due to upgrade challenges. Including ".Theme" in the wwwroot is redundant currently as it's already within a Themes parent folder. However one item on the future roadmap is to transition to using the more standard .NET _content folder approach for static assets which would mean it would be useful to include the identifier in the folder name. And it would be useful for assembly names as well so that you can easily be able to differentiate themes from modules in the /bin. So I would be in favor of using [Organization].Theme.[Theme] as a naming convention.

sbwalker:
Just submit a new PR for the change

PR #2902, opened 2023-06-19
optimize head component rendering

sbwalker:


sbwalker:
@leigh-pointer that is good news... the component was re-rendering more often than necessary. The real question is if it improved the "flash" scenario. To be honest I am having a really difficult time reproducing the "flash" scenario - it does not happen on Firefox, and it only happens on Edge if I Disable Caching and do a browser refresh - both of which are not common user configurations during standard browsing activity.

leigh-pointer:
@sbwalker I don't normally use Firefox, so I gave that a try and now also no flash and in all the performance is much better.

sbwalker:
@leigh-pointer that is why Firefox is my preferred browser... however it is not a real solution to the FOUC problem reported elsewhere.

sbwalker:
@leigh-pointer can you please get the latest code which includes #2905 and let me know if the flash is still occurring for you?

leigh-pointer:
@sbwalker tried with both Edge and Fire, there is still a flash moving between themes

sbwalker:
@leigh-pointer what about the page load scenario - either when first loading a page on startup, or refreshing a page in the browser? This is the main scenario I am focused on, as in a real world site/app you generally do not use completely different themes for different pages - generally they would use the same theme with some variations for layout, etc... for specific pages.

vnetonline:
@sbwalker I just got the latest dev branch and tried and still happens it flashes on page reload with the theme set at site level in Chrome

leigh-pointer:
@sbwalker 
> what about the page load scenario - either when first loading a page on startup, or refreshing a page in the browser? This is the main scenario I am focused on, as in a real world site/app you generally do not use completely different themes for different pages

**This works, no flash** on browsers tried; Edge, Fire Fox, Chrome, Epic, Brave

Changing between themes causes a flash.
> generally they would use the same theme with some variations for layout.  

I still think this needs to be addressed, as we allow the option to change Theme per page.

vnetonline:
@leigh-pointer on startup and on page load it is not working for me on Chrome what chrome version are you using ?

image

are you using same version?

vnetonline:
@leigh-pointer I am trying your theme and I get a flash see below ... it is more random now but it breaks 

image

leigh-pointer:
@vnetonline  the RocketFuel theme you have is using .Net6
attached RocketFuel text build 3.0.9 for .Net7 
[StudioElf.RocketFuel.3.9.0.zip](https://github.com/oqtane/oqtane.framework/files/11797487/StudioElf.RocketFuel.3.9.0.zip)

Net7-LoadingFlash

vnetonline:
@leigh-pointer can you please share with me how you are declaring the stylesheets in your Theme Info please

leigh-pointer:
@vnetonline the new way using the ThemeInfo Resources, something like this below.


```
        public Theme Theme => new Theme
        {
            Name = "ExternalDefault",
            Version = "1.0.0",
            PackageName = "StudioElf.ExternalDefault",
            Resources = new List<Resource>()
            {
		// obtained from https://cdnjs.com/libraries
                new Resource { ResourceType = ResourceType.Stylesheet, Url = "https://cdnjs.cloudflare.com/ajax/libs/bootstrap/5.3.0/css/bootstrap.min.css", Integrity = "sha512-t4GWSVZO1eC8BM339Xd7Uphw5s17a86tIZIj8qRxhnKub6WoyhnrxeCIMeAqBPgdZGlCcG2PrZjMc+Wr78+5Xg==", CrossOrigin = "anonymous" },
                new Resource { ResourceType = ResourceType.Stylesheet, Url = "~/Theme.css" },
                new Resource { ResourceType = ResourceType.Script, Url = "https://cdnjs.cloudflare.com/ajax/libs/bootstrap/5.3.0/js/bootstrap.bundle.min.js", Integrity = "sha512-VK2zcvntEufaimc+efOYi622VN5ZacdnufnmX7zIhCPmjhKnOi9ZDMtg1/ug5l183f19gG1/cBstPO4D8N/Img==", CrossOrigin = "anonymous" }
            }
        };
```

vnetonline:
spoke to soon still breaks

sbwalker:
@vnetonline since no one else seems to be able to reproduce the behavior you are experiencing, can you please check your Chrome browser settings to verify if you have customized them in any way ie. in the Network tab have you Disabled Caching or turned on Throttling?

sbwalker:
I am adding the older logic for managing stylesheets back into the dev branch to validate if that resolves the issue. The problem is that there were 2 aspects to the older logic - JS Interop in the client, and logic in _Host to include the resources in the initial page load on the server. The enhancement in 4.0 to allow Resources to be specified in ThemeInfo creates challenges for the _Host aspect as it would require additional reflection on every page load - which is one of workloads I was trying to eliminate in 4.0. So there is a possibility that the new Resources enhancement may need to be rolled back as well... however it will depend on whether the older stylesheet logic actually resolves the "problem" (I put "problem" in quotes because it feels like we are chasing a ghost with this issue - as it cannot be reliably reproduced).

sbwalker:
@leigh-pointer @vnetonline #2914 adds the old logic for managing style sheets via JS Interop - please test to determine if this is an improvement over the native Blazor renderer

leigh-pointer:
@sbwalker I am doing it right now.

leigh-pointer:
@sbwalker ran a simple light theme no flash between same theme, small hic between different themes.  However the Resources was / is still being loaded from ThemeInfo class.

sbwalker:
Switching between different themes on the same site will always be problematic unless they share the exact same version and source for the Bootstrap CSS and JavaScript. This is due to the characteristics of a SPA application. For example if Page1 has Theme1 and it uses CSS/JS from SourceA then the CSS resources will be included in the head and the JS will be loaded (note that JS cannot be unloaded once it is loaded). Then you navigate to Page2 which uses Theme2 and uses CSS/JS from SourceB. The CSS will need to be completely replaced in the head (which will cause a flash in the browser) AND it will load the JS library a second time (because the source is different) which will cause the JS functionality to stop working.  The only realistic way to handle this scenario is to do a full server refresh as part of the navigation to the new page.

vnetonline:
> @vnetonline since no one else seems to be able to reproduce the behavior you are experiencing, can you please check your Chrome browser settings to verify if you have customized them in any way ie. in the Network tab have you Disabled Caching or turned on Throttling?

@leigh-pointer the throttling or the cache is not on in my Chrome

image

@sbwalker & @leigh-pointer I am happy to report that I am **not getting the flash** any more after the revert to old logic for style sheets .. thank you for working on this I know it must have been frustrating for both of you to resolve this issuewhich was difficult for both of you to reproduce.

here is what i see now 

screen2gif-nicetheme-noflash

PR #2910, opened 2023-06-20
[ENHANCE] - Added Module to NameSpace [Owner].Module.[Module]

vnetonline:
[ENHANCE] - Added Module to NameSpace [Owner].Module.[Module]


sbwalker:
@vnetonline after further research... this PR impacts the ability for people to create their own custom templates. The template feature was developed with extensibility in mind... you can include your own custom templates beneath the Oqtane.Server\wwwroot\Modules\Templates folder and the UI will automatically recognize them and allow you to choose them from the list. However the problem is that there are some properties which are set independently from the template - they are set in code:

```
                    moduleDefinition.ModuleDefinitionName = moduleDefinition.Owner + ".Module." + moduleDefinition.Name + ", " + moduleDefinition.Owner + ".Module." + moduleDefinition.Name + ".Client.Oqtane";
                    moduleDefinition.ServerManagerType = moduleDefinition.Owner + ".Module." + moduleDefinition.Name + ".Manager." + moduleDefinition.Name + "Manager, " + moduleDefinition.Owner + ".Module." + moduleDefinition.Name + ".Server.Oqtane";
```

So if you were using a custom module template, this PR now forces a ".Module." segment in your namespaces which would be incompatible with your custom template. @leigh-pointer has some custom module templates which are available for download which likely now have issues due to this change. I believe that the optimal solution is for a template to be able to describe their behavior so that the framework does not make incorrect assumptions or force a specific approach. Perhaps these properties could be added to the Template.json file to mitgate this issue. This will require more research.

vnetonline:
@sbwalker  I see your point, on a further thought about this, wouldn't it be better if developers were able to specify their namespace as a whole?

 For example, a developer could have various related modules in a simple case:

`Acme.Bookstore` which has 

- The owner `Acme` and product `Bookstore

A developer could also have a more complex case like:

Acme.Software.Bookstore which has 

- The owner `Acme`, 
- The product `Software`
- The module in that software `Bookstore`

Further, they could have functionality within their module like Acme.Software.Bookstore.Payments

- The owner `Acme`, 
- The product `Software`
- The module in that software `Bookstore`
- The functionality in the Bookstore `Payments` or `Cart` etc

we could then parse the Namesapce to take the last element (nested Namespace) to create the `ServerManagers` like `PaymentManager` or CartManager or `PaymentRespository` or `CartManager`

Also, the template creator doesn't have plural in mind for rexample if I have the following 

Amazing.Module.Audience

The table name and Get List methods should be called `Audiences` which is plural of the module called `Audience`, however, the Entity should be Audience as it defines one Audience 

So having fields for just Owner and Product in the UI restrict this ability to be truly extensible for the developer we could just have one field to fill in which is their NameSpace.

This would remove the `.Module.` and allow the developer to have a lot more flexibility, with just some convention which would conform to some sort of consistency.

open to your thoughts!

sbwalker:
@vnetonline after reviewing this further, there may not be any need for the logic which sets the ModuleDefinitionName or ServerManagerType... these are only required for the Module Creator scenario (ie. where someone adds a Module Creator modue to a page). There is talk in a different thread about removing that option and only relying on the Create Module option in Module Management (which does not require the logic above).

In regards to the suggestion about asking the user for a Namespace, why couldn't a developer simply enter "Acme.Software.Bookstore" as their Organization? I believe the templating logic supported this previously. In fact, if you wanted to use the convention which includes ".Module" you could have entered it into the Organization field as "MyCompany.Module". But after the changes in this PR it actually forces the ".Module" segment whether you want it or not. I am wondering if this needs to be rolled back.

vnetonline:
Yup I agree we should roll this back

vnetonline:
Let me know if you want me to submit a PR

sbwalker:
@vnetonline #3200 introduces a Namespace property to the template.json which can be used to define the naming convention in a template. It eliminates the hard-coding and restores freedom to custom templates.

vnetonline:
Thanks @sbwalker i will give it a test and report back

sbwalker:
@vnetonline I am guessing you did not have a chance to test this. It appears that my change caused a problem with ServerManagerType which I will need to fix in 4.0.4.

vnetonline:
Sorry @sbwalker didn't get chance to to test this I will find some time today

Issue #2913, opened 2023-06-20
Create Module change some files automatically for windows and Linux


AnjaSimons:
I work successful with this project on Ubuntu with rider. I like this very much. But in creating a module, the automatically projectfiles are only for window. Please add two files named debug.sh and release.sh. 
debug.sh:
cp -f "../Client/bin/Debug/net7.0/xxx.Client.Oqtane.dll" "../../oqtane.framework/Oqtane.Server/bin/Debug/net7.0/"
cp -f "../Client/bin/Debug/net7.0/xxx.Client.Oqtane.pdb" "../../oqtane.framework/Oqtane.Server/bin/Debug/net7.0/"
cp -f "../Server/bin/Debug/net7.0/xxx.Server.Oqtane.dll" "../../oqtane.framework/Oqtane.Server/bin/Debug/net7.0/"
cp -f "../Server/bin/Debug/net7.0/xxx.Server.Oqtane.pdb" "../../oqtane.framework/Oqtane.Server/bin/Debug/net7.0/"
cp -f "../Shared/bin/Debug/net7.0/xxx.Shared.Oqtane.dll" "../../oqtane.framework/Oqtane.Server/bin/Debug/net7.0/"
cp -f "../Shared/bin/Debug/net7.0/xxx.Shared.Oqtane.pdb" "../../oqtane.framework/Oqtane.Server/bin/Debug/net7.0/"
cp -rf "../Server/wwwroot/"* "../../oqtane.framework/Oqtane.Server/wwwroot/"

And change the Package.csproj Target as follow:
```
<Target Name="PostBuild" AfterTargets="PostBuildEvent">
    <Exec Condition="'$(OS)' == 'Windows_NT' And '$(Configuration)' == 'Debug'" Command="bash $(ProjectDir)debug.cmd" />
    <Exec Condition="'$(OS)' != 'Windows_NT' And '$(Configuration)' == 'Debug'" Command="bash $(ProjectDir)debug.sh" />
    <Exec Condition="'$(OS)' == 'Windows_NT' And '$(Configuration)' == 'Release'" Command="bash $(ProjectDir)release.cmd" />
    <Exec Condition="'$(OS)' != 'Windows_NT' And '$(Configuration)' == 'Release'" Command="bash $(ProjectDir)release.sh" />
</Target>
```
So the Project will also run in a linux environment. I have tried this right now. It would be nice, if a linux developer must not do this alway per hand.

sbwalker:
[...]

PR #2924, opened 2023-06-22
external template changes to support non-Windows environments

sbwalker:


sbwalker:
fix #2913 credit @AnjaSimons

Issue #2633, opened 2023-03-04
Theme and container settings included in default template


thabaum:
Default template should include settings for turning features in the theme on and off such as the login component. ?  I can try to PR this if this is correct which I believe it is.

We can also create a "Blank" template possibly to select at time of theme creation, however the default theme template should be as feature packed as Oqtane default theme.

thabaum:
I see the localization and things could be included, I almost have a working prototype to offer that would align with the current Oqtane theme (same code migrated to the theme templates)

I will see how the last PR works out with the necessary basic changes.  Then I can add the rest what is needed to get all the stock template features offered in the default theme template.

sbwalker:
I am not opposed to including theme and container settings in the default template if it makes it easier for developers/designers to use the framework

thabaum:
I will doctor up a PR if it hasn't been done yet next time I can fire things up for it.  I believe it would help ease a start up project for a designer/developer.

thabaum:
I will get this done just prior to .NET 8 LTS release to incorporate any updates when that gets more towards release time so only minor adjustments may be needed if necessary.

vnetonline:
Pull Request #2940 submitted to address this issue

vnetonline:
@sbwalker or @leigh-pointer 

Can there be a Theme or Container setting for each Theme or Container?

From what I can see there can only be one `ThemeSettings` and `ContacinSetting` defined in the ThemeInfo?

leigh-pointer:
@vnetonline 
you would need to extend the namespace.. ie
StudioElf.RocketFuel.Theme1
   ThemeInfo
StudioElf.RocketFuel.Theme2
   ThemeInfo
StudioElf.RocketFuel.Theme3
   ThemeInfo

@sbwalker correct me if i am wrong?

vnetonline:
I don’t think that’s right if you extend namespace and want to use base method like @ThemePath() to reference assets it wouldn't work

sbwalker:
@vnetonline you are correct that you can only define 1 theme setting or container setting component for a theme *regardless of how many different theme or container components exist). It would be possible for you to enhance your theme setting component to allow a user to select from a list of possible of theme components that exist in your theme (ie. theme1, theme2, etc... ) and then display UI options that are specific to the one chosen (and load/save those options with the unique theme name in the setting name to differentiate them).

[...]

PR #2940, opened 2023-06-28
Added ThemeSettings and ContainerSettings to Default Theme Template

vnetonline:
Added ThemeSettings and ContainerSettings to Default Theme Template to Demonstrate how these features could be added to developers designing a theme as requested and discussed in Issue #2633

### DIFF
--- a/Oqtane.Server/Controllers/SiteController.cs
+++ b/Oqtane.Server/Controllers/SiteController.cs
@@ -246,7 +246,7 @@ namespace Oqtane.Controllers
                 foreach (Page child in children)
                 {
                     child.Level = level + 1;
-                    child.HasChildren = pages.Any(item => item.ParentId == child.PageId && !item.IsDeleted);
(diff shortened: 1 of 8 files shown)


sbwalker:
@vnetonline these changes look valid however my feedback is that the templates are not intended to replicate the features of the default templates - they are intended to provide a simple demonstration of how to use various Oqtane features. On that note, if we are going to add Theme Settings and Container Settings to the theme template, I would suggest that we only include a few setting options. For Theme Settings: Show Login and Show Register, and for Container Settings: Display Title. The other options should be removed. It also appears that the solution file was accidentally removed in this PR (https://github.com/oqtane/oqtane.framework/commit/3b0ffde1fb029d73595924dfb749842f32606590) - so I am glad to see this adds it back.

vnetonline:
[e2a9470 ](https://github.com/oqtane/oqtane.framework/pull/2940/commits/e2a9470f3112d09fa592290773e729b31decaedb) removed some settings as per your suggestion @sbwalker to demonstrate how to add simple Theme and Container Settings to a developer

PR #2964, opened 2023-07-01
Show all themes on page add/edit

markdav-is:
Found a small typo what would limit the theme dropdowns on page add/edits to show only the default theme.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Pages/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Pages/Add.razor
@@ -246,7 +246,7 @@
             if (UserSecurity.IsAuthorized(PageState.User, RoleNames.Admin) || (_parent != null && UserSecurity.IsAuthorized(PageState.User, PermissionNames.Edit, _parent.PermissionList)))
             {
                 _themetype = PageState.Site.DefaultThemeType;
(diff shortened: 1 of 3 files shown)


sbwalker:
@markdav-is this is actually not a typo - it was a deliberate change in 4.0. There is a long thread about this topic but the summary is that allowing completely different themes to be used on different pages within the same site has a lot of technical problems which cannot be resolved (due to SPA characteristics). It only works in scenarios where the themes are using the exact same version and source of JavaScript and CSS libraries (like the default Oqtane themes). However if themes use a different version/source of these JavaScript libraries, there will be conflicts and the site will break. Therefore a change was made so that you can only use themes from the same theme package within a site.

sbwalker:
@markdav-is closing this PR as I explained the expected behavior above in 4.0

markdav-is:
Thanks for the explanation.  I'll check out that thread.  We use the built-in theme for the admin pages and dashboard.  We use a non-boostrap theme for the actual apps.  I feel like we can have two themes as long as they don't both use boostrap.  With this change, I feel like we need to re-build all the admin UI in our custom theme.  So our themes will need to be "complete" and not just for apps.  This will add a lot of opportunity costs to our projects.

sbwalker:
@markdav-is I am going to revert the behavior and allow overriding themes at the page level.  I will simply add a warning message to the user.

Issue #2965, opened 2023-07-01
Update RazorLangVersion from 3 to 7


W6HBR:
While trying to utilize a .Net7 feature for @bind:get & @bind:set, the compiler throws an error due to the <RazorLangVersion> not being set to the current version:  **RZ10020 Attribute 'bind:get' can only be used with RazorLanguageVersion 7.0 or higher**.

Updating my project file to RazorLanguage 7 resolved the issue, but the Oqtane module template and project files should be updated from <RazorLangVersion>3.0</RazorLangVersion> to <RazorLangVersion>7</RazorLangVersion> for compatibility.

Oddly enough, the generated file Oqtane.Server.GeneratedMSBuildEditorConfig.editorconfig contains the 7.0 reference, but not the other projects in the Oqtane solution.

Example code: `<input id="StockNum" class="form-control" @bind:get="@_StockNum" @bind:set="SetStockNum" required />`

sbwalker:
@W6HBR & @leigh-pointer it appears that Microsoft made a change at some point, and RazorLangVersion no longer needs to be specified:

"Razor language version - When targeting the Microsoft.NET.Sdk.Web SDK, the Razor language version is inferred from the app's target framework version. For projects targeting the Microsoft.NET.Sdk.Razor SDK or in the rare case that the app requires a different Razor language version than the inferred value, a version can be configured by setting the <RazorLangVersion> property in the app's project file. Razor's language version is tightly integrated with the version of the runtime that it was built for. Targeting a language version that isn't designed for the runtime is unsupported and likely produces build errors."

To test this, I created a new default Blazor WebAssembly app on .NET 7 and it does not have RazorLangVersion specified:

```
<Project Sdk="Microsoft.NET.Sdk.BlazorWebAssembly">

  <PropertyGroup>
    <TargetFramework>net7.0</TargetFramework>
    <Nullable>enable</Nullable>
    <ImplicitUsings>enable</ImplicitUsings>
  </PropertyGroup>
```

So we should remove the RazorLangVersion references

leigh-pointer:
@sbwalker ok I will create a new PR

W6HBR:
Thanks guys.  I can confirm removing the reference from the client projects gives the intended results.

PR #2966, opened 2023-07-01
Fix for #2965 RazorLangVersion from 3 to 7

leigh-pointer:
Updated the RazorLangVersion from 3 to 7


leigh-pointer:
Closed see @sbwalker comment on #2965

PR #2970, opened 2023-07-03
Add Module name to edit mode action dropdown

jlidbeck:
As a content creator, I find it much easier to see which module I'm editing by clicking the dropdown arrow menu.


sbwalker:
@jlidbeck this PR is blocked because you have not accepted the CLA with the .NET Foundation. In regards to the PR itself, we should not impose cosmetic changes on this component as it could adversely affect third party themes which have been developed which have assumed a specific behavior. An alternate suggestion is to include the Module Definition Name in the Module Settings component so that it is accessible but does not affect the main page UI.

sbwalker:
replacing with #2998

Issue #2979, opened 2023-07-06
When a Registred user Clicks on My Page which is personalised an error is through becuase Display Name is Null


vnetonline:
When a `Registred User` clicks on My Page which is personalised an error is thrown becuase `DisplayName` of the user is `Null` and then when saving into the DB results in error 

![image](https://github.com/oqtane/oqtane.framework/assets/15242923/30e58f75-7d21-467c-803d-526dfcfd10b1)

![image](https://github.com/oqtane/oqtane.framework/assets/15242923/9e6baac0-369d-43cf-96f6-cf3765b15cc7)

I have changed this to be `Username` beucase it will always be there and submitted a pull request

vnetonline:
If the user has the Role of `Host Users` then the host user should not have a personalised page??

PR #2980, opened 2023-07-06
Fixing personalized page created to be UserName if DisplayName is null

vnetonline:
Fix for issue #2979

### DIFF
--- a/Oqtane.Server/Controllers/PageController.cs
+++ b/Oqtane.Server/Controllers/PageController.cs
@@ -183,7 +183,7 @@ namespace Oqtane.Controllers
                 page = new Page();
                 page.SiteId = parent.SiteId;
                 page.ParentId = parent.PageId;
-                page.Name = user.DisplayName;
+                page.Name = user.DisplayName != null ? user.DisplayName : user.Username;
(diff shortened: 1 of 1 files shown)


sbwalker:
@vnetonline I need to take a close look at this as the logic in the SiteRouter assumes a specific path convention for personalized pages - and this PR creates some variability in the path which will be a problem.

vnetonline:
@sbwalker Looking at the  `SiteSouter` I found the following which created a Personalized page if it doesn't exist looks like it is using `UserName`

https://github.com/oqtane/oqtane.framework/blob/dev/Oqtane.Client/UI/SiteRouter.razor#L224-L240

```
 if (page == null)
            {
                // look for personalized page
                page = await PageService.GetPageAsync(route.PagePath, site.SiteId);
            }
            else
            {
                if (user != null && page.IsPersonalizable)
                {
                    var personalized = await PageService.GetPageAsync(route.PagePath + "/" + user.Username, site.SiteId);
                    if (personalized != null)
                    {
                        // redirect to the personalized page
                        NavigationManager.NavigateTo(Utilities.NavigateUrl(SiteState.Alias.Path, personalized.Path, ""), false);
                    }
                }
            }
```

PR #2984, opened 2023-07-09
Fix for #2983 Oqtane. Server selected in for build in configmanager

leigh-pointer:
Removed Oqtane. Server build from configuration manager debug and release.


sbwalker:
@leigh-pointer Can you please also submit a PR for the Theme template which includes this change

leigh-pointer:
@sbwalker done

PR #2990, opened 2023-07-10
fix migration error caused by new IsRead field in Notifications. Microsoft.Data.SqlClient.SqlException: 'ALTER TABLE only allows columns to be added that can contain nulls, or have a DEFAULT definition specified.'

sbwalker:


vnetonline:
Okay got it thanks for fixing it!

PR #3005, opened 2023-07-11
The Down path is incorrect 

leigh-pointer:
The Down Path drops column IsPublic this is the wrong column the correct column is IsRead.


vnetonline:
@leigh-pointer thanks for picking this up ... i will do a better job of testing everything and then submit a PR next time

sbwalker:
@leigh-pointer thank you for noticing this issue.

PR #3009, opened 2023-07-12
Using DynamicComponent Instead of RenderFragment in ContainerBuilder

ajahangard:


leigh-pointer:
Ok i see this, but why is this better?

sbwalker:
@leigh-pointer it is not any "better" but it does make the code a bit clearer, as it eliminates the more primitive component "builder" logic and replaces it with the DynamicComponent (which did not exist when Oqtane was originally created).

sbwalker:
@leigh-pointer I would like to retain some of the other "builder" logic however, as it is educational for developers to understand the primitive methods which are available to them

Issue #2958, opened 2023-06-29
Manage Settings bug


rcpacheco:
### Discussed in https://github.com/oqtane/oqtane.framework/discussions/2957

<div type='discussions-op-text'>

<sup>Originally posted by **rcpacheco** June 29, 2023</sup>
After activating the DNF.Project, per video 2, It activated properly and was able to create a project.
But when I tried to display the module's "Manage Settings" I got the following error:

![image](https://github.com/oqtane/oqtane.framework/assets/845725/3e803546-bf71-4d10-8fc5-56dd02e3ced3)

![image](https://github.com/oqtane/oqtane.framework/assets/845725/f9f381f6-1f62-4443-be50-0af2e2fb8d8c)
</div>

sbwalker:
@rcpacheco very strange.... I am not able to reproduce this error on https://www.dnfprojects.com. I will keep investigating. Does Manage Settings work for any of your other modules in your site (ie. Html/Text, etc...)

sbwalker:
Ok... it does not sound like a framework... more likely an issue in the DNF.Projects module itself. I will keep investigating.

rcpacheco:
I did yesterday the exercise in a Windows 10 machine with the latest dev version.
Today I tried with a linux machine and got some interesting errors.
In linux I used dotnet console commands to build the project at inside the DNF.Module.Project folder. I compiled succesfully:
![image](https://github.com/oqtane/oqtane.framework/assets/845725/855bed5b-c966-4615-a624-ca74e9fa4d80)

But then I tried to check if the projects files where properly copied to the debug folder and saw:

![image](https://github.com/oqtane/oqtane.framework/assets/845725/b270400c-b40b-44b7-9f84-70a59839b5ad)

I noticed that strange ''$'\r' file.

I deleted it and repeat and manually run debug.sh with same result. Did cat to the file and copied the instrucctions from the console and executed them succesfully and got the content copied to the server project.
Launched the app again and while trying to activate the module got an error.

The message was:

```
System.ArgumentNullException: Value cannot be null. (Parameter 'source')
   at System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
   at System.Linq.OrderedEnumerable`2..ctor(IEnumerable`1 source, Func`2 keySelector, IComparer`1 comparer, Boolean descending, OrderedEnumerable`1 parent)
[...]

PR #3019, opened 2023-07-13
fix bash script line endings and preserve using gitattributes (credit @rcpacheco in #2958)

sbwalker:


rcpacheco:
It works great, thanks a lot!
Now the scripts run as they should, also in linux.

Issue #3010, opened 2023-07-12
Issue with using OnSelect event with FileManager


vnetonline:
I am trying to use the FileManager component so that iI can select an Image (File) it should populate the Name Field

Here is the code of the FileManager Component 

```
<div class="col-sm-9">
    <FileManager FileId="@_slideFileId" Filter="@Constants.ImageFiles" 
        ShowFolders="false" ShowFiles="true" UploadMultiple="false"  
        Folder="Public" @ref="fileManager" OnSelect="PopulateFileName" />
</div>

  private async Task PopulateFileName(int fileId)
  {

      if (fileId != -1 && _slideFileId != fileId)
      {
          _slideFile = await FileService.GetFileAsync(fileId);
          _name = _slideFile.Name;
      }

  }

```
Here is what I see when I edit an item - image flickers because FileId becomes -1 after the PopulateFileName is called because as per Blazor it automatically calls `StateHasChanged()`

![screen2gif-filemanager-onselect-behaviour-edit](https://github.com/oqtane/oqtane.framework/assets/15242923/a46832d8-285d-4108-b40c-e3447fa8e16d)

Here is what I see when I add an item - no image preview is displayed but the FileName is populated

![screen2gif-filemanager-onselect-behaviour-add](https://github.com/oqtane/oqtane.framework/assets/15242923/e487d1d7-cf48-489c-941d-9fa4ec837385)

vnetonline:
@leigh-pointer can you verify for me if this is an issue?

vnetonline:
@sbwalker & @leigh-pointer 

I definitely think there is a problem with the OnSelect Event callback, from what I have read that EventCallback call `StateHasChanged` automatically which results in The Parent Component and its child component rerendering which results in the FileId becoming -1 and the FileChanged firing again causing a loop.

Have eaither of you dealth with this issue?

vnetonline:
The code in my razor looks like this 

```
private void PopulateFileName(int fileId)
{

    if (fileId != -1 && _slideFileId != fileId)
    {
        _slideFile = fileManager.GetFile();
        Name = _slideFile.Name;
        _slideFileId = fileId;
        StateHasChanged();
    }

}
```

![screen2gif-filemanager-onselect-behaviour-fixed](https://github.com/oqtane/oqtane.framework/assets/15242923/80d12183-86e7-4493-a4d0-be4899c75310)

vnetonline:
Refer to this thread 

[...]

PR #3023, opened 2023-07-14
[FIX] - Looping issue with OnSelect if EventCallback is used #3010

vnetonline:
Made OnSelect an Action instead of EventCallback which causes the lopping issue

### DIFF
--- a/Oqtane.Client/Modules/Controls/FileManager.razor
+++ b/Oqtane.Client/Modules/Controls/FileManager.razor
@@ -1,5 +1,5 @@
 @namespace Oqtane.Modules.Controls
-@using System.Threading
+@using System.Threading;
 @inherits ModuleControlBase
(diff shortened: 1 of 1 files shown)


sbwalker:
@vnetonline I do not believe this is the correct fix - please see my comments on #3010

sbwalker:
I already checked in a fix in #3024 earlier as I do not believe the OnSelect call is necessary in either OnInitialized() or OnParametersSet()

vnetonline:
This pull requests also moves SetImage() to before the event fires and removes the redundant SateHasChanged so that the parent page can get the image from the fileManager rather then making a call through FileService to retire the image again.

vnetonline:
Create another pull request just for moving the SetImage() before the event fires and removing the redundant StateHasChanged

PR #3026, opened 2023-07-15
[ENHANCE] - Moved `SetImage()` before the OnSelect fires so that the …

vnetonline:
…Parent can retrieve the Image from `fileManager` rather then calling `FileService` to retrieve again

### DIFF
--- a/Oqtane.Client/Modules/Controls/FileManager.razor
+++ b/Oqtane.Client/Modules/Controls/FileManager.razor
@@ -271,13 +271,10 @@
     {
         _message = string.Empty;
         FileId = int.Parse((string)e.Value);
-        if (FileId != -1)
-        {
-            await OnSelect.InvokeAsync(FileId);
-        }
-
         await SetImage();
+        await OnSelect.InvokeAsync(FileId);
         StateHasChanged();
+
     }
 
     private async Task SetImage()


sbwalker:
@vnetonline I do not understand the purpose of this change. When a user selects the ```<Select File>``` option in the list (ie. the first item in the list) it means that no file is currently. In this scenario, whatever image was selected previously should no longer be displayed - it should be hidden. The logic in this PR does not hide the image when a user selects the ```<Select File>``` option. Perhaps the real "fix" you are looking for is that OnSelect should also be moved outside the if condition so that the calling component knows that the user has not chosen a file?

vnetonline:
@sbwalker the purpose of this change is if a user *selects* the file from nothing <Select File> to a option in dropdown list then the parent should be able to retrieve the file from the FileManager rather than using FileService to retrieve it again. Currently the _file is not set when the the _file is set till after the event fires.

From the name of the EventCallback OnSelect means that a file has been selected?? 

So it doesn’t make however if it was it was OnChange then the image should be set after the event is triggered.

OnSelect is what a parent component will want so they don’t have to get the file again using FileService by injecting it into their component. They can do it by the FileManager.

sbwalker:
@vnetonline I understand what you are saying... basically SetImage() sets the _file object - which allows a developer to then use the GetFile() method to retrieve it. However this PR is incorrect. The problem is that the ```<Select File>``` option (FileId = -1) in the list of files is an actual option which a user can choose... it remove the previous selection.... so when they  choose this option the SetImage() needs to hide the image and _file should be set to null. So I believe the proper fix is below:

```
    private async Task FileChanged(ChangeEventArgs e)
    {
        _message = string.Empty;
        FileId = int.Parse((string)e.Value);
        await SetImage();
        await OnSelect.InvokeAsync(FileId);
        StateHasChanged();

    }
```

This fix will handle the  ```<Select File>``` option correctly AND will set the _file object prior to calling the OnSelect() event. Your parent component needs to check if GetFileId() returns -1 or GetFile() returns null - as these are valid outputs.

vnetonline:
@sbwalker i have made the change as per your sugestion and added a commit to this pull request

as you mentioed this push will handle the <Select File> option correctly AND will set the _file object prior to calling the OnSelect() event. The parent component needs to check if fileId returns -1 or GetFile() returns null - as these are valid outputs.

PR #3029, opened 2023-07-15
Load module definitions if State is not migrated

rcpacheco:
Relocate  get module definition inside "if serverstate is migrated".
If the state is not migrated the value of moduleDefinitions is not used.


rcpacheco:
@dotnet-policy-service agree

sbwalker:
@rcpacheco I cannot merge this PR for 2 reasons:

1. It does not describe the "issue" which the PR resolves not does it provide details on how to reproduce the issue
2. The entire file appears in the Diff, making it impossible to determine the logic which was changed.

rcpacheco:
You are right sorry, I will do it again.

PR #3032, opened 2023-07-16
Update moduleDefinitions only if not serverstate.IsMigrated

rcpacheco:
In the SiteRepository, while doing the Site initialization,
currently the site Themes and modules definitions are read before checking if the server has been migrated.
nevertheless, the loaded modules are only used if the server has not been migrated.

The process to load the modules definition involves a lot of operations, therefore, asumming that the GetModuleDefinitions routine has no side efects, it seems to me that it would be better to do it only we know that the server state is migrated. In that way the Site initialization will be faster in most cases.

### DIFF
--- a/Oqtane.Server/Repository/SiteRepository.cs
+++ b/Oqtane.Server/Repository/SiteRepository.cs
@@ -97,16 +97,18 @@ namespace Oqtane.Repository
 
         public Site InitializeSite(Alias alias)
         {
-            var site = GetSite(alias.SiteId);            
+            var site = GetSite(alias.SiteId);
 
-            // load themes and module definitions 
+            // load themes
             site.Themes = _themeRepository.GetThemes().ToList();
-            var moduleDefinitions = _moduleDefinitionRepository.GetModuleDefinitions(alias.SiteId);
 
             // site migrations
             var serverstate = _serverState.GetServerState(alias.SiteId);
             if (!serverstate.IsMigrated)
             {
+                // load module definitions
+                var moduleDefinitions = _moduleDefinitionRepository.GetModuleDefinitions(alias.SiteId);
+
                 // ensure migrations are only executed once
                 lock (_lock)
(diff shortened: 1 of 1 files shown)


sbwalker:
@rcpacheco thank you for re-submitting with an explanation and a way to see the code which was changed. Just to clarify, were you able to demonstrate that this is an actual problem so that I am able to reproduce it, or is it a hypothetical issue you identified by reviewing the code?

rcpacheco:
Hello, it Is not an issue because  the code does work and don't crashes. 

But after debugging the code several times I noticed that the whole process of initializing the variable moduleDefinitions was necessary only when the server state is migrated. 

Loading the modules definitions implies several actions and time that can be saved only by moving the variable definition inside the if, once we know for sure that it's value will be used.

But of course, it is your call 😉

Great work, by the way; a lot to learn from it. Thanks for sharing!!

sbwalker:
@rcpacheco the logic is a bit difficult to follow so I will try to summarize it below...

- InitializeSite is called from _Host.cshtml - the server-side page handler invoked for every page request
- InitializeSite calls GetThemes() and GetModuleDefinitions() - it does this because these methods set some ServerState properties which are specific to the current Site (since Oqtane is multi-tenant there can be multiple Sites). The most important ServerState properties are Scripts and Assemblies as these are used by other workloads which happen later in the processing life cycle
- Both GetThemes() and GetModuleDefinitions() use caching so that once they are executed once, subsequent calls to them will return immediately
- InitializeSite also deals with executing site migration logic which may have been specified in the IModule interface for modules. This logic should only be executed once for each Site in the installation (as migration logic is not idempotent) - which is why the IsMigrated property exists and there is a lock to prevent multiple executions.

In summary, the code change you suggested will have no impact on the run-time characteristics of the framework so I am inclined to follow the "if it ain't broke don't fix it" principle.

Issue #3041, opened 2023-07-18
Module Creator problem while using Postgresql


rcpacheco:
In the page "My Page" I created a Module with owner: Sapo (toad) and Module name: Verde (green).
It was created properly and activated, but after activation I got an error while trying to display it.
I tested both in linux and windows with sql local (only win) and sqlite (both) with no problem, but in postgresql (both) was crashing.

Knowing that postgresql uses snake case in its tables and columns (UseSnakeCaseNamingConvention), I reviewed the extraction of data and realized that the problem was while trying to load de data in the repo:

 return _db.Verde.Where(item => item.ModuleId == ModuleId);

![image](https://github.com/oqtane/oqtane.framework/assets/845725/75ad59fd-9c2c-47df-9211-c1c255e94590)

That lead me to realize that the name of the table was defined with the Table attribute:

![image](https://github.com/oqtane/oqtane.framework/assets/845725/f047542a-2a66-4ddf-9123-dcad1d1cd69f)

Once I changed the Table attribute to match the name in the database in snake case, it worked.

leigh-pointer:
@rcpacheco This is strange and the replace logic will not alter the case of the values entered in the Module Creator.

> I created a Module with owner: Sapo (toad) and Module name: Verde (green).

Then the Table attribute should be "SapoVerde"

rcpacheco:
Yes, It Is strange. If you have the time, please give It a try, leaving the attribute as you sugest. Look forward to hear your results. Thanks a lot.

leigh-pointer:
I have test with the current version of Oqtane and everything seems to be ok.

Step One
![image](https://github.com/oqtane/oqtane.framework/assets/9447612/68de4826-9283-4584-b68b-b18b7ef205c3)

Step Two
![image](https://github.com/oqtane/oqtane.framework/assets/9447612/4ed29b6c-5423-489b-b5c0-ee36cfb08879)

Step Three
![image](https://github.com/oqtane/oqtane.framework/assets/9447612/56df396e-415f-4d6e-aaef-07413bcccdf5)

Step Four
![image](https://github.com/oqtane/oqtane.framework/assets/9447612/540175fe-fae6-4c61-97ec-e74d25ab9705)

Step Five
![image](https://github.com/oqtane/oqtane.framework/assets/9447612/a6cfeb32-644e-4ded-8a89-fdd115acf1bb)

Step Six
[...]

PR #3047, opened 2023-07-19
Fixes Module Creator problem  #3041

leigh-pointer:
This fix is to handle a returned null from the database.  This issue came to light while using Postgresql

### DIFF
--- a/Oqtane.Server/wwwroot/Modules/Templates/External/Client/Services/[Module]Service.cs
+++ b/Oqtane.Server/wwwroot/Modules/Templates/External/Client/Services/[Module]Service.cs
@@ -18,7 +18,7 @@ namespace [Owner].Module.[Module].Services
         public async Task<List<Models.[Module]>> Get[Module]sAsync(int ModuleId)
         {
             List<Models.[Module]> [Module]s = await GetJsonAsync<List<Models.[Module]>>(CreateAuthorizationPolicyUrl($"{Apiurl}?moduleid={ModuleId}", EntityNames.Module, ModuleId));
-            return [Module]s.OrderBy(item => item.Name).ToList();
+            return [Module]s?.OrderBy(item => item.Name).ToList() ?? new();
         }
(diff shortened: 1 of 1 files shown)


sbwalker:
@leigh-pointer I have not merged this PR because I would like to understand why this is an issue with PostgreSQL but not with Sqlite or Sql Server. Considering that the whole point of using an ORM like EF Core is to abstract away the differences between various databases, it concerns me that PostgreSQL appears to be returning a different result from the Repository to the Controller - which is then retrieved by the Client (after serialization) using GetJsonAsync() - which throws an error. If the source of the problem was in GetJsonAsync() then the error should be occurring when running on all database types - not just PostgreSQL.

leigh-pointer:
@sbwalker I have just retested using SQL Server.  The error in the Client VerdeService class

```
        public async Task<List<Models.Verde>> GetVerdesAsync(int ModuleId)
        {
            List<Models.Verde> Verdes = await GetJsonAsync<List<Models.Verde>>(CreateAuthorizationPolicyUrl($"{Apiurl}?moduleid={ModuleId}", EntityNames.Module, ModuleId));
            return Verdes.OrderBy(item => item.Name).ToList();
        }
```

When you first build a module you have no data in your table (table hasent been built yet) but Index calls on Task OnInitializedAsync()
_Verdes = await VerdeService.GetVerdesAsync(ModuleState.ModuleId);


The offending code is 
 `return Verdes.OrderBy(item => item.Name).ToList();`
and trying to OrderBy on a null value.

sbwalker:
OK, so the issue is not related to PostgreSQL - that is very good to know. It would be nice if OrderBy contained internal logic to not throw an exception if the collection is null - but obviously that is not the case. So the fix you suggested is valid - basically validating the List is not null and if so, returning a default value. The default returned could be an empty list or null - just as long as the calling code knows what to expect so it can make wise decisions.

PR #3097, opened 2023-08-03
fix #3082 - handle username claim as "unique_name" with "name" as fallback, improve validation logic and logging

### LINKED ISSUE #3082: JWT - breaking change - username claim is now stored as "unique_name"
When attempting to use JWT authentication, queries were returning a server error 500.  Upon investigation, I determined that line 101 of UserSecurity.cs was throwing an error due to a null username being passed.
` identity.AddClaim(new Claim(ClaimTypes.Name, user.Username));`

The username passed to this is pulled from the JWT claims presented.  However, line 44 of JwtMiddleware.cs is trying to retrieve a claim of "name" for the stored username which results in the null value returned.  As can be seen in the screenshot attached, when Oqtane 4.x creates a JWT token, it now includes the Oqtane username as "unique_name" and not just "name".

If you are using a JWT token issued prior to upgrading to Oqtane 4.x, then it is still functioning as normal.  However, if you generate a new token that is using the "unique_name" claim, it now fails.

This appears to be a change in the JWT functions in .Net.  The previous version of System.IdentityModel.Tokens.Jwt used with Oqtane 3.x was version 6.10.0.  From my reading it looks like this version was using an incorrect claim of "name" and it is now changed in later versions of the package to use "unique_name".

Looks like it caused issues with some Azure stuff as well:  https://github.com/Azure/azure-signalr/issues/1606

image

sbwalker:


W6HBR:
@sbwalker I was able to verify that this is now working using both a 3.x generated token {name} as well as a 4.x generated token {unique_name}.

sbwalker:
@W6HBR thank you for identifying the problem, researching the solution, and verifying the changes

PR #3106, opened 2023-08-04
update Module and Theme Install UI to match Marketplace - including logos and support for sorting

sbwalker:


sbwalker:
image

sbwalker:
note that the logos are not handling transparent backgrounds when resizing - I need to research options in ImageSharp to determine how to resolve this

PR #3129, opened 2023-08-09
Fix missing translations

alikoli:
Fixed in this PR:
image
image
image
image
image
image


alikoli:
@dotnet-policy-service agree

PR #3167, opened 2023-08-20
Created List Control with all the OI icon names ref #3162

leigh-pointer:
Add the ability to to select an OI icon instead of needing the knowledge.  The control also allows for Text input so it is possible to select an icon from a loaded Theme resource.

#3162


sbwalker:
@leigh-pointer some feedback on this PR:

1. It includes an appsettings.json file - which should be removed
2. It would be better to create a more generic component rather than one which is so specific to Icons ie. an InputList component which allows a developer to pass in a parameter of Dictionary< string, string > DataList containing the list of allowable options. This component could then be used in a variety of different developer scenarios.
3. Assuming we introduce a more generic component, the Page Add/Edit would use this generic component and pass the parameter containing the list of icons. There is already a class in Oqtane.Shared\Shared\Icons.cs which defines the icons... and an additional Dictionary constant could be added for utilization in Page Add/Edit.
4. This PR does not currently support Localization. In order to separate concerns, the values passed to the generic component should be localized within the calling component so that it has the flexibility to either store the localizations locally within the module's RESX files or globally
5. In this case since the locaization would be used in multiple places and is related to global constants, the best place to add the keys would be to the SharedResources.resx file.

leigh-pointer:
@sbwalker i will take all feedback and rework.

PR #3169, opened 2023-08-22
Add InputList control to select values from a dictionary of string

leigh-pointer:
The control accepts a dictionary of string that can be searched using an input control.
Pages Add and Edit have implemented the control to render the list of Icons found in the Oqtane Shared namespace.


leigh-pointer:
@sbwalker This PR does not currently support Localization as it would be the Dictionary that needs to be localized.  The Icons class the Constant name is used as the Key and the string value the Value.  Keys should not be localised.

leigh-pointer:
@sbwalker ok, thats cool.

sbwalker:
@leigh-pointer I retained your reflection logic but moved the workload from the client to the server (#3178)

Issue #3163, opened 2023-08-20
Extra message that shouldn't be showing


Tony20221:
I added a module creator to a page then deleted it as I wanted to start fresh.
Clicking the gear icon still shows 'Module Added to Page'. Had to refresh to make it go away.

![image](https://github.com/oqtane/oqtane.framework/assets/99626433/92ae1916-0c1c-4c41-876b-0165712a8903)

vnetonline:
I will take a look at this and see what can be done

leigh-pointer:
In my opinion the message should appear on the page and not in the control center! When you click to you are looking at the page ! Just makes sense!

PR #3171, opened 2023-08-23
[FIX] Added the ability to clear the Message in control panel

vnetonline:
This fix is related to issue #3163


sbwalker:
@vnetonline this is clever... I did not think of implementing it on the cog icon, but this approach handles more use cases. On that note, is it even necessary to implement it on the close button?

vnetonline:
Not it is not necessary on the close button but I put it there as well just in case and there is no harm!

PR #3183, opened 2023-08-24
Update IconRescources

leigh-pointer:
Add all the IconResources as  Resource Key and Values


sbwalker:
I believe the Values should be formatted in a "friendly" manner with spaces between the words

leigh-pointer:
@sbwalker ok I will fix it.

leigh-pointer:
@sbwalker Fixed.

Issue #3184, opened 2023-08-24
Schedular Allows incorrect Time format


leigh-pointer:
I updated a scheduled job with an incorrect time format.  When the Item saves the framework crashes.

PR #3185, opened 2023-08-24
Fix for Schedular Allows incorrect Time format #3184

leigh-pointer:
Modified code to accept correct time types.


sbwalker:
@leigh-pointer the changes to the Utilities.cs methods are breaking changes which we would want to avoid as third party developers may be using these methods. It is preferable to create new overloads to avoid this.

leigh-pointer:
@sbwalker ok will change.

leigh-pointer:
This created two new methods.
`public static (DateTime? date, string time) UtcAsLocalDateAndTime(DateTime? dateTime, TimeZoneInfo timeZone = null)`

When we want to override a method with a different return type, we are entering into a territory where we need to be careful, as C# does not allow return type-based method overloading by itself. We can't simply change the return type of a method while keeping the same name and parameters.

However, we can use a different approach, which involves giving the methods distinct names while maintaining their parameter signatures and distinct behaviors.

PR #3205, opened 2023-08-28
filter deleted pages and modules on the server

sbwalker:


sbwalker:
@leigh-pointer I was investigating a strange issue with Script Resources and I realized that deleted pages and modules were being sent from the server-side SiteController API to the client-side SiteRouter and were being processed by the UI composition logic. And then the PageState.Pages and PageState.Modules collections were later being filtered using the IsDeleted property so that consumers could only manage active entities. 

This does not seem valid... as it is transferring more data than necessary between the server and client, and causing unnecessary processing of that data on the client.... only to later remove it before exposing it to consumers. Obviously I am responsible for this inefficiency... and it most likely happened when I did the performance optimization and aggregated multiple API method calls into a single consolidated GetSite() method. 

So I am really just raising this in case you recall that there was a valid reason why these deleted entities needed to be sent to the client... which I may have forgotten. So far my testing indicates that this PR causes no side effects.

PR #3230, opened 2023-09-07
Url Path Change NavigateTo Fix - Fixes #3229

### LINKED ISSUE #3229: Bug: Moving a page or renaming the "Url Path" results in 404 error page after saving in Admin/Page/Edit.razor.
When administrating a page which updates the Url Path the page results in a 404.

This issue happens both when changing the location of the page (parent/child) and when updating the `Url Path` in Modules/Admin/Pages/Edit/razor component while administrating the page from the page itself.

```
if (!string.IsNullOrEmpty(PageState.ReturnUrl))
{
    NavigationManager.NavigateTo(_page.Path);
}                    
else
{
    NavigationManager.NavigateTo(NavigateUrl());
}
```

This final update after refactoring and testing changes the If statement that checks for a `PageState.ReturnUrl` and if it has one after being saved it will use the `_page.Path` for the `NavigateTo()` function.

After full testing both in Admin Page Management and on pages redirecting to the new page path while editing a page from a page works as expected instead of resulting in a 404 error page.

thabaum:
Fixes #3229 

This PR is a solution that fixes the issue with navigation to the new page path URL when is it updated in location or when `Url Path` form field property is updated.

Added logic for when there is a `ReturnUrl` to check if the `currentPath` is the same as the `_page.Path` after being updated in the Admin/Pages/Edit.razor component.

If the _page.Path is different to use that instead of the `ReturnUrl` that has been passed.

The final commit simplifies it the most and just always sets the return url to the path.  So either solution works on my end.  I am going to leave it with the most simple version let you decide. 

Hope this helps!

### DIFF
--- a/Oqtane.Client/Modules/Admin/Pages/Edit.razor
+++ b/Oqtane.Client/Modules/Admin/Pages/Edit.razor
@@ -613,15 +613,16 @@
                     }
 
                     await logger.LogInformation("Page Saved {Page}", _page);
-                    if (currentPath != _page.Path)
-                    {
-                        string newPageUrl = NavigateUrl(_page.Path);
 
-                        NavigationManager.NavigateTo(newPageUrl);
-                        return;
-                    }
                     if (!string.IsNullOrEmpty(PageState.ReturnUrl))
-                    {
+                    {                    
+                    if (currentPath != _page.Path)
+                        {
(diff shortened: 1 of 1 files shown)


sbwalker:
resolved in #3238

PR #3233, opened 2023-09-07
Head.razor Title & HeadContent indentation + new lines - Fixes #3232

thabaum:
Fixes #3232

Creates indentation and new lines for `Title` and `HeadContent` rendered in the HTML head source. (part 1)

These linked PR's to issue #3232 will help generate the following:

image

Notice the </script> tag still there as the scripts I put into the Head Content of the Page Content leave this behind when scripts are removed from `headcontent`

also the `app-stylesheet-page-xxxx` entries after the first one do not have indentation.

These issues remain and will be dealt with when I (or whomever) can find a solution. I did manage to get everything on it's own new lines.

Noticed this while adding resources to a module I created thought this would be a nice touch for Oqtanes source page head section readability.

PR #3234, opened 2023-09-07
ThemeBuilder.cs Head indentations and new lines for HeadContent and FavIcon link elements - Fixes #3232 p2

### LINKED ISSUE #3232: Issue: Head Title and Head Content rendering indents and line
The head section resources for `HeadContent` and `Title` elements can use some formatting for indents and line breaks when rendered as source.

I have all but the theme style sheets indenting with new lines as needed in PRs I will link with one by editing the `Head.razor` file and the other is editing the `ThemeBuilder.razor` file I will submit a PR for each to be reviewed.

I was unable to quickly get a solution for the theme.css files to load with indents.  The second+ are not indented, however the first is and the last element might be leaving an extra new line if it is not coming from somewhere else.

Not a big deal since most all apps (like github) you try to read the head source rendered are far worse than Oqtane's current rendering of these head elements.  However for developing and testing out new modules and themes resources loading it is nice to clearly see things load.

thabaum:
Fixes #3232 

This PR along with #3233 will add line breaks and indentation to `HeadContent` and  `FavIcon` HTML head elements making them easier to review page source while creating themes and modules.

### DIFF
--- a/Oqtane.Client/UI/ThemeBuilder.razor
+++ b/Oqtane.Client/UI/ThemeBuilder.razor
@@ -40,7 +40,7 @@
             favicon = Utilities.FileUrl(PageState.Alias, PageState.Site.FaviconFileId.Value);
             favicontype = favicon.Substring(favicon.LastIndexOf(".") + 1);
         }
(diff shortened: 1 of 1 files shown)


sbwalker:
see comments in #3232

PR #3275, opened 2023-09-20
Update to Profiles field list to include Title, Category and ViewOrder

W6HBR:
When adding additional profile fields, this enhancement makes it easier to see details of the profile fields for the purpose of field organization.


W6HBR:
Sample Image of Profiles screen with this enhancement.
image

sbwalker:
This was missing proper localization support - added in #3285

PR #3293, opened 2023-09-22
Add Email column to User Manager screen

W6HBR:
The logic for including this additional column is that the search functionality allows you to search on Username, DisplayName and Email.  However, only the Username and DisplayName columns are displayed, so you don't have direct confirmation of the email search match.  Additionally, looking up emails is one of the most used reasons that admins go to the User Manager screen.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Users/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Users/Index.razor
@@ -37,7 +37,8 @@ else
 					<th style="width: 1px;">&nbsp;</th>
 					<th style="width: 1px;">&nbsp;</th>
                     <th class="app-sort-th link-primary text-decoration-underline" @onclick="@(() => SortTable("Username"))">@Localizer["Username"]<i class="@(SetSortIcon("Username"))"></i></th>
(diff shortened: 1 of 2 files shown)


leigh-pointer:
@W6HBR the name column is a href link to the email address.  Entering part of an email address returns valid search results.

leigh-pointer:
@sbwalker i had already implemented this sometime ago and the PR was rejected because of the number of columns needed to be considered for MAUI.

W6HBR:
I added the HREF to the email address as well so both Name and Email are clickable.  My need here was for display purposes to see the results with Email in addition to the Name.  

When a user says they can't login and can't get a reset link, the email is always the first thing to check.  As it is, for an Admin to see the email, they either have to hover over the name to see the link, click on the name to create a message (non-intuitive) or click edit to see the user record.  Having the email displayed from the search/list results presents the cleanest and most intuitive result.

sbwalker:
@leigh-pointer yes, you are correct that I rejected your PR in the past for this... and it wasn't specifically for MAUI - it was more related to trying to ensure the admin UI is still functional on a smaller screen size.  One way to deal to with this is minimize the number of columns. Another way is to introduce horizontal scrolling of the grid component in cases where the screen screen is reduced (the Sql Management UI takes this approach)

PR #3310, opened 2023-09-24
upated LanguageSwitcher with end padding

leigh-pointer:
When the LanguageSwitcher is dsiplayed the wsa no end padding on the button so it appeared flush with the following button.  Adding "pe-1" to the class corrects the issue

### DIFF
--- a/Oqtane.Client/Themes/Controls/Theme/LanguageSwitcher.razor
+++ b/Oqtane.Client/Themes/Controls/Theme/LanguageSwitcher.razor
@@ -8,7 +8,7 @@
 
 @if (_supportedCultures?.Count() > 1)
 {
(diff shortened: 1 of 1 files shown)


hishamco:
Please screenshot how the change looks like before & after?

leigh-pointer:
Before!
image

After!
image

hishamco:
Please double-check in RTL.

leigh-pointer:
It will not effect anything to do with RTL because it is button padding and nothing to do with text.

hishamco:
From the screenshot, I see there is right padding has been added

hishamco:
I might need to do that while I am an Arabic guy

PR #3326, opened 2023-09-26
Empty _themetype check to avoid page crash

ijaz-saeed:
without this check, admin dashboard -> Page Management -> Edit page is crashing

image


sbwalker:
This was caused b y #3308

sbwalker:
Unfortunately this is going to require a 4.0.5 patch release as it affects some critical UI functionality. I guess that will make 4.0.4 officially a "not found" release ;)

PR #3342, opened 2023-09-27
style page names in Admin Dashboard so that they are always visible in Bootstrap themes

sbwalker:


sbwalker:
To prevent renderings such as below (using Bootswatch Vapor)

image

Issue #3325, opened 2023-09-26
The External Login Provider Did Not Provide A Valid Email Address For Your Account. Please Contact Your Administrator For Further Instructions.


Rodien:
I am trying to use the 'External Login Settings,' and everything works except when I get redirected back to oqtane. Upon redirection, I receive the following message: 'The External Login Provider Did Not Provide A Valid Email Address For Your Account. Please Contact Your Administrator For Further Instructions.

Calling the userinfo endpoint via Postman returns the following values: (userinfo-endpoint)

`{
    "name": "admin",
    "sub": "4rt0pmemefh6k1qpmvx7pbftwz",
    "email": "r.bas@dummy.nl",
    "email_verified": true,
    "roles": [
        "Administrator",
        "F M"
    ]
}`

That's a valid email, so I don't get why it's failing. Or is there maybe something I am forgetting in the admin area?

![image](https://github.com/oqtane/oqtane.framework/assets/50739829/be699d15-de48-42f5-b00d-fad80844bc11)

sbwalker:
@Rodien please look in your Event Log - there should be more detailed information related to why the login is failing.

Rodien:
Error detail: 
`Provider Did Not Return An Email Address To Uniquely Identify The User. The Email Claim Specified Was email And Actual Claim Types Are oc:entyp, sub, oi_au_id, azp, nonce, at_hash, oi_tkn_id, aud, exp, iss, iat. Login Denied.`

Here is a dump of the token values. 

`{
  "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier": "4rt0pmemefh6k1qpmvx7pbftwz",
  "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name": "admin",
  "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress": "r.bas@dummy.nl",
  "http://schemas.microsoft.com/ws/2008/06/identity/claims/role": [
    "Administrator",
    "Financieel Medewerker"
  ],
  "Permission": [
    "ManageSettings",
    "AccessAdminPanel",
    "ManageAdminSettings",
    "PublishContent",
    "EditContent",
    "DeleteContent",
    "PreviewContent",
    "CloneContent",
    "AccessContentApi",
    "ListContent",
    "EditContentOwner",
    "ApiViewContent",
    "ViewContentTypes",
    "EditContentTypes",
    "ExecuteGraphQLMutations",
[...]

PR #3352, opened 2023-10-03
Introduce a dropdown menu for authorization response types

Rodien:
Introduce a dropdown menu in the 'External Login' settings area for authentication flow response types. 

I used the response types found in the Microsoft.IdentityModel.Protocols.OpenIdConnect.OpenIdConnectResponseType class.

For more information take a look at: #3325

### DIFF
--- a/Oqtane.Client/Modules/Admin/Users/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Users/Index.razor
@@ -262,7 +262,22 @@ else
 									<input id="parameters" class="form-control" @bind="@_parameters" />
 								</div>
 							</div>
-							<div class="row mb-1 align-items-center">
+                            <div class="row mb-1 align-items-center">
+                                <Label Class="col-sm-3" For="authresponsetype" HelpText="Specify the authorization response type" ResourceKey="AuthResponseType">Authorization Response Type</Label>
+                                <div class="col-sm-9">
+                                    <select id="authresponsetype" class="form-select" @bind="@_authresponsetype" required>
+                                        <option value="code">@Localizer["AuthFlow.Code"]</option>
+                                        <option value="code id_token">@Localizer["AuthFlow.CodeIdToken"]</option>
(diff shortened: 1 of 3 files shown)


sbwalker:
@Rodien the majority of these response types are deprecated in OAuth 2.1 because they have security vulnerabilities. This is the reason why Oqtane only included support for Authorization Code Flow (as it is the only recommended approach in OAuth 2.1). So I guess the question is whether or not Oqtane should follow best practices or support legacy implementations. The problem is that if an Oqtane installation gets hacked and its because someone was using a deprecated flow, Oqtane's reputation will still be affected. This is a bit of a moral question - should Oqtane prevent users from exposing themselves to security vulnerabilities?

Rodien:
@sbwalker You are correct; I was thinking about this today. However, Oqtane should at least support the hybrid flow for now.

The OpenID Connect specification is still based on OAuth 2.0. Therefore, some Identity Providers (IdPs) are still using the 'code id_token' approach.

Another solution would be to make some changes to the token validation inside of Oqtane. Currently, Oqtane expects an email claim. If there is no email claim, Oqtane will throw an error. But what if I want to use something else as a unique identifier?

I mean, Oqtane just needs some unique value to register a user; it doesn't have to be an email address.

Is there a reason why Oqtane specifically expects an email claim?

For example, if I use the Authorization code flow, I will get a token back with the sub claim inside, which has a UUID. Is there a reason why I should not use that value to register a user?

I have been looking around, but I have not found anything yet that says it should not be done that way.

sbwalker:
Every external user login must be associated to a local user account, as everything in Oqtane is related to a UserId. Oqtane uses .NET Identity and it requires that a user must have an email address. Therefore an email claim is required for integration.

Rodien:
Got it

Rodien:
I would like to note that I only tested the 'Code Id_Token' flow and 'code' flow.

I can make a pull request and remove the rest of the flows, or you can do this yourself if you prefer.

Rodien:
Yes, Correct!

PR #3364, opened 2023-10-10
Pager Search Localization

leigh-pointer:
Localized the new Search Buttons
Added a Localized Placeholder that will display what columns are being searched.

image


sbwalker:
@leigh-pointer sorry, I created a conflict by merging #3365 - note that the general words Search and Reset are already defined in SharedResources. Can you please submit another PR for the placeholder enhancement

leigh-pointer:
Ok will do it now

PR #3370, opened 2023-10-11
Language Selector Alignment fix  #3368

### LINKED ISSUE #3368: Language Selector menu alignment.
The menu should be right aligned for readability. 

@sbwalker  I have done some work for a PR but due to the fact that the **LanguageSwitcher** component is embedded in the **Control** component, A **Parameter** needs to passed down.  It works ok, and the default would be left as is, but Oqtane theme would need to change?
 
image

image

leigh-pointer:
Themes fixed also


hishamco:
I might have a better idea, I can propose another PR if you & @sbwalker don't mind

PR #3372, opened 2023-10-12
optimize Pager search to remove redundant AllowSearch parameter and avoid consuming memory when not in use

sbwalker:


sbwalker:
@leigh-pointer I just noticed that the sort feature which was added to User Management does not work with the new Pager search capability. This is because the search is implemented within the Pager component and sorting is implemented in the User Management module component. There are a few options:

1. Remove the search capability from the User Management module
2. Remove the sort capability from the User Management module
3. Implement a custom search in the User Management module which emulates the search capability in the Pager component but also works with sorting

I am leaning towards 2. or 3. as I think the search is quite valuable in User Management - especially once you have a lot of users. Personally I am not sure if sorting is all that useful when you have a search capability. The only column where sorting might be useful is LastLogin - ie. to quickly see a list of the users which have logged in recently. I suppose having both search and sort in User Management provides the greatest functionality for admins.

Your thoughts?

leigh-pointer:
@sbwalker just started looking at this and discovered an issue with search, will look into it as well.
SearchIssue

sbwalker:
Good catch @leigh-pointer ... the problem is that UserManagement loads a collection of UserRole objects... and those UserRole objects contain related objects for User and Role - which is where the SearchProperties are located ie. it should be "User.Email" rather than "Email". However Type.GetProperty() does not handle related objects - so this will need some more work.

sbwalker:
@leigh-pointer #3377 resolves the issue you identified above. I added a recursive method so that it can resolve nested objects by using a dot notation ie. "User.Email" will search the Email property on the User object. I also modified the Placeholder so that it hides the complexity of nested objects from the end user.

sbwalker:
@leigh-pointer I was able to resolve the issue with making Pager search work with sorting - it was actually very simple. I simply needed to re-filter the item list during OnParametersSet if search criteria had been entered previously.

PR #3375, opened 2023-10-13
Module Creator allow punctuation in Description field

leigh-pointer:
Modified the Regex to all punctuation in the description. "^[A-Za-z0-9 .,!?]+$")


sbwalker:
You are certain that the additional punctuation characters are valid in an XML document?

leigh-pointer:
@sbwalker 
> You are certain that the additional punctuation characters are valid in an XML document?

Yes, I was creating a new module this morning and could not punctuate my description.  So it has be tested right through to module build.

Issue #3394, opened 2023-10-18
Issue:  Site Management - Green Checkmark, but error notificiation warning - Unable to see what field is errored easily.


thabaum:
![image](https://github.com/oqtane/oqtane.framework/assets/13577556/8bbb83f0-c4c3-4840-afae-3eb432ab5fb8)

This happens on I believe all of them, the green check mark appears like you did it successfully (dropdowns at least).

This should be a red X or Red Highlighted Border around the option not set properly to help identify the issue.  Green checkbox makes you feel like you did it right.

Also instead of "Select" it could be set to the actual "Default" setting for these inputs and would resolve most these issues instead of instructions to select.

leigh-pointer:
This guessing this is caused by the Dropdowns have a default value of "-" which would need to be space.empty.

sbwalker:
I agree that this UI should set the Site Template to the Default Site Template.

thabaum:
> I agree that this UI should set the Site Template to the Default Site Template.

It would be nice to be able to set the "Default" Theme/Container and Site Template in another area so that these are the defaults used when creating a site.  This way they are always set to your set "Site Defaults" when creating a new site to speed this process up.

There are two default themes... there should be only one and it should be selected.

![image](https://github.com/oqtane/oqtane.framework/assets/13577556/5d7f3a4d-782f-430d-8b6f-dd383bd1f5cd)

I will create another issue so this can be closed with fix #3398

sbwalker:
the Oqtane Theme is considered the default theme - the Blazor Theme is an alternate theme which is included only for testing purposes.

sbwalker:
Closing this issue - but another general issue could be logged for "Add Form Validation Support For Unselected Dropdowns"

PR #3398, opened 2023-10-18
Fix for #3394 Add Site validation

leigh-pointer:
Add Site validation now shows required values if not entered.

image


sbwalker:
@leigh-pointer Oqtane uses "-" in most select controls because Blazor does not recognize empty string as a valid value. For example if you have a default option with a value of "" and you are binding to a variable, the variable will not contain "" - it will contain null. It also may not set the default option correctly. Perhaps later versions of Blazor have addressed these problems but some additional investigation will be needed to verify.

sbwalker:
@leigh-pointer I am going to close this PR now that there is a .NET 8 branch, as I only want to maintain a single development branch at a time

PR #3403, opened 2023-10-18
Remove default connection string data

thabaum:


thabaum:
crap i think this is wrong branch, but anyhow...you can do this in a quick commit

PR #3406, opened 2023-10-18
Use IHeaderDictionary.Append

thabaum:
`Use IHeaderDictionary.Append or the indexer to append or set headers. IDictionary.Add will throw an ArgumentException when attempting to add a duplicate key.	`

This clears up this warning message.


sbwalker:
sorry I was resolving the compiler warning at the same time

PR #3409, opened 2023-10-18
development module theme templates

thabaum:
Updates module and theme creator templates to use latest .NET 8 Runtime and Package Dependencies.

Module and Theme Creator Work Great!


thabaum:
@sbwalker tested building and using both modules and themes, work great with this PR update.  Updating to Bootstrap version 5.3.2 causes an issue with dropdown list for page editing accessing module options to work with modules on pages so reverted back to Bootstrap 5.3.0.

PR #3419, opened 2023-10-20
UserProfile/View.razor: Sets Input Classes to col-sm-9 - Fixes #3418

### LINKED ISSUE #3418: Issue: UserProfile/View.razor - View Notification Dialog Title + Subject Input Elements Are Not Full Width
Subjects are cut off if they are long such as log notifications due to shortened input elements.

image

Also in the "View" section if you leave and come back to it you may forget who are you "replying" to or who sent the message or who you sent the message to you are viewing.

Also "View Notification" title should change to "Notification Reply" when the button is clicked to reply to let a user know they are now performing this action.  So the reply button should override the title of the dialog.

Maybe another issue but I have a fix for this one I am about to submit as well.

thabaum:
Fixes #3418

Results:

image

### DIFF
--- a/Oqtane.Client/Modules/Admin/UserProfile/View.razor
+++ b/Oqtane.Client/Modules/Admin/UserProfile/View.razor
@@ -13,13 +13,13 @@
             <label Class="col-sm-3">@Localizer["Title"] </label>
             @if (title == "From")
             {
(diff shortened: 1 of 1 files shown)


thabaum:
#3424 targets net8 branch targets

PR #3423, opened 2023-10-21
Models/Profile.cs: Sets Rows Default Value Equal to 1 - Fixes #3412

### LINKED ISSUE #3412: Issue: User Profile > Profile Tab - Input Fields are all TextArea Fields
As you can see below all the input fields show up as textarea fields.

image

thabaum:
Fixes #3412 

This fix works on both clean installs, and on new sites created.
image

Still I was trying to find the logic that creates a site and sets to 0 or does not set the value. For example if I wanted to create a site template with different profile configuration settings.

I would think this is where the "fix" is but maybe there is a better solution.

### DIFF
--- a/Oqtane.Shared/Models/Profile.cs
+++ b/Oqtane.Shared/Models/Profile.cs
@@ -77,6 +77,6 @@ namespace Oqtane.Models
         /// <summary>
         /// Optional number of rows (textarea)
         /// </summary>
-        public int Rows { get; set; }
+        public int Rows { get; set; } = 1;
     }
 }


sbwalker:
The proper fix for this should be in the Oqtane.Server\Repository\SiteRepository.cs class in the CreateSite() method... the calls to _profileRepository.AddProfile() need to set Rows = 1. This will initialize the values properly on all new installations.

Also the UserProfile should validate that Rows contains a value of 1 or greater - all other input values are invalid.

thabaum:
Gotcha, I see the code now,  I was all through the code and even opened this file but overlooked that method as it was a brief look, the only file I didn't 100% read of course.  This is great I really wanted to review this logic.

If you have not done so already I can create the PR as follows:

Are there any other settings that need a default value you see not getting set by chance I thought I would ask while we are working in this area.

image

I will update this PR with all the necessary profile entries as follows and test on .nET 8 branch:

```
_profileRepository.AddProfile(new Profile
    {SiteId = site.SiteId, Name = "Phone", Title = "Phone Number", Description = "Phone Number", Category = "Contact", ViewOrder = 8, MaxLength = 50, DefaultValue = "", IsRequired = false, IsPrivate = false, Options = "", Rows = 1 });
```

And I can also add that validation check on the profile management form for the "Rows" column.

One other note is I can add the "Biography" to this.  Also one other issue I am not sure how we can handle is localization of these profile properties.  When they are added/removed from a site a resource file will need updated as well.  This can be after .NET 8 release, but something I was thinking should get included by someone at some point.  Nothing I really need so not a big deal to me.

I can add and test a biography with 5 rows default for new sites created if OK.  I don't think we want to add a biography as a migration however as this can be something they add on their own to existing sites if admin needs.

thabaum:
@sbwalker it looks like there are a lot of fields that generate errors if you enter letters instead of numbers for example.

I am going to provide this fix for the "Rows" issue first.  Then I will create another issue regarding Profile Validation as those green check marks need to work with it instead of throwing a system error:  This can be worked on in a separate PR as it works on a different file and deals with a different issue.

image

I will include the "Biography" profile setting as shown in edited discussion above for new sites created.  I will update this pull request to reflect these changes for now and make any adjustments once you review.

sbwalker:
@thabaum Oqtane does not include default values in its models. models are intended to be POCO - with no embedded behaviors.

Please do not include a biography profile item. This is not a typical profile field for most applications.

thabaum:
@sbwalker  no problem, removed.  I am going to add a talking point to the issue discussion as a question prior to closing.

Issue #3426, opened 2023-10-21
Issue: Profiles/Edit.razor: Form Validation For Rows, Length and Order Input Fields


thabaum:
Rows needs to allow numbers from 1 or greater for input validation.
If a value less than 1 is entered into rows or letters are entered in any other fields that should only accept numbers where numbers in the Edit.razor form the green check mark should be a red exclamation mark to show those inputs need set properly with valid input expected.

Reproduce the error placing characters where numbers are expected.

![image](https://github.com/oqtane/oqtane.framework/assets/13577556/9f5bf02e-cae2-4cf0-ad88-0b6faeeccfb9)

thabaum:
@sbwalker I have fixed the validation for the rows,

The validation of Length and Order are next while I am here.

I understand Length is to be 0 or higher (Numbers only)

But for "Order" I want to be sure you will allow 0 which seems to be default or should this be 1 or greater?

sbwalker:
@thabaum In general I would recommend fixing the specific issue and not adding other enhancements

thabaum:
@ I kept it simple in the new PR using the ` type="number"` attribute in the input elements of the form.

thabaum:
closed with #3347

PR #3430, opened 2023-10-22
Admin\Profile\Edit.razor: Adds Form Validation For "Rows", "Length" and "Order" - Adds Missing "Rows" ResourceKey to Related Resource File.

thabaum:
Fixes Issue #3431 and Fixes Issue #3426

This PR if accepted will do the following for Admin Profile Management:

- Adds validation logic to Rows, Length and Order data input fields ` type="number"`.
- Adds localization resource file data for the "Rows: " label.
- Adjusts "Validations:" resource data to include a space after the colon to read uniform to other fields as "Validations: "
- Formats document using VS and then fixes it's formatting mistakes.

Could not of made this any harder on myself but it a was a great one to learn from!  For admin component consistency I re-added the @ to `@bind` directive variables.

So many commits, I may create a fresh PR.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Profiles/Edit.razor
+++ b/Oqtane.Client/Modules/Admin/Profiles/Edit.razor
@@ -164,7 +164,6 @@
         var interop = new Interop(JSRuntime);
 
         if (await interop.FormValid(form))
-
         {
             try
             {
--- a/Oqtane.Client/Modules/Admin/Profiles/Edit.razor
+++ b/Oqtane.Client/Modules/Admin/Profiles/Edit.razor
@@ -224,6 +224,7 @@
         else
         {
             AddModuleMessage(SharedLocalizer["Message.InfoRequired"], MessageType.Warning);
+            // Validation field warning messages
             if (!isNameValid) AddModuleMessage(Localizer["Error.Profile.Name"], MessageType.Warning);
             if (!isTitleValid) AddModuleMessage(Localizer["Error.Profile.Title"], MessageType.Warning);
             if (!isDescriptionValid) AddModuleMessage(Localizer["Error.Profile.Description"], MessageType.Warning);
--- a/Oqtane.Client/Modules/Admin/Profiles/Edit.razor
+++ b/Oqtane.Client/Modules/Admin/Profiles/Edit.razor
@@ -34,19 +34,19 @@
         <div class="row mb-1 align-items-center">
             <Label Class="col-sm-3" For="order" HelpText="The index order of where this profile item should be displayed" ResourceKey="Order">Order: </Label>
             <div class="col-sm-9">
-                <input id="order" class="form-control" @bind="_vieworder" maxlength="4" required />
+                <input id="order" class="form-control" @bind="_vieworder" maxlength="4" type="number" required />
             </div>
         </div>
         <div class="row mb-1 align-items-center">
             <Label Class="col-sm-3" For="length" HelpText="The max number of characters this profile item should accept (enter zero for unlimited)" ResourceKey="Length">Length: </Label>
             <div class="col-sm-9">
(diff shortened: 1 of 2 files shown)


sbwalker:
A quick review of this PR indicates that it will require a lot of refactoring. I will not have time to provide feedback until Monday.

thabaum:
@sbwalker No problem, I am just getting the kinks worked out of the PR.  VS was really giving me some issues.  I downloaded this branch from my repo and things look OK after firing up visual studio.

Most blazor docs use `<EditForm>` and data annotations so I was trying to do my best with something to start.  You are more than welcome to edit it as you please or let me know I can work things out as prescribed.

I look forward to feedback as I know there are a number of ways to do the same thing.  I had a few versions, I wanted to see how this one would work out.  I had to cleanup some of the leftovers from other attemtps I had so I think that may have played a part in some of the issues I had, but Visual Studio was messed up for sure as I could not copy paste things without things being bugged out with missing divs and formatting problems.

After shutting things down and restarting the project seems to be working normal again...

I am done for the weekend myself.

sbwalker:
@thabaum I have included my feedback below...

1. Localization was already implemented correctly in this module. If you review any other module in the framework you will see that the Label component is responsible for localization. It uses the ResourceKey in conjunction with attribute names to perform localization ie. Name.Text is the RESX key for the label text and Name.HelpText is the RESX for the label help text. So although the changes you made may still function correctly, they are not necessary, result in extra processing, and are not the standard way to implement localization in Oqtane.

2. Changing the maxlength to 300 for Validation is not a valid solution. This property is backed by a database field which has a maxmimum length of 200 characters. So if you try to submit a value greater than 200 characters it will result in a run-time error. Increasing the size of most fields requires a corresponding database migration (and testing on each supported database engine).

3. If you look at other data entry UIs in Oqtane, you will notice that they use the standard HTML5 form validation for required fields. This provides visual feedback to the user if they try to submit a form which is missing required elements. So there is no need for logic in the razor component to check if the Name, Title, Description is non-empty. These should be removed.

4. Although it may be tempting to include logic for future scenarios ie. isCategoryValid, etc... it is code that serves no current purpose and will need to be maintained in the long term. As a result it should not be included - gold-plating should be avoided.

5. Based on 3 & 4 above, the RESX file should have a number of unnecessary items removed. Every item represents a translation that a third party needs to include and maintain in their language-based localization packages, so Oqtane should only include new items that are absolutely necessary.

thabaum:
@sbwalker Looks like a few foul balls and a fast pitch down the middle caught looking at.

1. Remove the localization as it is done twice in the Label from the "ResourceKey" feature I need to review further to understand how Oqtane handles this fully for future reference.  Thanks for the clarity I thought I was missing something here.  I will review the Oqtane Label component code further.
2.  Database field needs updated for the regex validation set to 300 characters which makes sense for the error I received, I will have to review that but sounds like more testing than I can provide with the different database engines at the moment. I will remove it from this PR as well and keep that another for future that is not needed right now by me so not an important issue.
3. Remove the added validation logic for areas checked by standard HTML5 validation.
4. Remove added logic for areas outside of the scope of "Rows, Length and Order" validation
5. Remove all the new language resources that are not absolutely necessary.

Thanks for explaining these to me with feedback as it will help in the future.

I should of put this as a "draft" since I was adding more to it that I knew would not all get accepted, but put there in the event some other use cases are covered prior to the accepted solution.

I will thin this PR out and wait for final review as I may leave a bit more than is needed in the resource file area for the specific error messages, but I will most likely remove all of these from taking in your response as well in the end.  It would be nice to have the error message state which field is invalid.  But not sure if that fits as "absolutely necessary".

thabaum:
@sbwalker ready for a final review

thabaum:
closing this as too many commits to review...

PR #3440, opened 2023-10-23
UserProfile: Adds Notifications Tab Return Url - Simplifies @bind directive

### LINKED ISSUE #3417: Issue: UserProfile/Index.razor - Notification tab not focused after viewing/canceling or sending messages.
After you send or view a notification a user will no longer be focused on the notifications tab which is the area the would expect to be after clicking send or view-cancel (close).

thabaum:
Fixes #3417 
- Adds Notifications Tab To The Return Urls in the `UserProfile` Admin Component Notifications Tab Files Related
- Simplifies @bind Directive Used In `UserProfile` Component Files.


sbwalker:
@thabaum I am going to close this PR as #3443 resolves #3417. You may want to take a look at the details of this PR - especially Index.razor as it demonstrates how to pass a ReturnUrl parameter in ActionLink which can then be used in other module components via PageState.ReturnUrl - and it is also used when closing the modal dialog.

PR #3443, opened 2023-10-24
fixes to notifications UI in user profile

sbwalker:
- fixed missing localization in View.razor (includes labels and all other static text)
- fixed UI navigation back to Index.razor to include tab=Navigation querystring parameter
- fixed IsRead property update in View.razor to only occur for messages received
- fixed UpdateNotificationAsync API to update based on FromUserId or ToUserId
- included username AutoComplete in Add.razor


thabaum:
great PR the autocomplete feature is another thing I had noticed making it hard to find a user.  I will test this out.  Thanks!

PR #3462, opened 2023-11-09
Added login cookie expiration time functionality.

PfaffIC:
Added functinality to always remember user login. Provides the option to force login cookie expiration and dont fallback on the session timespan that is used as default when users dont choose the option 'Remember me'.
Added login cookie expiration time. Added setting in user settings to declare custom cookie expiration time. Cookie expiration time overwrites default expiration time of 14 days (if not session timespan is used).
These settings are optional.

Background info: We use oqtane as a framework that hosts our angular applications. The angular applications use a login system with login cookie expiration of 8 hours. Oqtane uses cookie session timespan. So when the login cookies in our application run out oqtane might still be logged in, which causes incompatibility between our applications and oqtane. We need the opportunity to declare the cookie expiration in oqtane and also tell it to force use this expiration time (the 'alwaysremember' functionality should ensure this, even for an oauth login).

### DIFF
--- a/Oqtane.Client/Modules/Admin/Login/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Login/Index.razor
@@ -36,10 +36,13 @@
 						</div>
 					</div>
 					<div class="form-group mt-2">
+						@if (!_alwaysremember)
+						{
 						<div class="form-check">
 							<input id="remember" type="checkbox" class="form-check-input" @bind="@_remember" />
 							<Label Class="control-label" For="remember" HelpText="Specify if you would like to be signed back in automatically the next time you visit this site" ResourceKey="Remember">Remember Me?</Label>
 						</div>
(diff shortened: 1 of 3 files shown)


PfaffIC:
@dotnet-policy-service agree

PfaffIC:
@sbwalker recreated the pull request that had trouble with CLA. Think it worked in this one. I added verification to my commits and reassured my primary mail in github and the mail for my commit were matching. Hope that fixed it :)

sbwalker:
@PfaffIC does this PR also handle external login using an external IDP such as OIDC?

PfaffIC:
@sbwalker added pull request #3471: Added cookie expiration for oauth2 external login. Unfortunately i'm not really familiar with OIDC and dont have it set up locally. So can't provide that right now.

PR #3470, opened 2023-11-12
[ENAHNCE] - Added TextBox to Action Dialog

vnetonline:
I am creating an Amazing.Module.CodeGen to generate code for Oqtane. I need a Dialog which captures input. 

I have enhanced the `ActionDialog` to include a TextBox as follows 

1. Introduced the following parameters

```csharp
    [Parameter]
    public bool ShowTextBox { get; set; } = false; // optional - specifies if the TextBox should be displayed in the dialog - default is false

    [Parameter]
    public string TextBoxLabel { get; set; } // optional - specifies if the TextBoxLabel for the text box
```

2. Usage is as follows 

```razor
<ActionDialog ShowTextBox="true" TextBoxLabel="Solution Path" 
          OnClose="GetInput" Header="Add an existing module"
          Message="You can enter your modules solution's file path. Examples: C:\MyProjects\Amazing\Amazing.Module.GodeGen.sln"
          Action="Add an existing module" 
          Security="SecurityAccessLevel.Admin" 
          Class="btn btn-success" 
          OnClick="@(async () => await LoadModule())" ResourceKey="LoadModule" />
```

3. Result as follows 

image


sbwalker:
@vnetonline I have been focused on the 5.0 release... I will have feedback on this PR soon...

.gitignore
```diff
@@ -29,3 +29,6 @@ Oqtane.Server/wwwroot/Modules
 Oqtane.Server/wwwroot/Themes
 !Oqtane.Server/wwwroot/Themes/Oqtane.Themes.*
 !Oqtane.Server/wwwroot/Themes/Templates
+
+Oqtane.Server/wwwroot/_content/syncfusion.*
```
hishamco, line 33: Seems this is custom to your project

.gitignore
```diff
@@ -29,3 +29,6 @@ Oqtane.Server/wwwroot/Modules
...
+/Oqtane.Server/appsettings.json
```
hishamco, line 34: Why?!!

Oqtane.Client/Modules/Controls/ActionDialog.razor
```diff
@@ -56,6 +71,12 @@
     [Parameter]
     public string Text { get; set; } // optional - defaults to Action if not specified
 
+    [Parameter]
+    public bool ShowTextBox { get; set; } = false; // optional - specifies if the TextBox should be displayed in the dialog - default is false
```
hishamco, line 75: I can't entirely agree with this because we might have multiple inputs, so it would be nice to have the body as a template

Oqtane.Client/Modules/Controls/ActionDialog.razor
```diff
@@ -83,6 +104,9 @@
     [Parameter]
     public string IconName { get; set; } // optional - specifies an icon for the link - default is no icon
 
+    [Parameter]
+    public EventCallback<string> OnClose { get; set; } // optional - executes a method in the calling component when the Modal is Closed
```
hishamco, line 108: I like to introduce event callback in such components as extensibility point, @sbwalker we might need to revise the available components

hishamco (changes requested):

vnetonline:
I will submit another pull request that uses `RenderFragment` for `Header`,  `Body` and `Footer` closing this as it is not flexible enough as pointed out by @hishamco 

Closing this pull request because it also has commits that relate to `syncfusion` and `.gitignore`

Issue #3497, opened 2023-11-21
Module Settings Container setting not saving.


leigh-pointer:
The container settings is not being retrieved and what saved not set.  This can lead to confusion as the message doesn't identify what tab it is erroring.

![ContainerSettingsNotSaving](https://github.com/oqtane/oqtane.framework/assets/9447612/33967054-1921-40be-a5f2-279efa7ee7c4)

sbwalker:
@leigh-pointer are you sure its not related to this item documented previously:

https://github.com/oqtane/oqtane.framework/discussions/2874

sbwalker:
I am not able to reproduce using the Container settings for the default Oqtane theme.

leigh-pointer:
@sbwalker I have done a bit of research, and it comes from changing themes on a page.  
I was viewing my module in RocketFuel Theme and then changed it over view it in Oqtane Theme.  Changing the Theme does not reset the Container in the database [dbo].[page].[DefaultContainerType].  The reason it is not updated, could be because the Container names are the same between the two themes.

I will look further and do a PR for a fix.

sbwalker:
@leigh-pointer This area is actually very complicated as there are many scenarios which need to be considered....

Containers are related to a specific Theme - so when a Theme is changed, the associated Container must be changed as well. 

Themes and Containers are hierarchical/cascading and can be set at the Site, Page, and Module level. 
- Site Default Theme / Container (default for entire Site)
- Page Default Theme / Container (only specified if it is different than the Site default - in which case it overrides - otherwise it should be unspecified)
- Module Container (only specified if it is different than the Page Default Container - in which case it overrides - otherwise it should be unspecified)

Scenario 1: A Page Default Theme is set which overrides the Site Default Theme. Later, the Site Default Theme  is changed.
>> should this reset the Page Default Theme settings for all Pages to inherit the new Site Default Theme? Or should it retain the existing Page Default Theme overrides?

Scenario 2: A Page Default Container is set which overrides the Site Default Container. Later, the Site Default Container  is changed.
>> I believe this should reset any Page Default Container settings for Pages where the previously set Page Default Container matches the new Site Default Container... and any Module Container settings where the previously set Module Container matches the new Site Default Container should be reset as well.

Scenario 3: A Module Container is set which overrides the Page Default Container. Later, the Page Default Container is  changed.
>> I believe this should reset any Module Container settings for the Page where the previously set Container matches the new Page Default Container

[...]

PR #3498, opened 2023-11-22
Fix for  #3497 Module Settings Container setting not saving.

leigh-pointer:
In the Add and Edit of Pages the Container was not being set, but only set if the Theme wasn't equal the default but not if the Theme was set back to the Default.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Pages/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Pages/Add.razor
@@ -306,11 +306,11 @@
     private void ThemeChanged(ChangeEventArgs e)
     {
         _themetype = (string)e.Value;
+        _containers = ThemeService.GetContainerControls(PageState.Site.Themes, _themetype);
+        _containertype = _containers.First().TypeName;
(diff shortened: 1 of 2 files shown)


sbwalker:
@leigh-pointer the concept behind themes and containers is that they are hierarchical. Basically this means that the Module Container should only be set to a value if it overrides the Default Container for the page - otherwise it should be blank and inherit the Default Container.

So when changing a theme in Add/Edit Page, the Module Containers for that page should all be set to blank so that they inherit the Default Container for the page.

leigh-pointer:
@sbwalker ok understand.  So this PR needs some extra work.  The way it was working before is the container wasn't getting set correctly when resetting the theme back to the Default.  The old container was still in the database.

sbwalker:
@leigh-pointer I am thinking it may be better to implement this server-side as opposed to implementing it in the UI. Do you want me to create a PR?

leigh-pointer:
@sbwalker i think server side will do the trick! So answer to your question, a PR would be appreciated.

sbwalker:
closing based on my comments in #3497

PR #3528, opened 2023-12-01
File Extension management - site wide.

leigh-pointer:
I have added to the site settings file extension management. The Constants remain for backward compatibility.
If the extensions are not updated then the Constant will be used.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Files/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Files/Add.razor
@@ -4,6 +4,7 @@
 @inject NavigationManager NavigationManager
 @inject IFileService FileService
 @inject IFolderService FolderService
+@inject ISettingService SettingService
 @inject IStringLocalizer<Add> Localizer
 @inject IStringLocalizer<SharedResources> SharedLocalizer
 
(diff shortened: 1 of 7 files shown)


sbwalker:
@leigh-pointer one other observation... the logic for using FileManager was very simple for developers in the past:

<FileManager FileId="@_logofileid" Filter="@Constants.ImageFiles" @ref="_logofilemanager" />

With the addition of a property to the Site model it could be:

<FileManager FileId="@_logofileid" Filter="@PageState.Site.ImageFiles" @ref="_logofilemanager" />

The ImageFiles property would be read only and would utilize the Site.Settings property internally to get the value.

This eliminate the need for developers to have to figure out how to get the value from Settings and would simplify the code.

leigh-pointer:
@sbwalker that is already being set.  Settings are only saved from the Site component.
image

sbwalker:
@leigh-pointer you are not understanding the behavior of the IsPrivate property. It has nothing to do with where settings are saved... it has to do with where settings are read. As an example, if you are not an administrator and you browse to your user profile so that you can upload a photo... this uses the FileManager component. The FileManager component uses the ImageFiles setting to set the list of allowable file extensions which a user can upload. In this context (non-Admin) the ImageFiles setting will NOT exist in the PageSettings.Site.Settings collection. This is because it was created as a Private setting - which means that it should NOT be shared externally. The only exception is when the user is an Administrator it will be in the collection because the Admin user is authorized to access it.

PR #3548, opened 2023-12-12
Module Management Ehancement - show where modules are used.

leigh-pointer:
image

This enhancement enables a check on where the moduleis used before it is deleted

image


sbwalker:
@leigh-pointer this is a useful enhancement. I would suggest however that we simplify it and only display the pages where the module exists - and not display the number of instances of the module on those pages. This would greatly simplify the query from a performance perspective ie. just get the unique pages where the module exists and display the list - do not worry about grouping them, etc... This would still satisfy the requirement to prevent deletion if the module exists on any page. Also can we name the tab simply "Pages".

PR #3570, opened 2023-12-18
Resolve Profile Photo Url Issue Fixes #3569

### LINKED ISSUE #3569: Bug: User Profile Image Not Showing
The user profile image source attribute is resolving with an api request instead of the profile url.  

The if statement checks for photo but then does not use photo to get the url.

I have a PR that resolves this and maybe acceptable I will submit in a minute.

thabaum:
Fixes #3569

This pull request changes the source attribute of the image element to use `@photo.Url` to set the correct path for the profile photo property to be viewed at the top of the user profile page.

Unless there is some caching or resizing large images going on with the other method of `ImageUrl` that needs to be worked out this maybe acceptable solution.

### DIFF
--- a/Oqtane.Client/Modules/Admin/UserProfile/Index.razor
+++ b/Oqtane.Client/Modules/Admin/UserProfile/Index.razor
@@ -16,7 +16,7 @@
 {
     @if (PageState.User != null && photo != null)
     {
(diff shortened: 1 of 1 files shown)


sbwalker:
@thabaum this is not the correct solution. The ImageUrl() helper uses the dynamic image resizing capability to ensure a photo is constrained to a specific size (and compressed). If an image is not appearing in your case it is related to a different problem.

PR #3572, opened 2023-12-18
Sets Logo max height to 90px plus padding 5px right and left.

### LINKED ISSUE #3571: Issue: Logo if bigger than 90px in height pushes navigation menu on top of content pane.
Logo bigger than 90px or so will overlap the content making things like titles of modules or pages hidden from users along with any notifications on top of the page in the event a user uploads an image with a 300px height for example.

thabaum:
Fixes #3571 

Sets Logo max height to 90px plus padding 5px right and left.

If even 150px it will cover up content in the admin pages and other places.

It would be worth considering creating variables that adjust logo and content margins.  I set this in the app.css at the top since it is the logo image first thing you see when loading app.  I had also thought this could go in the Oqtane Theme.css however .app-logo is part of the app.  I can produce another PR if this should be in the Oqtane theme.css instead.

I can also explore creating variables that can adjust according to the image height uploaded which was another idea I had but would take more investigating that possibility.

### DIFF
--- a/Oqtane.Server/wwwroot/Themes/Oqtane.Themes.OqtaneTheme/Theme.css
+++ b/Oqtane.Server/wwwroot/Themes/Oqtane.Themes.OqtaneTheme/Theme.css
@@ -4,6 +4,12 @@ body {
     padding-top: 7rem;
 }
 
+/* App Logo */
+.app-logo .img-fluid {
+    max-height: 90px;
+    padding: 0 5px 0 5px;
+}
+
 .table > :not(caption) > * > * {
     box-shadow: none;
 }
--- a/Oqtane.Server/wwwroot/Themes/Templates/External/Client/wwwroot/Themes/[Owner].Theme.[Theme]/Theme.css
+++ b/Oqtane.Server/wwwroot/Themes/Templates/External/Client/wwwroot/Themes/[Owner].Theme.[Theme]/Theme.css
@@ -4,6 +4,12 @@ body {
     padding-top: 7rem;
 }
 
+/* App Logo */
+.app-logo .img-fluid {
+    max-height: 90px;
+    padding: 0 5px 0 5px;
+}
+
 .table > :not(caption) > * > * {
     box-shadow: none;
 }
--- a/Oqtane.Server/wwwroot/css/app.css
+++ b/Oqtane.Server/wwwroot/css/app.css
@@ -10,12 +10,6 @@ app {
     flex-direction: column;
 }
 
-/* App Logo */
-.app-logo .img-fluid {
-    max-height: 90px;
-    padding: 0 5px 0 5px;
-}
-
 /* Admin Modal */
 .app-admin-modal .modal {
     position: fixed; /* Stay in place */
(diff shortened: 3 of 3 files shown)


sbwalker:
@thabaum did you ensure that this change works properly in responsive design scenarios ie. on mobile view?

sbwalker:
Adding this CSS to app.css has the potential to affect the behavior of every third party theme... so it is a highly risky change to introduce.

thabaum:
Removed from App.css and added to the theme creator as well as the Oqtane theme itself.

sbwalker:
@thabaum I have verified that this change looks fine in both standard mode and Responsive Design Mode.

Note that .NET MAUI is simply using an embedded WebView - which essentially means that it relies on responsive design for the display of the UI. So you can use your standard browser dev tools and the responsive design options to determine how the UI will be rendered at different resolutions - you do not need to run .NET MAUI to do this.

image

thabaum:
@sbwalker I did test with Maui using the Windows Store App Emulator, I just cannot do the Mac and Droid OS's currently which I will be able to shortly.

I want to work on utilizing some device native features such as camera and menu options along with harnessing Oqtane as a way to upload, deliver, authenticate, validate and control content with these types of application features for end users.

Thanks for checking, as you said I figured the same that it is just a web view so it should utilize the device browser which would look same as a responsive view making the window small.  Love this stuff in respect to bootstrap.  

What I want to be testing using Maui is the app user experience launching it and what happens when the app cannot connect.  Those types of experiences since it is connecting via api and is a standalone application loaded on devices that pulls up a site more or less but still needs to give a user a notification there is an issue connecting to the site, it may be under maintenance or your network connection has been disrupted.  I have not got that far to review how this experience lives on and I am not hearing too many discussions other than a couple at developer meetings about current app store use of Oqtane.

In fact I just went ahead and checked up on this as I ran a second instance of the Oqtane.Maui program installed.  I have a few issues that come to mind from the start we can maybe look to address:

1. Project name is used for the Oqtane.Maui App Name and the App Icon is default Oqtane icon. I understand this can be changed but I would think it would use the logo provided for the site and the name of the site.  So it would need a way to fetch this to include and build this app.  This issue can most likely be dismissed since it should be set manually most likely.

2. If the server is down, the app errors out with "reload" link at bottom.  If server goes back online you hit reload and the page / app is online.  This is not a desirable way to present a page error when it is an app.  Oqtane.Maui project app should handle this event more graciously and catering to the end users experience dealing with the server being down, or a connection issue.  It should check if there is internet on the device first and let the user know to enable a wireless connection to use this app.  If there is internet connectivity and still no connection then the server maybe undergoing or experiencing issues and a try again later type of message.  A button instead of a reload link of "try again" once these issues have been resolved.

Issue #1 not really an issue, but #2 is I believe.  This area of connectivity issues could be enhanced.  The app experience needs to have some level of handling this event that keeps a user from thinking the app is bugged out with respect to the error message and reload page.  I think a static page that in a fun way says oops something is wrong, this might help, come back later or click try again if you resolved internet connectivity issues.  Can put up a support message as needed.  Need some template for this I suppose.

PR #3586, opened 2023-12-20
Refreshes Admin Settings Page When User Updates Logo.  Fixes #3582

### LINKED ISSUE #3582: Issue: Site Settings - Save Button Action Does Not Update Logo View On Page Without A Page Refresh
Save does not update the page on Site Settings.

In a way you may even want to preview the change in the logo area prior to saving it to be sure it looks good. 

But after saving the site settings page component I am expecting to see the new logo I selected for the page without a page refresh.

Also on some pages with this notification on top if you hit save twice while staying on this page and close the notification box, a second notification box does not appear.  So a fade in effect that allows the message to generate some animation effect a little to let a user know it was saved and that is a new saved message or if the notification box is closed... a new notification is created on the next save while still visiting that page.

I maybe need to make a series of issues up here but putting the main issues in focus.

thabaum:
Fixes: #3582

This PR will simply refreshes the current page in order to view the change of the updated logo.

This is only performed when a logo file id has changed from original loaded on the page it gets on initialization from `logofileid `and compares it with` _logofileid` which is are used for different areas of code logic.  I originally had a new variable set to know it was the original but found one that would serve the same purpose not being used or reset to compare with the one that was updated from the file manager component. 

This all relates to the site settings component during Oqtane administration when a user changes the logo and performs the save button action.

I had a few other "NavigationManager" version to refresh, this one used in commit seemed so simple I used it.

After a few ideas this one is the best I have had so far as I tried `StateHasChanged()` and such but no luck.   Maybe when Oqtane 5.1 rolls out some of this can be handled differently.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Site/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Site/Index.razor
@@ -581,7 +581,8 @@
                             site.LogoFileId = logofileid;
                             if (logofileid != _logofileid)
                             {
-                                reload = true;
+                                _logofileid = logofileid;
+                                refresh = true;
                             }
                         }
                         int? faviconFieldId = _faviconfilemanager.GetFileId();
--- a/Oqtane.Client/Modules/Admin/Site/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Site/Index.razor
@@ -582,7 +582,7 @@
                             if (logofileid != _logofileid)
                             {
                                 _logofileid = logofileid;
-                                refresh = true;
+                                refresh = true; // needs to be refreshed on client
                             }
                         }
                         int? faviconFieldId = _faviconfilemanager.GetFileId();
--- a/Oqtane.Client/Modules/Admin/Site/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Site/Index.razor
@@ -673,6 +673,10 @@
 						}
 						else
 						{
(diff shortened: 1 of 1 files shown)


sbwalker:
@thabaum if you look at the SaveSite() method you will notice that there is already logic to track whether the page needs to be refreshed or reloaded (some content changes can be handled with a UI refresh whereas others need a full page reload). So the logic for checking if the logo changed should simply set the existing refresh = true boolean.

thabaum:
@sbwalker gotcha, I wasnt 100% sure on that part of the logic but I did see it.  I will doctor this PR to use this instead.

thabaum:
@sbwalker the adjustment has been made and tested to use this check and set to use refresh page boolean.  I put this logic inside the logo file manager if statement.

thabaum:
@sbwalker I played a bit with this today.  So I have come up with 2 solutions.  One updates the client so they feel like they see the changes which saved.

The code below will do this:

```
  if (logofileid != -1)
  {
      site.LogoFileId = logofileid;
      if (logofileid != _logofileid)
      { 
          _logofileid = logofileid;
          refresh = true;
          
      }
  }
```

The other option is a complete reload.

I can update the PR to utilize this new logic if you prefer the page not to refresh on the client.  I think it works for me and it is faster.  I just needed to update the variable to the new logo id on save logo event.

PR #3587, opened 2023-12-21
Maui app connectivity error message enhancement.  Fixes #3583

### LINKED ISSUE #3583: Enh: Improved Handling of Connectivity Issues in Oqtane.Maui
# Enhancement Request

Improved Handling of Connectivity Issues in Oqtane.Maui

## Description

The current behavior of the Oqtane.Maui app, when there is no connectivity to the server, generates a generic error without providing sufficient information or guidance to the user. This enhancement request aims to improve the user experience by implementing a more informative and user-friendly error handling mechanism.

## Proposed Solution

Implement a dynamic landing page or template stored in the app's (shared) project, providing static information such as theme color, logo, and relevant messaging for scenarios when the app encounters connectivity issues. The landing page should include:

- A "Try Again Later" message.
- Troubleshooting guidance for potential issues, including server maintenance, internet connectivity, etc.
- Internet connectivity check, instructing the user to resolve connectivity issues first if detected.
- Action buttons like "Retry" and "Exit" for user interaction.

Utilize the [Network Connectivity .NET 8](https://learn.microsoft.com/en-us/dotnet/maui/platform-integration/communication/networking?view=net-maui-8.0&tabs=windows) capabilities to enhance the error handling and provide a more comprehensive solution.

## Visual Representation

Error Page

The envisioned error page would follow the default Oqtane theme, incorporating a static layout with the app's logo, background colors, and actionable buttons for user engagement.

## Additional Notes

- This enhancement aligns with providing a consistent and user-friendly experience for all error scenarios in the Oqtane.Maui app.
- A consistency would be needed between the Web and Maui applications similar error messaging.
- Consideration should be given to CSS adjustments within the theme for handling error messages effectively.

---

Feel free to provide feedback and suggestions for further refinement.

thabaum:
Fixes #3583

This resolves the unhandled exception issue by giving a more clear path to an end user on what to do and why the issue may be happening in the first place.

Notes:
could be a work in progress...  
I am trying to understand how everything loads and if it can be improved upon out of the box.
The background color could be set to default black like the default Oqtane theme.

image

image

This has the following features:
1. Is responsive
2. Is accessible
3. Utilizes stock images and resources
4. Removes the dismiss X button since it will leave a user with nothing else to do but close the application.
5. Gives generic simple instructions and reasons for possible troubleshooting actions or other potential causes.

Other feedback relating:
What color should the background be?  Black I am thinking since the app has a black background to start.  So I can add this to match the default Oqtane theme as well if that would be OK or we can leave it white.  The "loading..." goes by so fast you dont notice it... it would be nice to have something more real to look at that goes until the message pops up.

I changed the color from light yellow to this light red to reflect it is an alert that something is wrong.  This can be and maybe should be changed back to the other color background.

If you shut down the server, then navigate to a link the app closes if running the app from desktop, and it will run into a debug stop point if running from debug mode.  It would be nice to resolve them back to this default Index.html.  This is something I am hoping can still get fixed and is looked further into for production.  An app just closing out of the blue isnt good, but it is better than an error I suppose.   I just created an issue for this #3588.

### DIFF
--- a/Oqtane.Maui/wwwroot/css/app.css
+++ b/Oqtane.Maui/wwwroot/css/app.css
@@ -196,20 +196,113 @@ app {
 }
 
 #blazor-error-ui {
-    background: lightyellow;
-    bottom: 0;
-    box-shadow: 0 -1px 2px rgba(0, 0, 0, 0.2);
-    display: none;
-    left: 0;
-    padding: 0.6rem 1.25rem 0.7rem 1.25rem;
     position: fixed;
-    width: 100%;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+    text-align: center;
+    max-width: 90%;
+    width:100%;
+    height: 90%;
+    padding: 10px;
+    background-color: #f8d7da;
+    box-shadow: 0 0 10px rgba(0, 0, 0, 0.2);
+    border-radius: 10px;
+    display: none;
     z-index: 1000;
 }
 
-#blazor-error-ui .dismiss {
-    cursor: pointer;
-    position: absolute;
-    right: 0.75rem;
-    top: 0.5rem;
+    #blazor-error-ui .logo {
+        display: flex;
+        justify-content: space-between;
+        align-items: center;
+    }
+
+        #blazor-error-ui .logo img {
+            max-width: 300px;
+            max-height: 150px;
+            height: auto;
+            display: block;
+            margin: 0 auto;
+            padding: 10px;
+        }
+
+    #blazor-error-ui .image {
+        text-align: center;
+        margin-top: 20px;
+    }
+
+        #blazor-error-ui .image img {
+            width: 90px;
+            height: auto;
+        }
+
+    #blazor-error-ui .heading {
+        color: #f44336;
+        margin-top: 10px;
+        font-size: 1.5em;
+    }
+
+    #blazor-error-ui .message {
+        font-size: 1em;
+        max-width: 720px;
+        margin: 20px auto;
+        padding: 10px;
+        align-content: center;
+        transition: opacity 0.3s ease;
+    }
+
+    #blazor-error-ui .reload {
+        display: block;
+        margin: 20px auto;
+        width: 200px;
+        padding: 15px 20px;
+        font-size: 1.2em;
+        background-color: #2a9fd6;
+        color: white;
+        border: none;
+        border-radius: 8px;
+        cursor: pointer;
+        transition: background-color 0.3s ease;
+    }
+        #blazor-error-ui .reload:hover {
+            background-color: #247aa8;
+        }
+
+    #blazor-error-ui .dismiss {
+        cursor: pointer;
+        position: absolute;
+        right: 0.75rem;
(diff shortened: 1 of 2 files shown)


sbwalker:
@thabaum the blazor-error-ui is a Blazor concept which is applicable to all hosting models - not only .NET MAUI. So whatever we do in this area should be done in a manner which is consistent across all platforms. In addition, the blazor-error-ui is not only related to connectivity errors - it is displayed for any unhandled exception. So this will need some further consideration before it can be merged.

thabaum:
you are correct `the blazor-error-ui` seems to only be used when a connectivity error happens.  The issue is this is a client side application and it needs this to default to something in the app which is this page.  You can see the Loading... for a blip of a second as well along with some rendering of the site if it gets loaded when it switches to it if it does not load this error message.   Some glitchy behavior might resolve in production.

And if you have something like a page created here some of it will show up with the Oqtane theme so it is rendered somewhat always in the application... just you dont see anything since it is a blank background and the theme styles will be on top eventually while the stack loads.

So this page only affects users of the Oqtane.Maui applications created.  It wont affect the web app since there is nothing shared between them.  I was thinking there should be and the Oqtane.Maui project added to the entire project but I understand not every case needs a Maui project and you can run this project on it's own.

I am not 100% this all should be used I want this to be a usable solution for all devices.  This is where the responsiveness of the css in the app comes into play and why I use that ID selector exclusively with all the styles.  This way it only affects Maui applications and I think we need a starting point and a way to style it from there to improve it as more devices get tested.

This was my suggested solution that if someone can one up from here.

This just takes that message and makes it into a notification with a reload action while removing the dismiss action.  I put the application logo along with a icon that vanishes if the screen gets smaller.  I can improve this as I believe a 540w x 360h type of screen or whatever maybe in that size area the text is pushed under.

I did make it so the button to reload is always fixed in this case.

I was thinking about putting a swipe down feature in it so you can refresh with a swipe anytime.  Sometimes the logo does not appear and I want to refresh it.  Reloading the app is the only way to fix it.

thabaum:
I am working around in this area currently a little more moving my current changes to a new project with the latest build information to test a few other Maui things out both local and IIS deployed environments using the Blazor Maui and Blazor Web Oqtane Framework applications.  I am separating concerns more as I had them bunched into one project and losing where I was working last.  I will get this area tested further asap and report back with an updated PR to review shortly.

I am experiencing an issue firing up Oqtane.Maui project after adding it to the Oqtane Framework solution which I did not have an issue before.  Project kept faulting when you add it forcing you to reload that project to make it build properly.  Not sure if that is a normal thing.

Things I will put into a more final rough draft will include more considerations for different cases and making it work well both web and server.  I can now see how it is also nice as a smaller message with dismiss as it leaves ability to navigate to a page that works without an error.

I would add theme.css and include these changes in the message and Oqtane Theme itself I believe to make the framework utilize the same look and feel as the app.

thabaum:
I am just going to close this for now, until I have more time to explore all the possibilities with the error message.  I will reopen if I can get around to it.  And really this is expected behavior so it is only a theme enhancement anyone can easily apply.

Issue #3538, opened 2023-12-11
ENH: Start Date and Expiry Date for Module instances and Pages


leigh-pointer:
# Start Date and Expiry Date for Module Instances and Pages in Oqtane

## Overview

The "Start Date and Expiry Date for Module Instances and Pages" feature in Oqtane enhances content management capabilities by providing administrators with control over the visibility and availability of modules and pages. This feature introduces time-based activation and deactivation for precise content scheduling.

### Module Instances Activation

- Administrators can set a "Start Date" for module instances to determine when a specific module becomes active.
- Modules remain inactive until the designated start date is reached, ensuring content is revealed at the intended time.

### Module Instances Deactivation

- "Expiry Date" can be assigned to module instances for automatic deactivation at a specified time.
- Modules seamlessly transition to an inactive state once the expiry date is reached, streamlining content lifecycle management.

### Page Activation and Deactivation

- Pages can be configured with "Start Date" and "Expiry Date" for granular control over visibility.
- A page's "Start Date" activates the entire page, while the "Expiry Date" automates deactivation based on the schedule.

### User Notification

- Oqtane provides notification mechanisms to alert administrators before module or page activation or expiry, facilitating proactive content management.

### Dashboard Control

- Centralized dashboard in the Oqtane administration interface offers an overview of all scheduled start and expiry dates.
- Efficiently manage and monitor content deployment and deactivation for streamlined content governance.

## Benefits

- **Content Strategy:** Facilitates strategic content planning by enabling administrators to schedule activations in advance.
- **Timely Information Delivery:** Ensures timely display of information through automated activation and deactivation.
- **Efficient Content Management:** Reduces manual workload by automating content activation and deactivation.
- **Improved User Experience:** Enhances user experience by presenting relevant and up-to-date content.

W6HBR:
This will be a much needed enhancement!

One thought that that comes to mind on page activation is that you will need to have a mechanism to be able to have a duplicate "Url Path" defined for new pages so that they can replace existing pages.  You currently cannot add multiple pages with the same Url Path (for obvious reasons).

leigh-pointer:
@W6HBR 
> duplicate "Url Path" 
 As far as i see it, a page will be either active or inactive (visible of invisible) 

How are you seeing this functioning?

W6HBR:
@leigh-pointer 
[...]

PR #3602, opened 2023-12-31
Start Date and Expiry Date for Module instances and Pages #3538

leigh-pointer:
This is complete excluding Reporting and Notifications which can be added at a later date.  I just really wanted to get the schema and the functionality into place.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Modules/Settings.razor
+++ b/Oqtane.Client/Modules/Admin/Modules/Settings.razor
@@ -49,6 +49,18 @@
                         </div>
                     </div>
                     <div class="row mb-1 align-items-center">
+                        <Label Class="col-sm-3" For="effectiveDate" HelpText="The date that this module is active" ResourceKey="EffectiveDate">Effective Date: </Label>
+                        <div class="col-sm-9">
+                            <input type="date" id="effectiveDate" class="form-control" @bind="@_effectivedate" />
+                        </div>
+                    </div>
+                    <div class="row mb-1 align-items-center">
+                        <Label Class="col-sm-3" For="expiryDate" HelpText="The date that this module expires" ResourceKey="ExpiryDate">Expiry Date: </Label>
+                        <div class="col-sm-9">
+                            <input type="date" id="expiryDate" class="form-control" @bind="@_expirydate" />
+                        </div>
+                    </div>
+                    <div class="row mb-1 align-items-center">
                         <Label Class="col-sm-3" For="allpages" HelpText="Indicate if this module should be displayed on all pages" ResourceKey="DisplayOnAllPages">Display On All Pages? </Label>
                         <div class="col-sm-9">
                             <select id="allpages" class="form-select" @bind="@_allPages" required>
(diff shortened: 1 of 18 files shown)


sbwalker:
Thank you @leigh-pointer this is a good starting point for this enhancement. I had not tackled it previously because there are a lot of aspects to consider. In reviewing the PR I have a number of comments/questions...

1. I believe we would want to allow users to enter dates in their local time and have the system translate them into UTC time. The Scheduler has methods for handling the translation for UI purposes.

2. ValidateEffectiveExpiryDates should not be repeated in many places - it should be a static method in the Shared project.

3. isAdminOrHost is not necessary as a Host user ALWAYS has Administrator rights. We should not introduce new security methods - use the existing UserSecurity.IsAuthorized(PageState.User, RoleNames.Admin)

4. From a functionality perspective, I believe a site admin should be able to access pages regardless of effective/expiry dates. Also, if a user has Edit rights for a page they should be able to access modules regardless of effective/expiry dates.

5. The GetSite method on the server should return pages and modules regardless of effective/expiry dates - ie. IsPageModuleVisible filtering should be done on the client. This is because the effective/expiry criteria is variable and is based on a specific moment in time. So it is possible to access a site at 11:59 PM - which will then cache the pages/modules on the client - and at 12:00 AM some of the pages/modules may become effective or expired and will need to be filtered.

sbwalker:
@leigh-pointer thank you for the changes...


isAdminOrHost is not named appropriately - but it is also not needed:

```
isAdminOrHost = UserSecurity.IsAuthorized(user, RoleNames.Admin) || UserSecurity.IsAuthorized(user, PermissionNames.Edit, page.PermissionList);
```

An Admin always has Edit permissions for a page - so there is no need to check if a user is an Admin - you simply need to check UserSecurity.IsAuthorized(user, PermissionNames.Edit, page.PermissionList)

sbwalker:
@leigh-pointer I also do not see where the PageModules are being filtered based on Effective/Expiry date on the client?

sbwalker:
@leigh-pointer this login in SiteController seems strange:

``
                foreach (PageModule pagemodule in _pageModules.GetPageModules(site.SiteId).Where(pm => !pm.IsDeleted && _userPermissions.IsAuthorized(User, PermissionNames.View, pm.Module.PermissionList)))
                {
                    if (!pagemodule.IsDeleted && _userPermissions.IsAuthorized(User, PermissionNames.View, pagemodule.Module.PermissionList))
```

it is using Linq to filter the collection but then checking the same condition inside the loop

**Comment by sbwalker** [comment]
@leigh-pointer this logic seems incorrect to me:

```
if ((Utilities.IsPageModuleVisible(page.EffectiveDate, page.ExpiryDate) || UserSecurity.IsAuthorized(user, PermissionNames.Edit, page.PermissionList)) || UserSecurity.IsAuthorized(user, PermissionNames.View, page.PermissionList))
```

Basically this says that if IsPageModuleVisible = true then all other conditions will be ignored (since they are all OR conditions) - and it would render the page - which is incorrect.

I believe the logic should be:

```
if (UserSecurity.IsAuthorized(user, PermissionNames.View, page.PermissionList) &&
(Utilities.IsPageModuleVisible(page.EffectiveDate, page.ExpiryDate) || UserSecurity.IsAuthorized(user, PermissionNames.Edit, page.PermissionList)))
```

this ensures the View permission is the primary criteria and is always validated

PR #3610, opened 2024-01-04
Update Oqtane.Client\Modules\Admin\Profiles\Edit.razor to fix breaking changes

W6HBR:
Prior functionality allowed 4 characters for "order" which now causes validation errors for any order value > 99.  This change allows a value of up to 9999 which is consistent with the prior 4 character length value.

Prior functionality allowed 2 characters for "rows" which now causes validattion error for any row count > 10.  This change allows a value up to 99 which is consistent with the prior 2 character length value.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Profiles/Edit.razor
+++ b/Oqtane.Client/Modules/Admin/Profiles/Edit.razor
@@ -34,7 +34,7 @@
         <div class="row mb-1 align-items-center">
             <Label Class="col-sm-3" For="order" HelpText="The index order of where this profile item should be displayed" ResourceKey="Order">Order: </Label>
             <div class="col-sm-9">
-                <input id="order" class="form-control" @bind="@_vieworder" min="0" max="99" type="number" required />
+                <input id="order" class="form-control" @bind="@_vieworder" min="0" max="9999" type="number" required />
             </div>
         </div>
         <div class="row mb-1 align-items-center">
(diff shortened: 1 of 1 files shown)


sbwalker:
@W6HBR the problem which the earlier PR was meant to address was that Profile Management did not have sufficient validation on the items entered. As a result some validation criteria was included. In the case of Order I can understand why the range of 0-99 is insufficient. However in the case of Rows I would prefer to not allow users to enter a value greater than 10 - as this creates a terrible UI experience. I am not concerned about the breaking behavioral change in this scenario, as I believe a validation error is the appropriate result.

W6HBR:
@sbwalker Yes, totally agree 99 rows is too much for a good UI experience.  I could probably live with 10, but having up to 20 rows would work better for my current needs.  But like I said... I could live with 10.

thabaum:
I felt 10-20 was about right, 10 on the low and 20 on the high.  I sided with the fact it is resizable and 10 rows is generally more than enough to enter some text to start and resize as needed if desired for user input.  I was concerned about the 9999 to 99 limitation, shocked it took this long to hear any reaction from the change but understandable.

thabaum:
@W6HBR and @sbwalker Since we are not limited by number of characters like 2 or 4 to keep the count in a range resulting in 0-9999 and and 0-99 we should not need to resort to these odd values.  I think it would make sense to use max="100000".   Adding an additional number not a big deal and neither is 9999 max value really. I just think since we are fine tuning this line of code it makes sense to allow an even number of categories ending with 0's.  Again not a big deal but thought hit me when reviewing the change made once it was set.  Just my opinion here.  Cheers!

PR #3649, opened 2024-01-13
Cleans protocol to check for duplicate alias prior to saving.  Fixes #3647

### LINKED ISSUE #3647: Issue: Site Settings - If you add multiple alias's that are the same duplicate entries are created.
### Steps to Reproduce
1. Create a site
2. Create an alias that is HTTP and one that is HTTPS for the same domain alias after the // ( https://domain.com, http://domain.com)
3. Results is two alias's available

### Expected Behavior
A message saying this alias already exists.

### Additional Context
This may also allow domain.com http://domain.com and https://domain.com and end up with 3.   i have not tested how many times and different ways I can make the same alias appear.

thabaum:
Fixes #3647

# Pull Request: Clean Protocol and Scroll to Top on Error

## Summary

This pull request includes changes to clean the protocol (HTTP or HTTPS) from the alias before checking for duplicates. It also includes changes to scroll to the top of the page when an error message is displayed.

## Changes

1. **Clean Protocol:** Modified the `SaveAlias` method to remove 'http://' and 'https://' from the alias name before checking for uniqueness in the database. This ensures that the alias is unique even after removing these protocols.

```
string cleanedAliasName = _aliasname.Replace("http://", "").Replace("https://", "");
var alias = aliases.Where(item => item.Name == cleanedAliasName).FirstOrDefault();
```

2. **Scroll to Top on Error:** Added code to scroll to the top of the page when an error message is displayed. This ensures that the user can see the error message even if they are scrolled down on the page.

## Test Plan
Enter an alias with the ‘http://’ or ‘https://’ protocol and save it.
Enter the same alias again with the ‘http://’ or ‘https://’ protocol and try to save it. The application should display an error message and scroll to the top of the page.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Site/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Site/Index.razor
@@ -812,41 +812,46 @@
 		}
 	}
 
-	private async Task SaveAlias()
-	{
-		if (UserSecurity.IsAuthorized(PageState.User, RoleNames.Host))
-		{
-			if (!string.IsNullOrEmpty(_aliasname))
-			{
-				var aliases = await AliasService.GetAliasesAsync();
-				var alias = aliases.Where(item => item.Name == _aliasname).FirstOrDefault();
-				bool unique = (alias == null || alias.AliasId == _aliasid);
-				if (unique)
-				{
-					if (_aliasid == 0)
-					{
-						alias = new Alias { SiteId = PageState.Site.SiteId, TenantId = PageState.Site.TenantId, Name = _aliasname, IsDefault = bool.Parse(_defaultalias) };
-						await AliasService.AddAliasAsync(alias);
-					}
-					else
-					{
-						alias = _aliases.Single(item => item.AliasId == _aliasid);
-						alias.Name = _aliasname;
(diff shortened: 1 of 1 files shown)


sbwalker:
@thabaum is there a reason why you introduced a new variable "cleanedAliasName" and simply did not sanitize the existing _aliasname?

thabaum:
@sbwalker I believe my reasoning behind it was in the event _aliasname was used or needed to be used down the road in other places and didnt need to be sanitized.  I will update the logic if it makes more sense to just sanitize _aliasname instead of introducing a new variable as I am probably overthinking it as it sounds like it would be preferred to not have the new variable.

thabaum:
@sbwalker made adjustments to pr to not introduce a new variable and is ready for your review.

Thank you.

sbwalker:
@thabaum sorry... one more item... the logic to remove http and https will only work if the text is lowercase... perhaps a better approach would be to look for "://" and remove any characters from the beginning of the string up to and including "://" - this would eliminate any case sensitivity issue.

thabaum:
@sbwalker you are correct and was an approach I was thinking of using as well and should have, I will make this correction today.  Thanks.

PR #3652, opened 2024-01-14
Enh: Autocomplete Attribute Enhancement

### LINKED ISSUE #3651: Enh: Profile Autocomplete Attribute Added To Profile Properties
# Autocomplete Attribute Enhancement

## Description

I propose adding the ability to set the `autocomplete` attribute for profile properties in the framework. This enhancement would allow developers to control whether the autocomplete feature is enabled or disabled for specific profile fields.

## Use Case

This enhancement is beneficial for scenarios where developers want to customize the user experience by enabling or disabling autocomplete for certain profile properties. For example, they may want to disable autocomplete for sensitive information like passwords.

## Current Behavior

Currently, the framework does not provide a way to control the autocomplete attribute for profile properties.

## Proposed Enhancement

I propose adding a new setting or property to profile properties that allows developers to specify whether the autocomplete attribute should be enabled or disabled by entering in text such on "on" or "off" or "given-name".

@sbwalker I have a working pull request I will submit for review.

thabaum:
# Autocomplete Attribute Enhancement Pull Request

## Description

This PR implements the enhancement proposed in issue #[Issue_Number]: [Link to Enhancement Issue]

## Changes Made

- Added a new setting or property to profile properties to control the `autocomplete` attribute.
- Implemented the logic to conditionally include the `autocomplete` attribute based on the new setting.
- Also Adds the Autocomplete attribute for "name" and "email" to the user profile account tab input elements.

## Related Issue

Closes #3651 

## Screenshots
image

image

## How to Test

1. Visit Profile Management
2. Add text to the autocomplete input
3. Visit your profile and view the source to see the autocomplete attributed added only if you have it set.  Blank will not generate the autocomplete attribute.

## Additional Context
I can set some default autocomplete settings for the default profile properties, or this can be done later or leave it to admins.

The order of where it is on the Profile Edit Form can be changed as well if requested.

This PR also refactored the code in the OnInitializedAsync method of the Index component to obtain user settings earlier, ensuring their availability for subsequent operations. `SettingService.GetUserSettingsAsync`
for input elements:
image
for select elements:
image

If the autocomplete is empty, this is the result for an example:

image

notice no "autocomplete" attribute included.

Cheers!

### DIFF
--- a/Oqtane.Client/Installer/Installer.razor
+++ b/Oqtane.Client/Installer/Installer.razor
@@ -71,7 +71,7 @@
                 <div class="row mb-1 align-items-center">
                     <Label Class="col-sm-3" For="username" HelpText="Provide a username for the primary user account" ResourceKey="Username">Username:</Label>
                     <div class="col-sm-9">
-                        <input id="username" type="text" class="form-control" @bind="@_hostUsername" autocomplete="username" />
+                        <input id="username" type="text" class="form-control" @bind="@_hostUsername" />
                     </div>
                 </div>
                 <div class="row mb-1 align-items-center">
@@ -95,7 +95,7 @@
                 <div class="row mb-1 align-items-center">
                     <Label Class="col-sm-3" For="email" HelpText="Provide the email address for the host user account" ResourceKey="Email">Email:</Label>
                     <div class="col-sm-9">
-                        <input type="text" class="form-control" @bind="@_hostEmail" autocomplete="email" />
+                        <input type="text" class="form-control" @bind="@_hostEmail" />
                     </div>
                 </div>
                 <div class="row mb-1 align-items-center">
--- a/Oqtane.Client/Installer/Installer.razor
+++ b/Oqtane.Client/Installer/Installer.razor
@@ -69,9 +69,9 @@
             <h2>@Localizer["ApplicationAdmin"]</h2><br />
             <div class="container">
                 <div class="row mb-1 align-items-center">
-                    <Label Class="col-sm-3" For="username" HelpText="Provide a username for the primary user accountt" ResourceKey="Username">Username:</Label>
+                    <Label Class="col-sm-3" For="username" HelpText="Provide a username for the primary user account" ResourceKey="Username">Username:</Label>
                     <div class="col-sm-9">
-                        <input id="username" type="text" class="form-control" @bind="@_hostUsername" />
+                        <input id="username" type="text" class="form-control" @bind="@_hostUsername" autocomplete="username" />
                     </div>
                 </div>
                 <div class="row mb-1 align-items-center">
@@ -95,7 +95,7 @@
                 <div class="row mb-1 align-items-center">
                     <Label Class="col-sm-3" For="email" HelpText="Provide the email address for the host user account" ResourceKey="Email">Email:</Label>
                     <div class="col-sm-9">
-                        <input type="text" class="form-control" @bind="@_hostEmail" />
+                        <input type="text" class="form-control" @bind="@_hostEmail" autocomplete="email" />
                     </div>
                 </div>
                 <div class="row mb-1 align-items-center">
--- a/Oqtane.Client/Modules/Admin/Profiles/Edit.razor
+++ b/Oqtane.Client/Modules/Admin/Profiles/Edit.razor
@@ -77,9 +77,9 @@
             </div>
         </div>
         <div class="row mb-1 align-items-center">
-            <Label Class="col-sm-3" For="autocomplete" HelpText="Autocomplete setting for the property" ResourceKey="Autocomplete">Autocomplete: </Label>
+            <Label Class="col-sm-3" For="autocomplete" HelpText="The HTML autocomplete attribute allows you to specify browser behavior for automated assistance in filling out form field values, with a character limit of 30 to ensure concise and effective usage." ResourceKey="Autocomplete">Autocomplete: </Label>
             <div class="col-sm-9">
-                <input id="autocomplete" class="form-control" @bind="@_autocomplete" maxlength="50" />
+                <input id="autocomplete" class="form-control" @bind="@_autocomplete" maxlength="30" />
             </div>
         </div>
         <div class="row mb-1 align-items-center">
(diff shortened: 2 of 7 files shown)


sbwalker:
@thabaum there are a number of functional issues with this PR:

- the changes to the Profile model class are missing
- the migration to add the new property to the database is missing

From a higher level the property is technical and could be confusing to users so it needs to be better explained in the help text ie. "The HTML autocomplete attribute allows you to specify browser behavior for automated assistance in filling out form field values"

sbwalker:
@thabaum from a backward compatibility perspective the new property should not affect legacy behavior - so by default, existing profile fields should not specify the autocomplete attribute.

sbwalker:
@thabaum I would prefer to NOT enable this for new sites - it should be opt-in for Administrators AFTER they create a site.

sbwalker:
@thabaum the framework does not use migration "scripts" so I assume you mean a code-based migration class.

sbwalker:
@thabaum just a note that it is generally not a good practice to submit an incomplete PR. In this case the PR does not even compile because the code references a property in a model which is not included. And the feature is not functional because it has no database support. So if I had merged the PR, it would have broken the dev branch for all developers.

thabaum:
@sbwalker my bad, it was a late night and I thought I had merged all the changes into this PR.  I was trying to beat you to it here prior to review but your on the east coast.

It was an accident to forget this file again I had it working and will provide the last changes in a minute.

sbwalker:
@thabaum is there a reason why the database field is declared with a length of 200 and the input maxlength is 50? What is a realistic max field length for this item?

thabaum:
The only reason I did this was because it was allowed in the other fields as well in Profile so I followed it thinking it can be regulated with validation.  

I am glad you touched on this as I wasnt sure if I should put 50, my original setting, but seen 200 set for others and followed that as a guide.

Let me know what you think this should be and I can make this adjustment.

thabaum:
@sbwalker I will review this more closely tomorrow, as you are correct Admins should not have autocomplete attribute to allow the browser to auto fill in the form, but a user of any type should if editing their own personal profile properties.  I will see what I can do to control the logic if admin or host is viewing and it is not their profile they are looking at Oqtane will not add this attribute.

sbwalker:
@thabaum you don't need to modify the logic, as you only implemented autocomplete in Oqtane.Client/Modules/Admin/UserProfile/Index.razor - which only allows a user to manage their own profile. However it should be implemented in Oqtane.Client/Modules/Admin/Register/Index.razor as well - as that is the registration form for anonymous users.

thabaum:
@sbwalker I went ahead and added this to the installer.razor file since we are going ahead with adding these autocomplete attribute to missing user related input elements that relate to a user profile.

sbwalker:
@thabaum my apologies... I was mistaken about Oqtane.Client/Modules/Admin/Register/Index.razor - it does not allow the user to enter any profile fields other than the default fields - so it does not need any autocomplete modifications. 

The goal of this enhancement is that it does not force any change in behavior in the framework. So autocomplete should ONLY be added to Profile fields where the administrator has the ability to enable/disable the feature. It should NOT be included on non-Profile fields, as there is no way to control its existence. So it should be removed from username, email, etc... as this would force a change in behavior which we are trying to avoid.

thabaum:
@sbwalker OK, I will remove these, but I liked the results when testing in edge.  I will clean up all the autocomplete attributes hard coded for now.

thabaum:
@sbwalker I have not removed the ones from the installer thinking maybe this one can be left without hurting the behaviors since it is only the installer page?  I fixed a typo on that page anyhow I will leave, but just wondering if this would be a breaking behavior change even though it is just with the installer page?  If so I can remove the username and email autocomplete I recently just added.

Other than that this PR just handles the profile management by admins to include autocomplete on the profile elements they create.

thabaum:
I went ahead and removed the hard coded changes to this PR but left the spelling correction.  I will keep all this for a separate time since it is really not needed.  Also if you like I can create another PR that only has these changes now showing.

I will be focused on the file upload issues this weekend.

sbwalker:
@thabaum I am going to merge this as is. I do not want to expand this beyond the current Profile fields (ie. those defined in the Profile table). There are very few legacy forms where this would be applicable in Oqtane as it is only intended for forms where the current user is entering personal information about themselves.

sbwalker:
@thabaum for Issue 2 above are you saying that most legacy forms in Oqtane are missing the autocomplete="off" specification - and it should be added to prevent undesired behavior? This would actually make sense as the issues raised in the past where field values in hidden sections have  been overridden and caused problems may have been a result of this (or they could also have been a result of browser password manager plugins which also auto populate fields unexpectedly).

sbwalker:
@thabaum I have been thinking a lot about this enhancement. I feel like it is the beyond the technical abilities of most administrators in terms of how to use it effectively. Basically they would need to do some research into the HTML autocomplete attribute to understand the  behavior and to determine the appropriate standardized taxonomy values to specify in order to get any benefit. So as a result, in the majority of cases it will not be utilized. In general I am not a big fan of maintaining code long term unless a feature provides sufficient value to a large number of users. So I have been wrestling with how to get more value from this.

Despite the fact that modern browsers are inconsistent in their behavior for assisting users with populating forms, they all seem to attempt to provide this functionality by default (ie. it is enabled by default which sometimes leads to poor outcomes). The autocomplete attribute essentially allows developers to exert some control over the behavior by providing hints to the browser. 

So perhaps the addition of these attributes in Oqtane should not be viewed as a breaking behavioral change but rather as a fix to an existing UX issue. If we took this perspective then we do not need an option to enable the new functionality - it becomes the expected/default functionality. For example, it would make sense to include default autocomplete values for the default Profile fields when creating a new Site. And it would make sense to include them on the non-Profile input elements for username, email, etc... in the specific forms where a user is expected to enter their own personal information. And the majority of other forms in Oqtqne for non-personal information should specify autocomplete="off" ( as you suggested above). 

Perhaps this is what you were trying to say all along... however I did not have the time to really concentrate on this item until today.

PR #3666, opened 2024-01-22
[Bug] - Fixed the ability to upload Images when unloadable files are …

vnetonline:
…specified and don't include Images


sbwalker:
resolved in #3667

PR #3667, opened 2024-01-22
fix #3656 - UploadableFiles and ImageFiles settings not loaded properly to handle empty string value resulting in upload issues

### LINKED ISSUE #3656: File Upload Failed after Theme changed
image

sbwalker:


thabaum:
@sbwalker thanks for completing this.  I see I made that mistake of adding the extra variable again when not necessary.  I also see you changed SiteSate to PageState for grabbing Alias.  I was testing the other areas out with breakpoints for the relating Maui issue and was testing both these out and wanted to ask which was appropriate.  This clears that up for me, thanks again!

PR #3737, opened 2024-02-06
Fixes #3736 Navbar bootstrap lg size responsive Theme CSS and MenuHorizontal.razor Theme UI Component

### LINKED ISSUE #3736: Enh: Ability To Set Theme Navigation Bar Horizontal Menu Component Bootstrap Responsive Breakpoint
# Issue

Theme Navigation Toggle Menu Items Are Not Very User Friendly Creating Challenges For Theme Developers Due To The `MenuHorizontal.razor` component limitations.

https://github.com/oqtane/oqtane.framework/blob/dev/Oqtane.Client/Themes/Controls/Theme/MenuHorizontal.razor

## Current Behavior

The breakpoints for when things change on the menu are not set to bootstrap standards for md.  Changing this alone would at least make things break at the same point as the CSS bootstrap library does.

## Expected Behavior

The menu nav items break point works with number of menu items displayed.

## Solution

Proposing Enhancement For Theme Navigation Bar Horizontal Menu Component and CSS.  This could be by allowing a theme developer to pass a parameter to set the navigation menu breakpoints in bootstrap ie. ...-sm-md-lg-xl-...

Potentially some logic can count the total characters in the menu and dynamically set the breakpoints, however the issue is the CSS would need updated as well so might need something to override the responsive styling.   Doing it manually will allow instructions to set this, or logic needed to edit the css file as a few ideas.

## Screenshots
image

covers title of html module as seen here:
image

image

and if width set just right you get this:
image
This is due to the media minimum width allowing about 1px of space and rare screen size but can be handled by updating this to `767.98px`.  PR #3765 will close this issue.
## Additional Context

If components are to be built per theme to be used and these are templates to build these components from we should focus on the themes Oqtane supports by default and set this to Bootstrap sizes `lg - 991.98px` for the responsive CSS and bootstrap classes used in these components.  This will allow more pages to be created without causing an extra line covering up parts of the app page content at the top when using the Oqtane Theme and Theme Creator Template.

I have a pull request I will submit to display the desired results. If this is a breaking change I will make the original default value when not set and if a theme passes it a value it can set the other breakpoints.

The example is: When you log in as host, you have 3 default pages, this will make it break before those three pages plus maybe one more drop to a new line.

thabaum:
Fixes #3736

# Pull Request
The MenuHorizontal.razor theme UI component needs to allow more pages than 2 for a menu to even be worth having a horizontal menu for a feature by default.

Setting to the bootstrap `navbar-expand-lg` allows the menu to hold more page menu items prior to causing the menu items to fall into a new line making some page content less discoverable.

Also setting the CSS of the Oqtane Theme and Theme Creator Template to `@media (max-width: 991.98px) { ... }` will set the responsive behavior to the same as bootstrap defaults.

## Screenshots
This demonstrates the changes that this pull request introduces
image

image

This demonstrates the changes that this pull request introduces.

The MenuHorizontal.razor theme UI component needs to allow more pages than 2 for a menu to even be worth having this feature.

## Additional Context
We should consider introducing parameter to set this from a theme if this is a breaking change.

`<MenuHorizontal Breakpoint="LG" />`

And

`<Menu Orientation="Horizontal" Breakpoint="LG" />`

For examples.

Or if developers are expected to create their own UI controls for their themes this should not be a breaking change, otherwise it will be.

We can at least set the break points to 767.98px for the media max-width to match bootstrap in the Theme.css files so the navbar will respond more as expected.

### DIFF
--- a/Oqtane.Client/Themes/Controls/Theme/MenuHorizontal.razor
+++ b/Oqtane.Client/Themes/Controls/Theme/MenuHorizontal.razor
@@ -4,12 +4,12 @@
 
 @if (MenuPages.Any())
 {
-    <span class="app-menu-toggler navbar-expand-md">
+    <span class="app-menu-toggler navbar-expand-lg">
         <button type="button" class="navbar-toggler" data-bs-toggle="collapse" data-bs-target="#Menu" aria-controls="Menu" aria-expanded="false" aria-label="Toggle Navigation">
             <span class="navbar-toggler-icon"></span>
         </button>
     </span>
-    <div class="app-menu navbar-expand-md">
+    <div class="app-menu navbar-expand-lg">
         <div class="collapse navbar-collapse navbar-nav-scroll" id="Menu">
             <MenuItemsHorizontal ParentPage="null" Pages="MenuPages" />
         </div>
--- a/Oqtane.Server/wwwroot/Themes/Oqtane.Themes.OqtaneTheme/Theme.css
+++ b/Oqtane.Server/wwwroot/Themes/Oqtane.Themes.OqtaneTheme/Theme.css
@@ -96,7 +96,7 @@ div.app-moduleactions a.dropdown-toggle, div.app-moduleactions div.dropdown-menu
     z-index: 1000;
 }
 
-@media (max-width: 767px) {
+@media (max-width: 991.98px) {
 
     .app-menu {
         width: 100%;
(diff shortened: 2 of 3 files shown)


thabaum:
Closing since it would introduce a breaking change.

This should be an enhancement added to the theme.

PR #3751, opened 2024-02-09
specify default interactive render mode

sbwalker:


thabaum:
@sbwalker `@RenderMode.GetInteractiveRenderMode(PageState.Site.Runtime, PageState.Site.Prerender)`

Would you maybe need to encapsulate this ie `"@(RenderMode.GetInteractiveRenderMode(PageState.Site.Runtime, PageState.Site.Prerender))"?  Also maybe renaming the namespace for RenderMode to something like `ComponentRenderMode` may help keep a conflict of interest over that namespace from colliding between modules and components.

And/or try removing the `@` so it reads like documentation `@rendermode="RenderMode.InteractiveServer"`

Issue #3801, opened 2024-02-15
Missing/uninstalled theme leave the page not working and not editable


pyramidsbuilder:
Steps:
1- Install a theme
2- Assign the theme to a page
3- uninstall the theme
4- open the page

result:
Page does not load
Not able to edit the page to assign different theme either from page side panel or from Admin Dashboard

pyramidsbuilder:
PR #3812

pyramidsbuilder:
issue not happening anymore

PR #3802, opened 2024-02-15
Missing/uninstalled theme leave the page not working and not editable…

pyramidsbuilder:
… #3801

Changes made to App.razor to handle missing/deleted theme to use the first theme if missing
Changes made to edit.razor to avoid the error  of missing containers and load the edit form so user can select other theme and container


pyramidsbuilder:
I will reopen after removing the two added commits

PR #3803, opened 2024-02-15
create server-side SiteService

sbwalker:


sbwalker:
- 2 different implementations of ISiteService interface
- client implementation registered in client project Program.cs
- server implementation registered in server project Startup.cs
- logic from SiteController migrated to SiteService
- HttpContextAccessor used to obtain context in SiteService
- authorization attributes need to have equivalent IsInRole() calls added to SiteService methods to ensure consistent security
- HttpContext.Response.StatusCode logic needs to remain in Controller
- some Controller methods needs to refactored to integrate with results returned from SiteService
- caching migrated to SiteService so that calls originating on server will also get performance benefit
- SiteService uses Task.Run() to execute Repository methods - this is not optimal however there is significant effort to introduce a full set of new Repository async methods (the sync methods need to be retained as existing modules rely on them)

Issue #3794, opened 2024-02-14
Registration Emails Localization - Server app missing localization 


pyramidsbuilder:
Currently the texts on server app are not localized for example the email verification email content is hardcoded, 
I think using resx files won't be a good solution and the framework may need some new entity like "LocalizedContent" to be managed from the Dashboard (Language Manager?)
Any ideas?

sbwalker:
Personally I believe server-side resources should be localized using a consistent approach as the client-side. This will enable translators to localize the client and server using the same tooling and process. It will also be aligned with the current translation Nuget package distribution approach which allows for simple installation for administrators.

pyramidsbuilder:
@sbwalker,  Yes I think the same but sites may require "per site" templates, example saying "MySite Email Verification" 
I think this is more templating than localization.
I will do the resx files for now

thabaum:
Maybe the HTML editor can be incorporated to safely edit these resx entries as templates?

hishamco:
> I think using resx files won't be a good solution and the framework may need some new entity like "LocalizedContent" to be managed from the Dashboard (Language Manager?)

We need to differentiate between localization resources and content localization. IMHO we should use a standard ASP.NET way for localizing resources using `IStringLocalizer` I know the `ResourceManager` is the default implementation which is fine for static resources, but might not be good for some corner cases such as Pluralization

Let me know if you need any help on this also I might create a PR for it

sbwalker:
@hishamco you are correct that content localization is a different scenario than localized resources. Content localization is not on the product roadmap at this stage - there are other items which are higher priority.

hishamco:
Hope to see any localization related feature ASAP :)

sbwalker:
merged

PR #3821, opened 2024-02-16
localized usermanager email messages and message formatting to includ…

pyramidsbuilder:
…e site name and link #3794

### DIFF
--- a/Oqtane.Server/Managers/UserManager.cs
+++ b/Oqtane.Server/Managers/UserManager.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Identity;
+using Microsoft.Extensions.Localization;
 using Oqtane.Enums;
 using Oqtane.Infrastructure;
 using Oqtane.Models;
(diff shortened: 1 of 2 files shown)


pyramidsbuilder:
image

image

image

image

sbwalker:
@pyramidsbuilder this is the correct approach to take for localizing resources. One change I would make for consistency with the Client project is that the folder structure within the Resources folder should match the folder structure in the Server project (rather than using "." notation). 

Adding localization support for Server resources has been discussed previously, as it was known that this would be needed in the future. However there would be impacts to how translators are managing resources today. Specifically, when they build the satellite assemblies and package the them, they would need to include both the Client and Server DLLs. I would like to get some feedback from @leigh-pointer on this as he is the maintainer of the nl-NL translation which has been kept up to date with every Oqtane release (whereas a lot of the resources for other languages in the https://github.com/oqtane/oqtane.translations repo are now way out of date).

leigh-pointer:
@sbwalker Just wanted to share my thoughts on the impact of this managing resources task. While I don't see it being too tricky to translate files, it might add a bit more to our workload. However, I think it'd be really cool to consider localizing all string values if possible. It could make things more accessible and user-friendly for everyone involved!

PR #3822, opened 2024-02-16
fix account lockout to read from correct setting, consider timezone o…

pyramidsbuilder:
fix account lockout to read from correct setting, consider timezone offset and enable lockout


pyramidsbuilder:
image

image

image

image

Issue #3839, opened 2024-02-19
Bug: Page Management - Add/Edit Page - Switch Theme to Blazor Exception Thrown. (5.1.0)


thabaum:
### Description

While using the Oqtane Theme changing the theme on a page (edit page or add page component forms) to Blazor Theme generates a Visual Studio error as shown below in the event log and screenshots.
[x] Tested - Issue Does Not Exist Using Blazor Theme for Site Settings
### Event Log Error
```
An Unexpected Error Has Occurred In Oqtane.Modules.Admin.Pages, Oqtane.Client: The component type must implement Microsoft.AspNetCore.Components.IComponent.

System.ArgumentException: The component type must implement Microsoft.AspNetCore.Components.IComponent.
   at Microsoft.AspNetCore.Components.Rendering.RenderTreeBuilder.OpenComponent(Int32 sequence, Type componentType)
   at Oqtane.Modules.Admin.Pages.Add.<ThemeSettings>b__42_1(RenderTreeBuilder builder) in C:\Dev\Oqtane\Dev\5.1.0\021824\Oqtane.Client\Modules\Admin\Pages\Add.razor:line 348
   at Microsoft.AspNetCore.Components.Rendering.RenderTreeBuilder.AddContent(Int32 sequence, RenderFragment fragment)
   at Oqtane.Modules.Admin.Pages.Add.<BuildRenderTree>b__0_4(RenderTreeBuilder __builder3) in C:\Dev\Oqtane\Dev\5.1.0\021824\Oqtane.Client\Modules\Admin\Pages\Add.razor:line 204
   at Microsoft.AspNetCore.Components.Rendering.RenderTreeBuilder.AddContent(Int32 sequence, RenderFragment fragment)
   at Oqtane.Modules.Controls.TabPanel.BuildRenderTree(RenderTreeBuilder __builder) in C:\Dev\Oqtane\Dev\5.1.0\021824\Oqtane.Client\Modules\Controls\TabPanel.razor:line 13
   at Microsoft.AspNetCore.Components.Rendering.ComponentState.RenderIntoBatch(RenderBatchBuilder batchBuilder, RenderFragment renderFragment, Exception& renderFragmentException)
   
   {"ModuleDefinitionName":"Oqtane.Modules.Admin.Pages, Oqtane.Client","Error":"The component type must implement Microsoft.AspNetCore.Components.IComponent."}
```
### Screenshots
[...]

PR #3849, opened 2024-02-20
#3839 handle missing theme to site default theme and oqtane default theme

pyramidsbuilder:
…theme
#3839

### DIFF
--- a/Oqtane.Client/UI/ThemeBuilder.razor
+++ b/Oqtane.Client/UI/ThemeBuilder.razor
@@ -64,9 +64,15 @@
         // send to Head component
         SiteState.Properties.HeadContent = headcontent;
 
+        var typeString = PageState.Page.ThemeType;
+        if (string.IsNullOrEmpty(typeString))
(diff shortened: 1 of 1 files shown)


sbwalker:
@pyramidsbuilder I resolved #3839 with PR #3855 . I am not sure how to reproduce the problem which this PR is intended to address?

pyramidsbuilder:
@sbwalker I get the error in issue #3839 when the Page does not have a themetype defined (empty string), (probably the system won't let this data inconsistency) and this reason is different although firing same error.

This PR load the theme from the site default theme type when it is not defined in page, and when the site does not have a theme type it provides a hardcoded default Oqtane theme type to make sure the page renders

it can be closed if this is not an issue

sbwalker:
@pyramidsbuilder I believe this issue was caused by the fact that the dev branch was still being migrated to Blazor Web when you were working with the code base. I will close this PR as I do not think it is relevant now that the migration is nearly complete.

PR #3856, opened 2024-02-20
fix #3838 - Schedule Jobs looping after new install

### LINKED ISSUE #3838: Bug: Admin - New Installs, Scheduled Jobs Will Not Load Default Installed Jobs On First Visit Flashing A Notification To Restart Stuck In A Loop (5.1.0)
### Description
**Scheduled Jobs** component in the **Control Panel > Admin Dashboard** has a notification that flashes repeatedly by clearing itself and then appearing again in a loop and does not stop until navigating to a new page and also the page needs to be refreshed as it will be keeping the output logs in VS busy until you do.  This will make the **Control Panel > Admin Dashboard** UI button control inactive as well.

https://github.com/oqtane/oqtane.framework/blob/dev/Oqtane.Client/Modules/Admin/Jobs/Index.razor

### Screenshots

image

image
image

image

### Additional Context
I have also noticed the double flashes (2 page UI reloads/refreshes) within the pages for static SSR and wonder if some caching or something needs to come into play to smooth that out to feel like it is one page load?

I just happen to be clicking through all of the admin pages today to discover this issue as I looked to see if any admin pages didn't double flash in Visual Studio IIS Express.  Currently this is the only page with this odd of a behavior with this notification loop.

More info...

After investigating more it went away, so I almost closed the issue.  But then I noticed how there where 2 scheduled jobs, and 0 showing before.

So after you stop the site in visual studio, or restart the app (only stable way is with stopping VS) then starting it again the problem goes away.

This was seen in all render modes.  It is a fresh install issue only on first visit, then restart and it goes away displaying the two jobs shown below.

image

Out of all pages tested located in the admin control panel dashboard parent pages it felt like admin/system/Index.razor or "System Info" loaded the quickest without any double flashing going on, just the form and UI is loaded, then data in the form comes later as you would expect.  Site Settings will load the entire UI twice compared to System Info, or at least it feels this way.

sbwalker:


thabaum:
This does fix #3838, however maybe we need to log another issue relating to deleting the scheduled jobs maybe should not be allowed (delete button inactive) since once you uninstall them, the next time you restart Oqtane these scheduled jobs are installed again.  Or would these jobs being added again if not present be the expected behavior?  

I suppose I am inquiring about the current behavior, to review if another issue is present or not.  If these two scheduled jobs are expected to always be installed then moving on, unless they should be not meant to be deleted/reinstalled.

sbwalker:
@thabaum you are correct, there should not be a delete button (similar to the fact that there is no Add button). Scheduled jobs are automatically added on startup if the assembly exists in the /bin.

sbwalker:
@zyhfish it should not occur in other components if those components are developed correctly.

zyhfish:
Thanks @sbwalker , these controls should need to be updated as they called AddModuleMessage in OnParametersSet method:
Oqtane.Client\Modules\Admin\Files\Index.razor
Oqtane.Client\Modules\Admin\Logs\Index.razor
Oqtane.Client\Modules\Admin\ModuleDefinitions\Index.razor
Oqtane.Client\Modules\Admin\Themes\Index.razor
Oqtane.Client\Modules\HtmlText\Index.razor

I will try to analyze the possible calls to StateHasChanged in OnParametersSet by any other functions to check whether has other controls need to be updated, then submit a PR to handle them.
Thanks,

sbwalker:
@zyhfish the components you identified are indeed problematic and could result in infinite loops if an exception occurred in the OnParametersSet method:

ie.

```
    protected override async Task OnParametersSetAsync()
    {
        try
        {
            // do some stuff
        }
        catch (Exception ex)
        {
            await logger.LogError(ex, "Error {Error}", ex.Message);
            AddModuleMessage("Error", MessageType.Error); // this will call OnParametersSet resulting in an infinite loop
        }
    }
```

Most likely the best approach is to remove the AddModuleMessage() method call from the exception handler in these methods. The user will not get any feedback that an error occurred - but it will prevent an infinite loop. Inifinite loops are especially problematic in this case because they logger.LogError() generates Notifications which are sent by email. So your email box could get spammed with hundreds of messages as a result of an infinite loop.

sbwalker:
@zyhfish PR #3875 resolves the issues in the components you identified

PR #3860, opened 2024-02-21
remove Delete option from Scheduled Jobs - they are automatically created at startup

sbwalker:


thabaum:
This is fixed but I was confused at first as there were no jobs in the Scheduled Jobs on a new installation until after you restart the application.  Could there be a flag for first time running as it says to restart to install these.  So the page itself is executing the logic somehow to install these scheduled jobs?  I would think they would already be installed and show up when you first visit this page.  Maybe another related issue to review?

sbwalker:
On a new install the database does not exist at the point in time when the scheduled jobs startup and try to register themselves. This is why a restart is necessary for the jobs to be registered. The message displayed is intended to warn a user about this situation. I believe the existing message is clear enough.

PR #3865, opened 2024-02-21
Fixes #3828 - Adds page refresh logic in SavePage() method.

### LINKED ISSUE #3828: Bug: Saving Page Settings After Editing Does Not Always Display Updates in the UI
### Description
Saving Page Settings After Editing Does Not Always Display Updates in the UI

### Steps to Reproduce
1. Login as administrator and add a new page to the navigation menu root (or child too I believe).
2. After creating the new page you are navigated to the page, click on the gear icon and edit page button.
3. Select an icon from the available icons.  You will notice them update immediately while reviewing them.
4. Save the page using the button on the page.
5. Notice the page is created and shown in the UI, however the icon is not shown.  The page has to be refreshed (F5) or navigate away from the page to see the update in the navigation menu.

Pretty much create a new page without an icon, edit the page adding an icon then click save in order to notice the UI is not updated always displaying the change in icon.

### Expected Behavior
The UI is reloaded/refreshed to immediately show this has been updated on the page menu when an admin saves after editing page settings for things like this.

### Additional Context
This could be done in a few ways.  I have a PR that resolves this issue which I hope is acceptable or one like it is used to replace it.  More conditional logic can be used but a simple true reload boolean fixes this I believe and is what the pull request will introduce as a possible solution.

This issue also happens when you update the Site Name in Site Settings.  You must refresh the page to see changes, unlike if you save after changing the logo which you now see immediately.  Both areas Admin Dashboard > Site Settings and Control Panel > Page Edit save actions can use an enhanced user experience by updating the UI when these types of changes to settings are saved.

thabaum:
## Pull Request

Fixes #3828

### Description

If conditional check is preferred this pull request would be in favor of #3829 to close #3828 with logic similar to site settings which includes a conditional check prior to doing a full UI page refresh to update the UI when the page name, title, or icon has been changed.

### Additional Context

I see `private bool _refresh = false;` which maybe should be used instead of introducing another variable. I went ahead and removed the new `refresh` variable as it seems to work using the existing `_refresh` boolean with a new commit to this pull request.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Pages/Edit.razor
+++ b/Oqtane.Client/Modules/Admin/Pages/Edit.razor
@@ -499,7 +499,6 @@
     private async Task SavePage()
     {
         validated = true;
-        var refresh = false;
         var interop = new Interop(JSRuntime);
         if (await interop.FormValid(form))
         {
@@ -512,10 +511,10 @@
                 }
                 if (!string.IsNullOrEmpty(_themetype) && _containertype != "-")
                 {
-					// check if a page refresh is needed
+					// ensure page refreshes if these have been changed
 					if (_page.Name != _name || _page.Icon != _icon || _page.Title != _title)
 					{
-						refresh = true;
+						_refresh = true;
 					}
 
                     string currentPath = _page.Path;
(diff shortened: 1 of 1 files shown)


sbwalker:
@thabaum in Site Settings there was logic included during save to indicate which fields required a full page refresh and which fields did not. This efficiency was more important in a SPA/interactive scenario as a full page refresh is an expensive operation as it need to reload all of the state. However now that Oqtane needs to support both static and interactive scenarios, it probably makes sense to simplify this area and just do a full page refresh in all cases. This will satisfy both rendering modes and make the code easier to maintain in the future.

thabaum:
@sbwalker thanks for the clarification on the updates expected.  I will close this PR in favor of the always refresh on save version.

PR #3881, opened 2024-02-23
fix #3868 - static rendering support for page themes within site

### LINKED ISSUE #3868: Bug: Page Refresh Needed To View Different Themes On Different Pages (Static Server Rendering - 5.1.0)
### Description
After changing the theme on a page, the user is not shown the new page theme or site theme, it will be one or the other like it is cached.

### Steps to reproduce
1. Install Two themes
2. Set a page to use one theme.
3. Switch between pages in the navigation menu
4. notice the theme behavior described and shown in this issue.

Below are screenshots of the behavior.

I am using two themes, the Oqtane Default Theme and Oqtane Theme Creator Template Theme.

My Page has the theme creator template theme, while the rest should use the stock default Oqtane theme.

### Screenshots

image

image

image

Hit refresh:
image
image
hit refresh:
image

The theme is not updated with current theme provided by the page.

sbwalker:


thabaum:
@sbwalker I had even thought it was something in the navigation menu causing the bug presented by the screenshots in the relating issue. I found it odd the theme actually changed but the css at after the menu loads did not. Nice find, who would of ever thought!

PR #3890, opened 2024-02-24
added property Platform to sitestate to detect maui platform

pyramidsbuilder:
To detect the platform (ex: Android) in Maui rendering, and be able to do adjustments in Theme/Module:

added property platform to sitestate
pass the platform name from Maui
Use the platform name for adjustments in Theme/Module

### DIFF
--- a/Oqtane.Client/UI/Head.razor
+++ b/Oqtane.Client/UI/Head.razor
@@ -23,11 +23,19 @@
     public string RenderMode { get; set; }
 
     [Parameter]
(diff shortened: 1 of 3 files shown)


sbwalker:
@pyramidsbuilder I do not understand the benefit of this enhancement. Adding this property to SiteState basically means that a module would need to inject the SiteState service into their component in  order to reference the property value - which is more work than simply calling  DeviceInfo.Current.Platform.ToString() from the component directly?

PR #3916, opened 2024-02-28
change Admin Dashboard to use static rendering and fix invalid markup (H2 within P element)

sbwalker:


thabaum (commented):
logic may need to be included with a conditional statement or may break a UI that has custom CSS and themes that might target selecting this specific element.

sbwalker:
@thabaum This component was using h2 elements within p elements - which is invalid according to the HTML spec and was causing rendering issues on static rendering.

thabaum:
The change is OK and I understand. I just wanted to be aware someone could have a UI break on upgrade, but I see it as a necessary breaking change if so as it will need to be updated to be correct.

Changing <p> to <div> will be a potential breaking change for a theme that uses this.  But nothing can do as it would make it harder for theme developers to make themes and the trade-off is not worth it.

sbwalker:
@thabaum the strange part is that this component worked fine on Interactive rendering - but choked on Static rendering. This leads me to conclude that the "enhanced navigation" DOM patching algorithm is not the same as the Interactive "shadow DOM" algorithm.

PR #3925, opened 2024-02-29
added exception middleware to server for logging

pyramidsbuilder:
added exception middlware to log server exceptions

### DIFF
--- a/Oqtane.Server/Extensions/ApplicationBuilderExtensions.cs
+++ b/Oqtane.Server/Extensions/ApplicationBuilderExtensions.cs
@@ -44,5 +44,9 @@ namespace Oqtane.Extensions
 
         public static IApplicationBuilder UseJwtAuthorization(this IApplicationBuilder builder)
           => builder.UseMiddleware<JwtMiddleware>();
+
+        public static IApplicationBuilder UseExceptionMiddleWare(this IApplicationBuilder builder)
+          => builder.UseMiddleware<ExceptionMiddleware>();
+
     }
 }
--- a/Oqtane.Server/Infrastructure/Middleware/ExceptionMiddleware.cs
+++ b/Oqtane.Server/Infrastructure/Middleware/ExceptionMiddleware.cs
@@ -1,0 +1,45 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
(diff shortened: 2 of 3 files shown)


sbwalker:
Oqtane already has database logging for all scenarios where a tenant can be identified... and file logging for scenarios where it cannot. This PR adds more logging on top of the existing logging - which does not make sense?

pyramidsbuilder:
@sbwalker  Oqtane has logging which I used in the middleware, 
In my case I was writing module and when the module throws an exception, it is not logged.
unless the module try/catch exceptions on the "methods" level and loggs it, you cannot get information about the error happened.
The middleware handles any exception happens on server app and log it.
On client side it is already there, it uses OnError and catch/log client exceptions by default, but on server side it is missing this feature

sbwalker:
Unhandled exceptions in module components are supposed to be caught by the ErrorBoundary which is in Oqtane.Client\UI\RenderModeBoundary.razor. The ErrorBoundary logs the detailed error and displays a friendly error message to the UI. So I am curious why your module components would not function properly. Are the module components specifying public override string RenderMode => RenderModes.Static... or are they Interactive (default)?

pyramidsbuilder:
@sbwalker I have the client error logging working but it is not a detailed messages it looks like this
`Request /api/EndPoint?moduleid=1&authmoduleid=1 Failed With Status 500 - Internal Server Error`

while I need the actual error stack trace and line number which the exception middleware can provide, for example:

```
System.InvalidCastException: Unable to cast object of type 'System.String' to type 'System.Guid'.
   at Microsoft.Data.SqlClient.SqlBuffer.get_Guid()
***************************
*********************************************\Server\Repository\Repository.cs:line 23
   
****************************Server\Controllers\Controller.cs:line 57

```

sbwalker:
Oh ok... so this is intended for server-side logging of unhandled exceptions. Did you look in the Oqtane.Server\Content\Log\error.log file to determine if these errors were logged there? This error log is also displayed in the UI when you go to Admin Dashboard / System Info and choose the Log tab.

pyramidsbuilder:
@sbwalker  yes it is logged there, thanks for details, it is not necessary then

sbwalker:
@pyramidsbuilder I actually think there is value in this PR, as in those scenarios where the tenant can be determined, it can log the exception to the appropriate database (tenant) for the correct site. This keeps the exceptions grouped together and much easier to locate. And in the cases where the tenant cannot be determined, it can fallback to use the file logger.

PR #3931, opened 2024-03-02
 Improved Pager Search Input Text Field Autocomplete Behavior

leigh-pointer:
This pull request enhances the behavior of the search input text field by disabling the browser’s autocomplete feature. Here’s why I’ve opted for the <form> tag approach:

**Form-Level Control:**
By placing autocomplete="off" on the <form> tag, we ensure consistent behavior across all input fields within the form. Although only one field, adding it to the Input control gives strange effects across browsers.

**Edge Compatibility:**
As mentioned earlier, Edge 18 has limited support for autocomplete="off".
By setting it at the form level, we address this issue globally for all browsers, including Edge.

**Why This Matters:**
Improved user experience: Users won’t be distracted by irrelevant autofill suggestions.
Consistency: The search input field now behaves uniformly across browsers.

**Note for Edge Users:**
While autocomplete="off" works as expected in Chrome and Firefox, it may have limited support in Edge (specifically Edge 18). Users of Edge 18 are advised to disable autofill from the browser settings directly.

### DIFF
--- a/Oqtane.Client/Modules/Controls/Pager.razor
+++ b/Oqtane.Client/Modules/Controls/Pager.razor
@@ -8,13 +8,13 @@
 {
     @if (!string.IsNullOrEmpty(SearchProperties))
     {
-        @* <form autocomplete="off"> *@
+        <form autocomplete="off">
             <div class="input-group my-3">
-                <input type="text" autocomplete="off" id="search" class="form-control" placeholder=@string.Format(Localizer["SearchPlaceholder"], FormatSearchProperties()) @bind="@_search" />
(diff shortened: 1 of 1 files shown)


sbwalker:
@leigh-pointer I suspect the other problem is that the id of the input element is "search" - which is very generic and since browsers use the id/name as the key for form autocomplete it increases the probability of it relating the wrong information to this element. If we changed the id to "pagersearch" it would be more unique and would probably cause less problems.

sbwalker:
@leigh-pointer we can also add the autocomplete attribute as it will help in some browsers

PR #3938, opened 2024-03-03
Fix #3927: change to async method to avoid the exception.

### LINKED ISSUE #3927: Closing Module Management 5.1.0 - Create Module Dialog using the [x]
Created a new module. 
Clicked the [x]  code was halted.

CreateModuleDialog

zyhfish:


sbwalker:
@zyhfish this does not resolve the issue. I enabled the "break on exception" for NavigationManager (which I had previously disabled) and the exception still occurs:

image

sbwalker:
I am going to roll back this PR as it causes a compilation warning as well (ie. "his async method lacks 'await' operators and will run synchronously")

zyhfish:
Hi @sbwalker , this is weird, I verified the changes in my local and it works fine:
here is the screencast of how it works on my local: 
https://drive.google.com/file/d/179nLo2bW6A3CHtVov0Ez5II00NF0R0Wn/view?usp=sharing

sbwalker:
In your screencast in the exception settings, it looks different than mine. The default Visual Studio setting is to have "Break when this exception type is thrown" (like my screen shot above). In yours, this option is not selected - instead you have "break when the exception is user handled" - which I believe is responsible for the different behavior.

image

zyhfish:
Hi @sbwalker , thx so much for you checking, let me try to check whether can find another possible way to avoid this behavior.

sbwalker:
Note that in the Microsoft issue I also provided some code which is in the Identity bits added to a default .NET 8 Blazor Web app... which appears to provide a workaround for this bug:

https://github.com/dotnet/aspnetcore/issues/53996#issuecomment-1971734784

JPScerri:
> @zyhfish Note that in the Microsoft issue I also provided some code which is in the Identity bits added to a default .NET 8 Blazor Web app... which appears to provide a workaround for this bug:
> 
> [dotnet/aspnetcore#53996 (comment)](https://github.com/dotnet/aspnetcore/issues/53996#issuecomment-1971734784)
> 
> Note this this exception does not happen in production - it only happens in Visual Studio when running a debugger - so although it is annoying for developers, there is a workaround by modifying the exception settings - and it does not affect production apps.

I got the same issue. If I ignore this error, when I publish in Azure, the server is throwing Internal Error.!!

PR #3981, opened 2024-03-12
Update Module Template IServerStartup to be ServerStartup instead of DbContextFactory

mdmontesinos:
This naming convention better aligns with current implementation (HtmlText module) and is less confusing.

I'm not sure if ServerStartup is better than [Module]ServerStartup.

Related to #3971


mdmontesinos:
@dotnet-policy-service agree

PR #3988, opened 2024-03-14
Fix #3977: force reload the page after update the settings.

### LINKED ISSUE #3977: Bug: Updating Module Title Twice Does Not Save Title Changes - 5.1.0
# ISSUE

## Description

The module title is not being updated once you set it, and try to reset it, without refreshing the page to get a fresh UI instance.

##  Steps to Reproduce...

1.  Edit page module title in page settings
2. Save changes.
3. While still on that page and most likely still in edit mode try to edit and save the title again
4. Notice the title is still the same and does not get updated.

Happens both static and interactive Oqtane render modes.

I noticed in static at least that the carrot icon becomes inactive as well at times.

image

Not sure if the above issue with inactive dropdown actions is limited to my installation. It is something to be aware of while testing solutions.

zyhfish:


thabaum:
this does resolve the issue, I had this exact fix, then tested one more time against the new branch and it worked I could not reproduce and adding or removing that had no difference.  I did not try testing in Interactive mode... :)  Thanks @zyhfish

sbwalker:
@zyhfish I am unable to reproduce the issue reported in #3977 in either Static or Interactive render modes. So I am not going to merge this PR unless there is a way to reliably reproduce the problem. It is possible that another PR which has been merged since has already addressed the issue.

zyhfish:
Hi @sbwalker , it's easily to be reproduced and I have identified the root cause of this problem, we can setup a meeting to check the issue together if you want. I submitted PR #4006 to fix this issue, I will close this PR as this is just a temp solution.

PR #3997, opened 2024-03-15
Fix #3990: sync html content when switch the tab.

### LINKED ISSUE #3990: Enh: HTML/Text Editor Auto-Synchronize Content When Switching Between Rich Text and Raw HTML Editors
# Enhancement Request

## Description

Enhance the HTML/Text Editor to auto-synchronize content when switching between the `Rich Text Editor` and `Raw HTML Editor`.

## Solution Proposed

When a user clicks a tab for the different editors, the synchronize content action takes place if content has changed.

## Screenshots

image

#### Additional Info

It is nice to at least have the button to be able to ensure changes are displayed.  I was just curious if we can make this automatic when switching tabs, synchronize content.

zyhfish:


thabaum:
@zyhfish Thank you for this contribution, I just tested this and seems to work as expected in both static and interactive render modes.

Looks like this will also include an enhancement to the `TabPanel` and `TabStrip` UI components adding a callback parameter for the active tab.

Nice work!

thabaum:
@zyhfish hello, thanks again for this contribution.  After further review a couple things come to mind:

1. Is there a reason to create the extra methods and not just call them directly?  Seems to work the same instead of using a method to call another method.
 
2. I have added logic for a reset button I will share below maybe we can add this to the PR or if @sbwalker prefers in another PR after this one is merged.

## RichTextEditor.razor
Line 12 about: ` <TabPanel Name="Rich" Heading="Rich Text Editor" ResourceKey="RichTextEditor" OnActivePanel="RefreshRichText">`
line 22 about: 
```
                        @if (AllowRawHtml)
                        {
							<button type="button" class="btn btn-secondary" @onclick="ResetContent">@Localizer["ResetContent"]</button>
							@((MarkupString)"&nbsp;&nbsp;")

							<button type="button" class="btn btn-secondary" @onclick="RefreshRichText">@Localizer["SynchronizeContent"]</button>
                            @((MarkupString)"&nbsp;&nbsp;")
                        }
```

same in next tab:

`<TabPanel Name="Raw" Heading="Raw HTML Editor" ResourceKey="HtmlEditor" OnActivePanel="async () => await RefreshRawHtml()">`

and this button added as well need the resource file to include with.
```
						<button type="button" class="btn btn-secondary" @onclick="ResetContent">@Localizer["ResetContent"]</button>&nbsp;&nbsp;
						<button type="button" class="btn btn-secondary" @onclick="RefreshRawHtml">@Localizer["SynchronizeContent"]</button>&nbsp;&nbsp;
```



adding this function at the bottom of the` @code`

```
	public async Task ResetContent()
	{
		_richhtml = _originalrichhtml;
		_rawhtml = _originalrawhtml;

		var interop = new RichTextEditorInterop(JSRuntime);
		await interop.LoadEditorContent(_editorElement, _richhtml);
		StateHasChanged();
	}
```

So depending on how @sbwalker reviews this I will wait if I should add an enhancement request as well?  They somewhat relate...  I have to wait for this one to get through to propose the enhancement is why I ask how we can proceed.  I am OK with what gets it done if it is a solution that is acceptable or something similar.

zyhfish:
Hi @thabaum ,
thanks so much for your update, I guess one PR should resolve one problem, this will be easier to manage the changes.

the reason of create a new method wrapper is to make the code easier to read, also can add more functions in future if needed.

sbwalker:
@zyhfish I am sure this PR does allow a user to switch tabs and have the content auto synchronized - however I am not in favor of this enhancement because it will cause problems with HTML content. I have received the issue here:

https://github.com/oqtane/oqtane.framework/issues/3990#issuecomment-2000829913

My suggestion is to remove all synchronization from the editor as it can cause a poor user experience and in a worst case, a loss of content.

zyhfish:
Hi @sbwalker , thanks so much for the update but I still confused. is this something need to be fixed in the rich text editor side?

such as after I edit the content in raw editor, then I will not able to use the rich text editor to edit the content anymore as it will display blank or missing content. in this case the edit experience will be bad as user don't know how to keep the correct content next time, as he can't remember exactly which type editor was used last time.

btw, I just tried this scenario but can't reproduce it with below steps:
1. edit a html module, switch to raw editor, change the content to "<p>Hello World</p>", save the content.
2. edit the module again, the content display in the rich text editor is correctly.

sbwalker:
@zyhfish it is a problem with specific HTML elements. For example, if you specify ```<div>Hello World</div>``` I believe it will remove the content. And there are many other types of HTML it will modify/remove as well. So I do NOT approve the enhancement suggested by @thabaum (#3990) and suggest removing the Synchronize buttons and logic instead as alternative.

PR #4013, opened 2024-03-18
Fixes 4012: Quill bootstrap theme support

### LINKED ISSUE #4012: [BUG] HTML Module - Rich Text Editor Does Not Display Default Site Template Theme Styles
# Issue

## Description

As shown below the HTML Text Editor Does Not Display Theme Styles.

## Screenshots

image

thabaum:
# Pull Request

Fixes #4012

## Description
 
Adds Quill Bootstrap theme support to all bootstrap themes.

## Screenshots

image

## Additional Information

This PR demonstrates using Bootstrap theme variables to help style the Quill rich text editor component into a Bootstrap enabled theme like the default themes used to showcase the Oqtane Framework.  Switch these variables to your flavor, and away you go 🚀    

_Note:  The `!important` is necessary to override the module loading after the theme._

### DIFF
--- a/Oqtane.Server/wwwroot/Themes/Oqtane.Themes.BlazorTheme/Theme.css
+++ b/Oqtane.Server/wwwroot/Themes/Oqtane.Themes.BlazorTheme/Theme.css
@@ -117,6 +117,35 @@
     margin: .5rem;
 }
 
+/* Oqtane Quill Rich Text Editor Toolbar */
+.ql-editor, .ql-toolbar, .ql-picker-options, .ql-picker-label {
+    background-color: var(--bs-body-bg) !important;
+    color: var(--bs-body-color) !important;
+}
+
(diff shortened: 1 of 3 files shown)


thabaum:
@sbwalker I have a few more enhancements here to make I noticed for example with the heading CSS I am working on resolving.  I will keep this a draft to let others know this is being worked on.  I will let you know when ready for review shortly. Thank you.

sbwalker:
@thabaum you will have to explain how this PR is intended to work with the existing Quill Theme support:

RichTextEditor.razor has a parameter which defines the Quill Theme which should be used:

```
    [Parameter]
    public string Theme { get; set; } = "snow";
```

And the Html/Text Module (Oqtane.Client\Modules\HtmlText\Edit.razor) defines the CSS resources related to the Theme. It is declared this way so that the CSS resources are only injected into the page if the Quill editor is being used on that page (ie. for performance reasons it loads resources based on actual usage).

```
	public override List<Resource> Resources => new List<Resource>()
	{
        new Resource { ResourceType = ResourceType.Stylesheet, Url = "css/quill/quill.bubble.css" },
        new Resource { ResourceType = ResourceType.Stylesheet, Url = "css/quill/quill.snow.css" }
    };
```

thabaum:
@sbwalker To ensure flexibility across different themes without being tied to a specific selector like ".snow", I've refrained from using such selectors in the styling. This approach allows any theme to be selected later, with the priority being given to the styling consistency of the theme used in Oqtane.

The options considered include placing the styling in either the module's CSS (module.css) or the theme's CSS (theme.css). Placing it in theme.css seems more logical, especially considering that the styling might vary significantly if someone opts for a different theme such as Material.

While I acknowledge the potential performance impact, there's also a significant user interface (UI) cost involved. The current UI styling doesn't align with the intended style, making it necessary to provide specific styling for the HTML editor.

If there's a way to selectively include the HTML styles based on the theme being used, I'm open to exploring that option. One possibility could be to import the necessary styles into the module.css file for the HTML module.  Issue is where do we store this in a theme for that module to pickup?  Another option is an editor for CSS to extend modules abilities to override it's CSS.

While loading additional CSS on each page load may raise concerns about performance, the amount of CSS added to adjust the settings is relatively small. Given the current situation where button icons are barely visible and the background and colors don't match well, it's essential to address these styling issues for better usability.

In summary, incorporating the HTML module styling into the theme's CSS, a way to import into module CSS a theme file or providing a module CSS editor could help address these concerns while maintaining styling consistency across different themes.

One last idea I had was creating a theme variables which can get set throughout modules and themes and imported then used.   This way a theme can adjust these and all modules that need some styling love can get it.

I knew all of this was going to get discussed, so this is what I had already thought of.  I am just prepping some CSS for the final game plan if any...

thabaum:
@sbwalker Thanks again for the feedback and consideration for the future roadmap.  I will continue to improve this solution for Bootstrap theme's used for Oqtane Framework's demonstration, sorry I need a placeholder for it and feedback while exploring options.  

I have 2  3 issues to work out with the toolbar currently, it's 95% ready.

The editor content area only needed a few styles which are at the bottom.  take away the two toolbar icon svg stroke styles and this shrinks it a lot since those select many items.  I was going to see how I can minimize this.  I removed the `.snow` so it works with any theme.  I want to test the bare theme out if possible.

I was trying to figure out how to accomplish this task and found this difficult to come up with a sound solution, after weighing everything I thought this was a great starting point to review.

An enhancement would be generate CSS settings per site/theme/page/module with an editor.  Module I thought can have a setting of "per instance" or "global" for custom CSS.

Since the HTML module is part of the framework, it needs to be able to be styled properly.

Thanks again for time reviewing, I will be updating this soon with all issues fixed in the toolbar area along with any ways I can minimize it's size.  Looking forward to 5.1 issues being resolved.  I will keep contributing to them as I can.

Cheers!

thabaum:
closing in support of 4022 as it has much smaller footprint only changing the theme.css for Oqtane Theme, so it is not going to follow the module creator, but if they want to they can reviewing Oqtane Theme more and put their own twist to it.  Light themes do not really need any help...

The ability to add css to a module and module instance would be a great enhancement that maybe needs suggested.

PR #4022, opened 2024-03-19
Fixes #4012: Adds Quill Editor Bootstrap CSS Styles To The Oqtane Default Theme

### LINKED ISSUE #4012: [BUG] HTML Module - Rich Text Editor Does Not Display Default Site Template Theme Styles

thabaum:
# Pull Request

Fixes #4012

## Description

This is to add bootstrap theme support to the Oqtane theme.  This is a minimized version allowing Quill themes to rule the toolbar area while styling the content editing areas.

## Screenshots

Current (Latest)
image
image
image
image

Without dark toolbar as I preferred with it:
image

image

## Additional Information
More styling can be added to help mimic the Oqtane theme.  More solutions can be made down the road.  This makes the buttons and information in content areas and toolbar match the Oqtane Theme for better user experience.  It is a lot less CSS than previous.  

The `.tooltip` can be removed along with some other styles if necessary.

If desired I can make the `.toolbar` background darker as well so you can se the icons better.  Maybe `--bs-modal-bg` ? like the `tooltip`.  But this is very similar to what is already there if not the same...

Adding `.toolbar` class selector background will make it dark like the theme is as shown below:
```
.ql-editor, .ql-header .ql-picker-options, .ql-toolbar {
    background-color: var(--bs-body-bg) !important;
    color: var(--bs-body-color) !important;
}
```

image

I am going to close #4013 which also has relating conversation details from previous feedback this PR should be more acceptable near term with less overhead.

In summary from the last PR this is on the back burner on the road map for after Oqtane Framework version 5.1 is stabilized and released.

### DIFF
--- a/Oqtane.Server/wwwroot/Themes/Oqtane.Themes.OqtaneTheme/Theme.css
+++ b/Oqtane.Server/wwwroot/Themes/Oqtane.Themes.OqtaneTheme/Theme.css
@@ -106,6 +106,10 @@ div.app-moduleactions a.dropdown-toggle, div.app-moduleactions div.dropdown-menu
     color: var(--bs-heading-color) !important;
 }
 
+    .ql-header .ql-picker-options .ql-picker-item:first-child {
+        color: var(--bs-body-color) !important;
+    }
+
 .ql-editor a {
     color: var(--bs-link-color) !important;
 }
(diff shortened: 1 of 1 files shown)


thabaum:
@sbwalker I really like this in it's current state for the Oqtane Theme when ready to review it has been lightened up and fine tuned enough to keep settlers happy.  This is only for the Oqtane theme, if anyone even uses it, they can build a theme without as it only is really needed for this Cyborg theme currently as it is a dark theme, light themes this works great as is... but the out of the box UX makes it worth adding this css enhancement.  I wont spend much more on this until this issue gets re-visted.  Thank you.

thabaum:
Closing due to potential new upcoming Oqtane Radzen UI Component HTML Editor proposed at developer meeting with a POC to be revealed by @ADefWebserver at the next meeting.

Cheers!

🚀

PR #4027, opened 2024-03-20
Fix #4025: add user setting to keep the view mode.

### LINKED ISSUE #4025: Bug: HTML/Text Editor Save, Cancel and Close do not return user in edit mode.
# Issue

## Description
HTML/Text Editor Save, Cancel and Close X action buttons do not return user in edit mode.

## Steps to reproduce
1. Enter edit mode
2. click edit button to edit HTML Text Editor content
3. Click Cancel, Save or the Close X at top of modal
4. notice user is no longer in edit mode.
## Expected Behavior
User remains in edit mode when returning from the HTML Editor.

zyhfish:


sbwalker:
Interactive render mode supports the ability to stay in Edit Mode by utilizing PageState - which is scoped to the user session. Static render mode is stateless and therefore needs an alternative for maintaining this information. 

In reviewing this PR, it appears to be using both cookies and UserSettings. In general I am not in favor of using cookies - especially the way this is implemented where it forces a new cookie for both authenticated and non-authenticated users. EditMode is only applicable to authenticated users so the state should only be managed for that scenario. 

It seems like this could be simplified...

- Based on a single UserSetting item named "EditModePageId"
- When a user clicks the button to toggle Edit Mode it will either set the UserSetting value to PageState.Page.PageId or ""
- The SiteRouter can reference PageState.User.Settings("EditModePageId") and if it matches PageState.Page.PageId then set PageState.EditMode accordingly, and if it does not match then set the UserSetting value to ""

zyhfish:
Hi @sbwalker , thx so much for the review, code is updated.

sbwalker:
@zyhfish this PR is still calling SettingService.GetUserSettingsAsync(user.UserId) in the SiteRouter which will affect performance and should be using PageState.User.Settings() since it already exists. I can merge this PR and make some final adjustments.

zyhfish:
Hi @sbwalker , thx so much for the refactoring.

PR #4029, opened 2024-03-20
Fix #4028: check whether the user info is null.

### LINKED ISSUE #4028: Bug: page can't be accessed when set edit permission to all users
Reproduce Steps:
1. Create a new page with edit permission to all users.
2. browse to the page with anonymous user.

Expected Result:
the page should open correctly.

Actually Result:
Exception throw out:
image

zyhfish:


sbwalker:
@zyhfish this code change is not valid:

```
 if (PageState.Runtime == Shared.Runtime.Hybrid && PageState.User != null)
```

It changes the meaning of the condition - instead of having logic dedicated to the Hybrid scenario, it will now execute the non-Hybrid logic if the User is null.  If we want to include a null check then it needs to be within the Hybrid logic:

```
 if (PageState.Runtime == Shared.Runtime.Hybrid)
 {
    if (PageState.User != null)
    {
```

zyhfish:
Hi @sbwalker , thx so much for the review, code is updated.

PR #4064, opened 2024-03-25
Fixes #4062: App.razor - moves loading body resources above blazor.web.js

### LINKED ISSUE #4062: Bug: 5.1.0 Body Scripts are loaded after opening <body> instead of before the closing </body> at the end (bottom).
# Issue

## Description
Body scripts are loaded at beginning (top of page source) of the body instead of at the end of the body (bottom of page source).

## Steps to reproduce
1. 

## Behavior
Body scripts are loaded at beginning (top) of the file.

## Expected Behavior 
Body scripts are loaded prior to the closing `</body>` element of the page.

## Additional Context
For visual performance issues when relating to user experience it is generally recommended to load these last.  Was this moved to the top?

If I am not mistaken I though we had set this to go in the `body`, `bottom` location.  Maybe an issue with the way resources are being loaded?  There should be two load points?   Top or Bottom?  Otherwise it should be bottom default.

## Screenshots

image

thabaum:
# Pull Request

Fixes #4062

## Description

Moves resources loading in the body of `App.razor` file above loading the `blazor.web.js`

## Screenshots

image
image

### DIFF
--- a/Oqtane.Server/Components/App.razor
+++ b/Oqtane.Server/Components/App.razor
@@ -68,11 +68,6 @@
                 <Routes PageState="@_pageState" RenderMode="@_renderMode" Runtime="@_runtime" AntiForgeryToken="@_antiForgeryToken" AuthorizationToken="@_authorizationToken" Platform="@_platform" @rendermode="InteractiveRenderMode.GetInteractiveRenderMode(_runtime, _prerender)" />
             }
 
-            <script src="js/app.js"></script>
-            <script src="js/loadjs.min.js"></script>
-            <script src="js/interop.js"></script>
-            <script src="_framework/blazor.web.js"></script>
-
             @if (!string.IsNullOrEmpty(_reconnectScript))
             {
                 @((MarkupString)_reconnectScript)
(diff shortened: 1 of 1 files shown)


thabaum:
@sbwalker I am not sure I just wanted to share this result, but I am not 100% it does not affect something I am not thinking of in Oqtane.  I appreciate your time reviewing.

sbwalker:
@thabaum basically you moved all statically defined JS references to the end of the body section - correct?

thabaum:
@sbwalker yes, and this improved logging out interactive/auto mode blink of an eye...

thabaum:
@sbwalker I pretty much main thing is put the `blazor.web.js` at the end as this pulls off all the magic tricks.

thabaum:
@sbwalker after further review, not sure how I ended up wtih those results shown, but this PR does not do it.  And when I had those results it worked great in interactive auto mode.  Anyhow this can be reverted back unless it does not seem to cause any issues.. my apologies.  Somehow I had it show me this source, and also the interactive/auto ran fast logging out, now I cannot reproduce this behavior or make the code get displayed as shown.  Strange... you can see in #4062 screenshot it is the same editor.

I am not sure what to think here as I am pretty sure I generated those results above somehow while in Interactive/Auto.  While I could produce those results seems like it would log out rapidly.

Maybe I am just experiencing some glitches in VS?   Or is there something finding scripts in the page body and re-writing them in oqtane?

I am updating visual studio now, it may have bugged me up as I didn't notice it was available.

sbwalker:
@thabaum I don't think it has an impact on the actual framework at run-time, but I do like the idea of the Blazor JS reference being at the very bottom of the body so I will not revert this PR.

PR #4094, opened 2024-04-01
Uniform configuration builder for DbContext and Startup.cs

pepsinio:
# Pull Request
Fixes an issue where DbContext is not taking into account .release config files or environment variables. This causes every module having DbContext to fail. 

# Description
Currently application starts fine with the respective environment variables and creation of TenantDBContext, however throws exception in case any module creates it's own dbcontext as it is re-creating connection string from the base DbContext class.

### DIFF
--- a/Oqtane.Server/Repository/Context/DBContextBase.cs
+++ b/Oqtane.Server/Repository/Context/DBContextBase.cs
@@ -2,6 +2,7 @@ using System;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Identity.EntityFrameworkCore;
@@ -19,7 +20,7 @@ using Oqtane.Shared;
 
 namespace Oqtane.Repository
 {
-    public class DBContextBase :  IdentityUserContext<IdentityUser>
+    public class DBContextBase : IdentityUserContext<IdentityUser>
     {
         private readonly ITenantManager _tenantManager;
         private readonly IHttpContextAccessor _accessor;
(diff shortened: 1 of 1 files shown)


sbwalker:
@pepsinio please notice the comment on the method you modified:

```
[Obsolete("This constructor is obsolete. Use DBContextBase(IDBContextDependencies DBContextDependencies) instead.", false)]
```

A new constructor was introduced in Feb 2023 which uses DBContextDependencies because adding any additional parameters to the older constructor would have resulted in a breaking API change. The old constructor was only retained for backward compatibility however it contains a note that it is using an anti-pattern and therefore should be avoided. Modules should use the new constructor to avoid this issue.

pepsinio:
@sbwalker fair point. I was considering to not touch it as it requires rather larger refactoring of the whole class to use only IDBContextDependencies, however if that is your preference, is fine with me :)
I tested it and seems to work fine.

sbwalker:
@pepsinio if you read the Blog article linked above it will be more clear why the old constructor is marked as obsolete and cannot be changed.

sbwalker:
@pepsinio if you use the new constructor in your module I believe everything should work fine as Startup already handles ConfigurationBuilder properly:

```
            var builder = new ConfigurationBuilder()
                .SetBasePath(env.ContentRootPath)
                .AddJsonFile("appsettings.json", false, true)
                .AddJsonFile($"appsettings.{env.EnvironmentName}.json", true, true)
                .AddEnvironmentVariables();
```

pepsinio:
Thanks @sbwalker and apologize for my ignorance here. Will close this PR.

sbwalker:
No worries... this thread is actually a reminder that beyond including Obsolete compiler warnings, Oqtane should have a policy for actually removing these methods from the framework after a specific amount of time ie. after 12 months, or once the major version is 2 increments ahead (obsolete methods from <= 3.x removed when 5.x is released).

pepsinio:
Such policy would make sense as otherwise allows for too many variations and (as in my case) can lead to confusion. Actually i am trying to use the Oqtane.Survey in this context (pun intented) here. Migrated all broken pieces to latest 5.1. Might send a PR for it as well soon.

sbwalker:
@pepsinio the best way to get a fully functional module which demonstrates the latest patterns, etc... is to use the Create Module option in the Module Management UI in Oqtane. It contains a default module template which is maintained with every release. Other open source repos which are part of the Oqtane organization on GitHub are not updated as frequently (as you can tell from the Survey module).

PR #4099, opened 2024-04-03
Fix #3625: add the clear logs function.

### LINKED ISSUE #3625: [ENHANCE] - Clear Button in Event Log
While developing it would be useful to have a clear button in the event log, i notice there isn't one at the moment 

image

zyhfish:
Fix #3625: add the clear logs function.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Logs/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Logs/Index.razor
@@ -86,47 +86,48 @@ else
 				</div>
 			</div>
 			<br />
-			<button type="button" class="btn btn-success" @onclick="SaveSiteSettings">@SharedLocalizer["Save"]</button>
+            <button type="button" class="btn btn-success" @onclick="SaveSiteSettings">@SharedLocalizer["Save"]</button>
+            <ActionDialog Header="Clear Logs" Message="@Localizer["Confirm.ClearLogs"]" Action="ClearLogs" Class="btn btn-secondary" OnClick="@(async () => await ClearLogs())" ResourceKey="ClearLogs" />
 		</TabPanel>
 	</TabStrip>
 }
 
 @code {
-	private string _level = "-";
-	private string _function = "-";
-	private string _rows = "10";
-	private int _page = 1;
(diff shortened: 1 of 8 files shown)


sbwalker:
@zyhfish this looks good... just a few comments/questions:

1. In Oqtane we try to avoid optional/default parameters on methods because they can cause future problems if the API needs to be changed. So we should avoid the following approach:

```
public int DeleteLogs(int siteId, int age, bool includeErrors = false)
```

Instead it is better to use overloads where the parameters are always explicitly defined. I have further comments related to the  includeErrors parameter below.

2. Oqtane already has the ability to specify how many days of logs you want to retain.. and a PurgeJob which does the housekeeping to ensure the log remains trimmed based on the retention setting. The addition of the includeErrors parameters seems to change the existing functionality, as it now retains error log entries by default. This means that error log items are never trimmed automatically which means that a database will become bloated with log entries over time - which is exactly what the system is trying to avoid. My preference would be to not include this new includeErrors parameter and preserve the existing functionality (ie. it should trim all items older than the retention period). The new Clear Log feature should do exactly what a user expects - fully clears the log by removing ALL items.

zyhfish:
Hi @sbwalker , thanks so much for the update, I will update the changes later.

actually in the original method, it will never delete error log entries:
https://github.com/oqtane/oqtane.framework/blob/578b7b0512922166d1dce1640b9424403d8a5e51/Oqtane.Server/Repository/LogRepository.cs#L62
`var logs = db.Log.Where(item => item.SiteId == siteId && item.Level != "Error" && item.LogDate < purgedate)`

so I think we need to review the purgejob to make sure it really purge everything including error log entries.

sbwalker:
@zyhfish I think you just identified another problem... the purge job really should delete error items as the goal of the job is to ensure that databases do not grow continuously over time.

hishamco:
Do we need a button to clear a log? this might be harmful because it will delete excellent information. While Oqtane already can specify how many days of logs you want to retain, so it should do the same thing

vnetonline:
This button should only run when developing so that the developer can delete the logs

PR #4107, opened 2024-04-05
DRAFT: Insert Scripts regardless of render modes and at the top of the script block

markdav-is:
To support 3rd part custom control libraries the JS scripts need to be included in the page for all  render modes.  To ensure that the JS is available when the app starts, ensure they are added to the top of the scripts block before the built-in scripts.  Source of this was a custom MudBlazor Theme that stopped working when upgrading to 5.1.0


markdav-is:
I put this is DRAFT mode because I noticed another place where the scripts might be getting set.

markdav-is:
I put this is DRAFT mode because I noticed another place where the scripts might be getting set.

markdav-is:
I found a work-around by using the BodyContent setting of the site to register a global script in the right place.  Closing this PR.  I'll might make an issue.

PR #4142, opened 2024-04-13
Upgraded to Bootstrap 5.3.3

leigh-pointer:
Update Oqtane Theme to 5.3.3
Reworked Blazor theme to load Resources from Info class. Bootstrap 5.3.3 caused issue with Close button [x] appearing next to the Title and not in the top right corner. Updated Theme Template

### DIFF
--- a/Oqtane.Client/Themes/BlazorTheme/ThemeInfo.cs
+++ b/Oqtane.Client/Themes/BlazorTheme/ThemeInfo.cs
@@ -1,5 +1,7 @@
+using System.Collections.Generic;
 using Oqtane.Documentation;
 using Oqtane.Models;
+using Oqtane.Shared;
 
 namespace Oqtane.Themes.BlazorTheme
 {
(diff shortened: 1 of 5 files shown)


sbwalker:
@leigh-pointer some feedback...

1. Based on the fact that everytime we update Bootstrap it introduces new cosmetic impacts which are really identified through full regression testing, I am hesitant to introduce this change. You have already identified one "breaking" change which you resolved but it is likely there are more that could impact custom themes and modules. I am actually wondering if it would be better to reserve Bootstrap changes for major releases ie. 5.2.0
2. The reason why the Blazor Theme is using the older Resource registration within the component is so that it is possible to easily test that both approaches are still working properly in the framework. So I would actually prefer to not make this change.

sbwalker:
@leigh-pointer what is your experience with Bootstrap 5.3.3?

leigh-pointer:
@sbwalker I have had no issues with 5.3.3.  I updated my branch with the latest and all is ok in both interactive and static.  I would say it is safe to merge.

PR #4164, opened 2024-04-21
Datetime formatting issue

ohba-ikuo:
### Oqtane Info

Version - 5.1.1
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the bug

When transitioning to visitor management in the Japanese location, an application error occurs.
datetime-format-issue-screen
[Event Log]
datetime-format-issue

### Expected Behavior
It is believed that an error will always occur when transitioning to Visitor Management in locations that use multibyte characters.

### DIFF
--- a/Oqtane.Client/Services/VisitorService.cs
+++ b/Oqtane.Client/Services/VisitorService.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using Oqtane.Documentation;
 using Oqtane.Shared;
 using System;
+using System.Globalization;
 
 namespace Oqtane.Services
 {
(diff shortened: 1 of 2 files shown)


ohba-ikuo:
@dotnet-policy-service agree

leigh-pointer:
Hi @ohba-ikuo are you using Oqtane with the default language or have you created a Japanese language pack?

sbwalker:
@ohba-ikuo thank you for reporting. I can see that multi-byte characters can cause some Internationalization challenges when using MMM format. The reason why the dd-MMM-yyyy format was chosen was so that there would be no ambiguity between months and days. For example, is 2023-03-01 equivalent to March 1 or January 3? I assume that the date needs to be encoded and decoded using the same Culture for this to work properly - which is not being explicitly specified in Oqtane currently as it is using DateTime.Parse(fromdate).

ohba-ikuo:
@sbwalker 
Thank you for your response. I now understand why you are using the "dd-MMM-yyyy" format for the date.

ohba-ikuo:
@dotnet-policy-service agree

sbwalker:
Thank you for making these changes... this is the recommended pattern for handling i18n (internationalization) for dates and it works well in Urls because there are no extended characters which need to be encoded.

Issue #4178, opened 2024-04-23
[ENH] - Add Prerender IModuleControl property (similar to RenderMode) to enable and disable prerender at component level 


vnetonline:
@sbwalker, I am having a rethink about this. I think this property would be beneficial for specific components. 

I have an Action (Add Course) that needs to be interactive to add courses. This ACtion is front facing the `OnInitializedAsync` and is triggered twice, causing flickering. If I was to disable prerender for this Action then it wouldn't flicker.

Can we add this property?

_Originally posted by @vnetonline in https://github.com/oqtane/oqtane.framework/discussions/4128#discussioncomment-9195637_

sbwalker:
implemented in #4179

PR #4179, opened 2024-04-23
[ENH] - Add Prerender IModuleControl property (similar to RenderMode)…

vnetonline:
… #4178

@sbwalker I had a go at adding Prerender just like RenderMode, I hope my implementation meets your standards let me know if you would like me to modify anything

### DIFF
--- a/Oqtane.Client/Modules/ModuleBase.cs
+++ b/Oqtane.Client/Modules/ModuleBase.cs
@@ -52,6 +52,8 @@ namespace Oqtane.Modules
 
         public virtual string RenderMode { get { return RenderModes.Interactive; } } // interactive by default
 
+        public virtual bool Prerender { get { return true; } } // prerender at server by default
+
         // url parameters
         public virtual string UrlParametersTemplate { get; set; }
 
--- a/Oqtane.Client/Modules/ModuleBase.cs
+++ b/Oqtane.Client/Modules/ModuleBase.cs
@@ -52,7 +52,7 @@ namespace Oqtane.Modules
 
         public virtual string RenderMode { get { return RenderModes.Interactive; } } // interactive by default
 
-        public virtual bool Prerender { get { return true; } } // prerender at server by default
+        public virtual bool? Prerender { get { return null; } } // prerender at server by default
 
         // url parameters
         public virtual string UrlParametersTemplate { get; set; }
--- a/Oqtane.Client/UI/ModuleInstance.razor
+++ b/Oqtane.Client/UI/ModuleInstance.razor
@@ -7,7 +7,7 @@
 }
 else
 {
-    if (ModuleState.Prerender != false)
+    if (ModuleState.Prerender == false)
     {
         <RenderModeBoundary ModuleState="@ModuleState" PageState="@PageState" SiteState="@SiteState" @rendermode="@InteractiveRenderMode.GetInteractiveRenderMode(PageState.Site.Runtime, false)" />
 
--- a/Oqtane.Client/UI/ModuleInstance.razor
+++ b/Oqtane.Client/UI/ModuleInstance.razor
@@ -7,7 +7,7 @@
 }
 else
 {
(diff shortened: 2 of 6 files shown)


sbwalker:
@vnetonline this looks good... the only question I have is related to the default behavior:

Currently there is a Site setting for Prerender which controls whether or not pre-rendering is enabled for the entire site. Currently it is possible to set Prerendering to False at the Site level and only enable Prerendering for some specific components (ie. only some Interactive components on public facing pages which need to be indexed for SEO purposes). Based on this PR, the scenario I described would not be possible due to the default value and logic below:

ModuleState.Prerender != false ? PageState.Site.Prerender : ModuleState.Prerender

Basically this says that if the component-level property is set to True then it should use the Site Setting - Prerender property setting (which would be False in the case I described above). Perhaps a more appropriate solution would be if the component-level Prerender property was a nullable boolean type and had a default value of null. Then the logic  would be able to determine if the value was overridden or not:

ModuleState.Prerender == null ? PageState.Site.Prerender : ModuleState.Prerender

The above logic uses the Site Setting unless it is explicitly overridden at the component level.

Thoughts?

vnetonline:
@sbwalker, the above logic sounds good, I would reverse it from a conventional perspective, it's more readable like this 

`ModuleState.Prerender != null ?  ModuleState.Prerender : PageState.Site.Prerender`

I will modify the PR and submit it againt

vnetonline:
@sbwalker i can do what you suggested using below code block ... but it doesn't look elegant enough 

```
if (ModuleState.Prerender != false)
{
    <RenderModeBoundary ModuleState="@ModuleState" PageState="@PageState" SiteState="@SiteState" 
        @rendermode="@InteractiveRenderMode.GetInteractiveRenderMode(PageState.Site.Runtime, false)" />

} 
else if (ModuleState.Prerender == true)
{
    <RenderModeBoundary ModuleState="@ModuleState" PageState="@PageState" SiteState="@SiteState" 
        @rendermode="@InteractiveRenderMode.GetInteractiveRenderMode(PageState.Site.Runtime, true)" />
}
else
{
    <RenderModeBoundary ModuleState="@ModuleState" PageState="@PageState" SiteState="@SiteState" 
        @rendermode="@InteractiveRenderMode.GetInteractiveRenderMode(PageState.Site.Runtime, 
        PageState.Site.Prerender)"/>
}
```

vnetonline:
> @sbwalker i can do what you suggested using below code block ... but it doesn't look elegant enough
> 
> ```
> if (ModuleState.Prerender == false)
> {
>     <RenderModeBoundary ModuleState="@ModuleState" PageState="@PageState" SiteState="@SiteState" 
>         @rendermode="@InteractiveRenderMode.GetInteractiveRenderMode(PageState.Site.Runtime, false)" />
> 
> } 
> else if (ModuleState.Prerender == true)
> {
>     <RenderModeBoundary ModuleState="@ModuleState" PageState="@PageState" SiteState="@SiteState" 
>         @rendermode="@InteractiveRenderMode.GetInteractiveRenderMode(PageState.Site.Runtime, true)" />
> }
> else
> {
>     <RenderModeBoundary ModuleState="@ModuleState" PageState="@PageState" SiteState="@SiteState" 
>         @rendermode="@InteractiveRenderMode.GetInteractiveRenderMode(PageState.Site.Runtime, 
>         PageState.Site.Prerender)"/>
> }
> ```

Simplefied Version 

```
    var rendermode = InteractiveRenderMode.GetInteractiveRenderMode(PageState.Site.Runtime, ModuleState.Prerender ? 
       PageState.Site.Prerender);

    <RenderModeBoundary ModuleState="@ModuleState" PageState="@PageState" SiteState="@SiteState" 
        @rendermode="rendermode" />
```

sbwalker:
@vnetonline the PR is already merged

vnetonline:
> @vnetonline the PR is already merged

Yes thank you was just adding a comment as to what changed in this commit

PR #4198, opened 2024-04-27
Fix #4030: add copy module option for add existing module function.

### LINKED ISSUE #4030: Enh: Add Support For Copy Existing Module to Control Panel
Reproduce Steps
1. Create a new page, use "Add existing module" function to copy a module from home page.
2. edit the module with some new content.
3. browse to the home page.

Expected Result:
home page should keep it's original content.

Actually Result:
After make changes in the new page, home page's content will also be affected. at least we should give an option for whether it's a copy or reference action to avoid any confuse.

zyhfish:
Fix #4030: add copy module option for add existing module function.

### DIFF
--- a/Oqtane.Client/Resources/Themes/Controls/ControlPanelInteractive.resx
+++ b/Oqtane.Client/Resources/Themes/Controls/ControlPanelInteractive.resx
@@ -154,7 +154,7 @@
     <value>Not Authorized</value>
   </data>
   <data name="Message.Require.ModuleSelect" xml:space="preserve">
-    <value>You Must Select A Module And The Copy Mode.</value>
+    <value>You Must Select A Module.</value>
   </data>
   <data name="Module.Manage" xml:space="preserve">
     <value>Module Management:</value>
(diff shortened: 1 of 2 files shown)


sbwalker:
@zyhfish my preference for implementing this feature is that we keep the UI very simply and consistent with the approach used currently. I would simply add a "Copy Existing Module" option to the first option in Module Management:

image

This new option will behave similar to "Add Existing Module" ie. it will allow you to select a Page/Module:

image

However, copy is only possible if a module implements IPortable - so the list of modules should either be filtered so that modules which do not support IPortable are not displayed (ie. using module.ModuleDefinition.IsPortable) or the system allows a user to select the module and then provides feedback in the existing message area that the module does not support the ability to be copied. The first option prevents users from selecting certain modules but may result in users logging issues because they think the behavior is a bug. The second option displays all modules and provides explicit feedback to the user which offers better transparency and clarity - so is probably the better approach.

I realize that for some folks with a DNN background, the technical concepts of "sync", etc... makes sense, however Oqtane is focused on a broader audience so I would prefer to not follow the DNN patterns.

zyhfish:
Hi @sbwalker , thanks so much for the update, the code has been updated, could you please help to check again? I also noticed that some times control panel was hidden but the mask still display, this may need open another ticket to analyze.

sbwalker:
@zyhfish I will merge this PR and make some minor adjustments

PR #4217, opened 2024-05-01
fix support for Site-level Scripts in Resources

sbwalker:


sbwalker:
@markdav-is when you get a chance, please verify if this PR resolves your issue with MudBlazor

PR #4225, opened 2024-05-03
Fix #4224: close the control panel after adding a module.

### LINKED ISSUE #4224: [BUG] the control panel mask element won't hide after add a new module into page
### Oqtane Info

Version - 5.1.1
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the bug
the control panel mask element won't hide after add a new module into page, user has to click on the page to hide the mask manually.

### Expected Behavior
The dialog mask should be hidden automatically after the control panel closed.

### Steps To Reproduce
1. login as host, click the gear button to open control panel.
2. the control panel will be closed after adding a new module into the page, but the mask still there.
image

zyhfish:
Fix #4224: close the control panel after adding a module.

### DIFF
--- a/Oqtane.Client/Themes/Controls/Theme/ControlPanelInteractive.razor
+++ b/Oqtane.Client/Themes/Controls/Theme/ControlPanelInteractive.razor
@@ -432,6 +432,7 @@
 
                 _message = $"<div class=\"alert alert-success mt-2 text-center\" role=\"alert\">{Localizer["Success.Page.ModuleAdd"]}</div>";
                 _title = "";
+                await new Interop(jsRuntime).HideControlPanel();
                 NavigationManager.NavigateTo(Utilities.NavigateUrl(PageState.Alias.Path, PageState.Page.Path, ""));
             }
             else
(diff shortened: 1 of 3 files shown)


sbwalker:
@zyhfish we want to keep the current behavior so that it is possible to add multiple modules without having to re-open the control panel

zyhfish:
Hi @sbwalker , current behavior is the control panel will be closed but just left the dialog mask element display on the page.

I will try to find a solution to keep the control panel open after add the module.

leigh-pointer:
@zyhfish i have looked at the functionality in Edge, Chrome, FireFox, Brave and don't see any mask issue.
What browser are you using ?

zyhfish:
Hi @leigh-pointer , I'm using Chrome, and weird that it's CNR from my end now, it was a persist issue before, could that related to the CDN of bootstrap and the provider update the resource?

leigh-pointer:
@zyhfish I heard something on the interweb about some issues with a few CDN's not performing well. This  must have been the issue.

zyhfish:
Hi @leigh-pointer , I confirmed this is reproducible only with Static render mode. in Interactive mode it works fine, the control panel will keep open after adding a module.

PR #4226, opened 2024-05-03
Fix #4223 In the Ubuntu environment, an error occurs when trying to upload a file.

### LINKED ISSUE #4223: [BUG] In the Ubuntu environment, an error occurs when trying to upload a file.
### Oqtane Info

Version - 5.1.1
Render Mode - Static 
Interactivity - Server
Database - SQL Server
OS - Ubuntu 20.04.6 LTS
WebServer - Nginx

### Describe the bug

On Ubuntu 
In the "File Management" feature, specifying "--Public" as the folder and uploading an image file causes an application error.

upload-file-issue

The following is output in the error log.

upload-file-issue_01

The system information is as follows.

upload-file-issue_02

### Expected Behavior

The actual directory created is as follows.

`/var/oqtane/var/oqtane/wwwroot/Content/Tenants/1/Sites/1/Public`

correct

`/var/oqtane/wwwroot/Content/Tenants/1/Sites/1/Public`

This error is likely caused by the different directory formats between Unix and Windows.

### Steps To Reproduce

In an Ubuntu environment, upload a file to a folder where no files have been uploaded before.

ohba-ikuo:


sbwalker:
@ohba-ikuo this is not the appropriate fix as it removes functionality. The current logic calls CreateDirectory() for all directories which do not exist that are part of the path - which is required functionality.

ohba-ikuo:
@sbwalker 
I reverted the source back and fixed it so that the absolute path argument does not become a relative path in the Unix environment.

Issue #4246, opened 2024-05-09
[BUG] ModuleMessage Throws Exception in Static Render Mode For Static Components


sbwalker:
### Oqtane Info

Version - 5.1.1
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the bug

When a module component is using Static render mode, the AddModuleMessage() can be used to display a message to a user. However when a user clicks the X icon to close the message, an error will be produced in the browser similar to the following:

```
Cannot submit the form 'ModuleMessageForm33' because no form on the page currently has that name.
```

If you look at the HTML markup of the page prior to clicking the X icon, it appears that the form does exist:

![image](https://github.com/oqtane/oqtane.framework/assets/4840590/fb5fd6f7-a479-4a9a-9aab-0823fb25431b)

### Expected Behavior

The message should be closed and no error should occur.

### Anything else?

I believe Blazor may be removing the form element from the DOM before the onsubmit event is fired? Or perhaps the elementreferences in the ModuleMessage component are causing issues. 

A secondary issue is that the RenderModeBoundary is creating 2 ModuleMessage components regardless of whether a message needs to be displayed or not. This adds a lot of extra components to a page which is contrary to the performance goals of Oqtane. It would be preferable if ModuleMessage components are only added when a message needs to be displayed.

PR #4268, opened 2024-05-18
Fix #3885: only re-render the component when message changed.

zyhfish:
Fix #3885: only re-render the component when message changed.

### DIFF
--- a/Oqtane.Client/Modules/Controls/ModuleMessage.razor
+++ b/Oqtane.Client/Modules/Controls/ModuleMessage.razor
@@ -12,12 +12,13 @@
             {
                 <NavLink class="ms-2" href="@NavigateUrl("admin/log")">View Details</NavLink>
             }
-            <a href="@NavigationManager.Uri" class="btn-close" data-dismiss="alert" aria-label="close"></a>
+            <button type="button" class="btn-close" data-dismiss="alert" aria-label="close" @onclick="CloseMessage"></button>
         }
     </div>
 }
 
 @code {
+    private string _message = string.Empty;
     private string _classname = string.Empty;
     private string _formname = "ModuleMessageForm";
 
@@ -27,15 +28,8 @@
     [Parameter]
     public MessageType Type { get; set; }
 
-    public void RefreshMessage(string message, MessageType type)
-    {
-        Message = message;
-        Type = type;
-
-        UpdateClassName();
-
-        StateHasChanged();
-    }
+    [Parameter]
+    public RenderModeBoundary Parent { get; set; }
 
     protected override void OnInitialized()
     {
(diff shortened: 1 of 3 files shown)


sbwalker:
@zyhfish Static rendering does not support buttons with onclick handlers - this is why the close option was changed to a hyperlnk.

zyhfish:
Hi @sbwalker, the hyper link doesn't work well in static mode, it cause different behaviors, that's why I changed it to button, and I verified both with Static and Interactive mode, it works fine in the module management page with below steps:
1. create a new module with "Oqtane" in the owner name field;
2. click save button will display the module message;
3. click X button to close the message.

sbwalker:
@zyhfish I know these render modes are a bit confusing. In order to test this scenario in both render modes you need to ensure your Site Settings are set to use Static render mode (the default). However, when you create a new module from the module template, by default all of its module components are configured to be Interactive. So when you run the module, the site will be Static but the module components will be rendered Interactively. And since interactivity is inherited it means that the ModuleMessage component inherits the Interactive rendering from the module component. This is why the button onclick handler in the ModuleMessage is working for you (which is a clear indicator that that module is running interactively as button onclick does not fire at all in static rendering).

In order to change the behavior of the module to use Static rendering, you need to modify the Index.razor component and uncomment the line:

```
    public override string RenderMode => RenderModes.Static;
```

This will result in the module component using Static rendering within the Static site. And you will notice in this configuration the button onclick will no longer function properly. The way this was resolved for the 5.1 release was the ModuleMessage was modified to use a form element and button onsubmit handler - which is supposed to work in both Interactive and Static rendering. And it does work in most cases... however there is a case where it does not work. You can see the problem in action by visiting this page https://www.oqtane.org/blog scrolling to the bottom, enter your email address, and click Subscribe. The blog module will display a ModuleMessage component however if you click the Close X icon, an error message will be displayed in the browser "Error: 400 Bad Request". In a development environment the error message is a bit more detailed and will say "Cannot submit the form 'ModuleMessageForm##' because no form on the page currently has that name" - which was described in the Issue #4246. I personally think this is another bug in Blazor however it is not acceptable behavior so there needs to be a solution in Oqtane.

The solution I found for static render mode was to use a hyperlink. However it does not seem to work well in interactive render mode. So perhaps this is another case where we need to include conditional logic in ModuleMessage to handle both render modes. For static it can use a hyperlink and interactive it can use button onclick. I will merge your PR and make the necessary changes.

In the future if Microsoft fixes the form onsubmit behavior we may be able to revert back to a single solution which can support both render modes (I still need to log this issue in the Blazor repo with a simplistic repro example).

sbwalker:
@zyhfish #4269 includes conditional logic to support static render mode. I tested the following scenarios:

Static module component
1. AddModuleMessage called in OnParametersSet
2. AddModuleMessage called in onsubmit event handler raised by form element button

Interactive module component
1. AddModuleMessage called in OnParametersSet
2. AddModuleMessage called in onclick event handler raised by standard button

These all appear to work now. And the RenderModeBoundary is only rendering a ModuleMessage component if a message needs to be displayed - which resolves the original issue identified related to performance/scalability. Thank you!!

Issue #4284, opened 2024-05-24
[BUG] - When a user has a role that has a expiry date that is not Valid (i.e. in the past) it is still authorised for that role


vnetonline:
### Oqtane Info

Version - 5.1.2 (dev Branch)
Render Mode - Static
Interactivity - Server
Database - SQL Server

### Describe the bug
- When a user has a role that has a expiry date that is not Valid (i.e. in the past) it is still authorised for that role 

### Expected Behavior
- The user should not be authorised if the user has a role which has expired

### Steps To Reproduce
1. Add a role and set the expiry date to yesterday
2. Add that role to a user 
3. Add HTML module and set its permission to view and edit to that specific role 
4. Log in as the user assigned that role, and you will still be able to view and edit the HTML Module

leigh-pointer:
@vnetonline Good point! However, roles should expire and be removed automatically after their expiry date. We might need a system to identify and handle temporary roles for automatic removal upon expiration.

vnetonline:
Will sync events be good to remove the roles on expiry??

leigh-pointer:
> Will sync events be good to remove the roles on expiry??

Think that is over use ov the framework, when it could be implemented with a simple remove on expiry flag. 
I see this being applied before the role is used thus removing the assignment and returning false for authorization.

vnetonline:
Close this as you can use the UserRoleService to validate the effective and expiry date in your own module

vnetonline:
@sbwalker reopening this issue as I think this should be handled by the framework just like it does for Page

leigh-pointer:
I will look in to this and raise a PR

PR #4287, opened 2024-05-25
[FIX] - #4284 

vnetonline:
When a user has a role that has an expiry date that is not Valid (i.e. in the past) it is still authorised for that role


vnetonline:
Closing this as you can use the UserRoleService to validate the effective and expiry date in your own module

PR #4293, opened 2024-05-28
performance optimization

zyhfish:


sbwalker:
#4299 addresses this scenario

PR #4319, opened 2024-06-06
Fix #4318: remove the page info when perform a new search.

### LINKED ISSUE #4318: [BUG] page index will not be reset when perform a new search
### Oqtane Info

Version - Latest
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the bug
the page index will not reset when perform a new search after there is search results display and switched to other pages.

### Expected Behavior
the page index should be reset to the first page when perform a new search.

### Steps To Reproduce
1. add more than 10 pages with a same keyword in the name.
2. wait for the search index job to run.
3. search for the keyword, you will see more than one page in the pager list, click some page numbers to browse to the pages other than the first page.
4. change the search keyword to perform a new search.

zyhfish:
Fix #4318: remove the page info when perform a new search.

### DIFF
--- a/Oqtane.Client/Modules/Admin/SearchResults/Index.razor
+++ b/Oqtane.Client/Modules/Admin/SearchResults/Index.razor
@@ -106,6 +106,10 @@
             ClearModuleMessage();
 
             _currentPage = 0;
+            if(PageState.QueryString.ContainsKey("page"))
+            {
+                PageState.QueryString.Remove("page");
+            }
             await PerformSearch();
         }
     }


sbwalker:
@zyhfish I am not sure I understand this PR. The PageState object is intended to be used in a read-only manner. Basically it is populated in the SiteRouter and its values should not be modified - they should only be consumed by downstream components.

zyhfish:
Hi @sbwalker , This is because in the Pager component the query string value has higher priority than the current page value, so we have to remove the query string from the collection:
https://github.com/oqtane/oqtane.framework/blob/dev/Oqtane.Client/Modules/Controls/Pager.razor#L455

sbwalker:
closing - replaced by #4336

PR #4328, opened 2024-06-10
Update Project Templates from 8.0.5 - 8.0.6

leigh-pointer:


sbwalker:
@leigh-pointer search is still a work in progress. I just renamed the ISearchable method which will break this PR.

leigh-pointer:
> @leigh-pointer search is still a work in progress. I just renamed the ISearchable method which will break this PR.

PR updated with new implementation.

PR #4334, opened 2024-06-11
Upgrade to Bootstrap 5.3.3

leigh-pointer:
Oqtane ThemeInfo updated
Blazor Default.razor updated


sbwalker:
@leigh-pointer I thought there was a cosmetic breaking change in the previous Bootstrap 5.3.3 PR which you included a fix for?

leigh-pointer:
There was, and I checked it and in the BS 5.3.3 css, the issue has somehow been resolved.  It was with the modal and the [x] not justifying right.

PR #4345, opened 2024-06-18
update the search function.

zyhfish:
update the search function.


sbwalker:
I believe all of these changes have been integrated through other PRs

PR #4348, opened 2024-06-21
Issue 4284 Role was not being checked if still effective.

leigh-pointer:


sbwalker:
@leigh-pointer I have reviewed the PR and I have some comments:

1. The changes to the Control Panel do not seem to be related to user roles and do not seem to be valid. The purpose of the logic is to determine if an authenticated user should be able to view the Admin Dashboard button. Basically if they have the ability View any Admin Page they should be able to see this button. You have changed the logic to evaluate if they have Edit rights to any Admin Pages - which is not correct.

2. The changes to UserPermissions will cause performance issues. The call to GetUser() is called in MANY places and is intended to be a very fast method to populate a User object from the Principal (which contains all of the essential user information). This PR introduces a database call which will occur whenever GetUser() is called which will result in a LOT of database activity which will impact performance. One way to mitigate this is to cache the user roles but this will cause other headaches in terms of cache invalidation. My vote is to remove this logic for now - even if it means that effective/expiry for the role assignment is only enforced at login and not in real-time.

sbwalker:
@leigh-pointer I will investigate adding caching for the UserRoles to address the performance issue mentioned above

sbwalker:
resolved by #4453

PR #4479, opened 2024-08-02
Fix admin/pages not showing 404 for unauthorized users

pollux:
When logged in with an account only in the Registered Users role, the page under the URL "admin/pages" does not show a 404 page. Instead it stays blank. This behavior is different to all the other admin pages, and this PR fixes that.


sbwalker:
@pollux this behavior might actually be by design - I believe it needs to be set this way in order for Oqtane to support "delegated administration".  I will need to investigate further.

sbwalker:
@pollux I believe the original reason why the Page Management page was made visible to Registered Users in the template was so that the Control Panel was able to dynamically determine the Page Management page path using the PageState.Pages collection for users who had the right to Edit the page (but were not Administrators). This logic has since been modified to use a Site Setting instead to pass this information from the server via the SiteService. So I believe this PR is valid and I will merge it.

sbwalker:
@pollux it turns out the Registered Users role assignment is necessary to support the Personalized Pages feature.

PR #4520, opened 2024-08-14
move HtmlText caching from repository to service layer

sbwalker:


leigh-pointer:
Ok, but why? interested!

sbwalker:
@leigh-pointer the Search Indexer Job runs every minute by default and indexes every HtmlText module in a site. Due to the fact that caching was implemented in the Repository layer it meant that the content of every HtmlText module would be loaded into memory. This is good for performance but not good for overall memory utilization on the server as these items will never be expired from the cache (since they are referenced every minute). By moving caching to the service layer it means that the UI will serve HtmlText content from the cache - but if that content is not accessed for 30 minutes it will be flushed from the cache - resulting in more efficient memory utilization on the server. The Search Indexer Job will call the repository without any caching - which is fine because it is happening in a background process.

PR #4546, opened 2024-08-22
Fix #4545: clear the cache after site is updated of deleted.

### LINKED ISSUE #4545: [BUG] Site Settings / UI Component Settings sometimes shows old value after save
### Oqtane Info

Version - 5.2.1
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the bug
After changing the UI Component Serttings in Admin Dashboard / Site Settings and press the Save button the values are updated on disk - but not always on screen.
I guess it has to do with the cache.
See video.

### Anything else?
[UIComponentVideo.zip](https://github.com/user-attachments/files/16707069/UIComponentVideo.zip)

zyhfish:


sbwalker:
@zyhfish cache invalidation for site updates is already handled in CacheInvalidationEventSubscriber - can you please describe the scenario where it is not working so I can reproduce the issue.

sbwalker:
@zyhfish also the cache invalidation on a delete operation should not be necessary as it will automatically timeout in 20 minutes if it is not referenced

sbwalker:
I merged #4548 to handle the site deletion scenario

PR #4574, opened 2024-08-28
display the upgrade progress.

zyhfish:


sbwalker:
@zyhfish if you search the code base for "bootstrap/5" you will find a number of external references to Bootstrap:

- Install Wizard
- Blazor Theme
- Oqtane Theme (for JavaScript - not CSS)

These all use https://cdnjs.cloudflare.com as the source (including the Integrity attribute)

The new app_offline.bak uses https://cdn.jsdelivr.net (without Integrity attribute)

For consistency and ease in maintaining the framework, could you please abstract these references into the Constants.cs (as public static readonly strings):

BootstrapScriptUrl = "https://cdnjs.cloudflare.com/ajax/libs/bootstrap/5.3.3/js/bootstrap.bundle.min.js" 
BootstrapScriptIntegrity = "sha512-7Pi/otdlbbCR+LnW+F7PwFcSDJOuUJB3OxtEHbg4vSMvzvJjde4Po1v4BR9Gdc9aXNUNFVUY+SK51wWT8WF0Gg=="
BootstrapStylesheetUrl = "https://cdnjs.cloudflare.com/ajax/libs/bootstrap/5.3.3/css/bootstrap.min.css" 
BootstrapStylesheetIntegrity = "sha512-jnSuA4Ss2PkkikSOLtYs8BlYIeeIK1h99ty4YfvRPAlzr377vr3CXDb7sb7eEEBYjDtcYj+AjBH3FLv5uSJuXg=="

And reference the constant in all of the various locations above. This will allow us to update the Bootstrap reference in 1 place in the future.

zyhfish:
Hi @sbwalker ,
thanks so much for review the changes, one issue is that the Oqtane.Update project is a standalone project and it don't have any reference to the Oqtane.Shared project so it can't  use the Constants class,  should we add the reference?

Thanks,
Ben

sbwalker:
I don't think it's a problem to include the Shared reference as the Updater assembly is located in the /bin - however the app_offline.bak is part of the Server project... and the fact that it is an HTML file means that it cannot reference constants anyways right?

zyhfish:
submitted #4577 to handle this change request.

PR #4599, opened 2024-09-10
Avoid NRE

hishamco:
This PR avoids Null Reference Exceptions that might occur in many places

### DIFF
--- a/Oqtane.Client/Modules/Admin/ModuleDefinitions/Edit.razor
+++ b/Oqtane.Client/Modules/Admin/ModuleDefinitions/Edit.razor
@@ -408,10 +408,12 @@
 
     private async Task GetPackage(string packagename)
     {
-        var version = _packages.Where(item => item.PackageId == packagename).FirstOrDefault().Version;
+        var version = _packages.Where(item => item.PackageId == packagename)
+            .FirstOrDefault()?.Version;
         try
         {
             _package = await PackageService.GetPackageAsync(packagename, version, false);
+
             if (_package != null)
             {
                 StateHasChanged();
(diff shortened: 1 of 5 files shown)


sbwalker:
@hishamco can you please explain how you were able to trigger null reference exceptions in your testing? I have been trying and I have not been able to reproduce the problem. Basically, if it is possible to trigger a null reference exception in the areas identified it would mean that your installation has some major integrity issues ie. there are core configurations missing from your database... so even if the NREs were prevented in these areas, your application would still crash because it is missing critical dependencies.

hishamco:
Because we are missing unit tests :(

I will struggle to add unit tests to OqtaneFramework if you don't mind, there are many places where I can reproduce the bugs. It's not necessary to trigger them from within UI. IMHO having a robust unit & integration tests will increase the quality of the project

sbwalker:
@hishamco you are correct that Oqtane does not have unit tests, but it does not answer my question of how you were able to trigger the issue ie. the steps to reproduce the current problem you reported.

hishamco:
> but it does not answer my question of how you were able to trigger the issue

It's simple to simulate the scenario by creating a unit test

sbwalker:
Please provide an example of the unit test so that I can verify if the unit test actually relates to a real world scenario. My approach for maintaining Oqtane is to never merge a PR unless I am able to validate that the issue is a problem. Reproducing the issue allows me to investigate to determine if the PR addresses the issue correctly, causes side effects, or is illogical.

hishamco:
Agree but sometimes in OSS you need to make sure that your code is not buggy because the developer might use some services that are not used in UI. 

I had a great experience when I worked on ASP.NET Core source and a few OSS projects when you needed to provide unit or integration tests to validate the introduced APIs

PR #4600, opened 2024-09-10
Fix UserManager.GetUser()

hishamco:


Oqtane.Server/Managers/UserManager.cs
```diff
@@ -66,7 +66,7 @@ public User GetUser(int userid, int siteid)
                     user.SiteId = siteid;
                     user.Roles = GetUserRoles(user.UserId, user.SiteId);
                     List<Setting> settings = _settings.GetSettings(EntityNames.User, user.UserId).ToList();
-                    user.Settings = settings.Where(item => !item.IsPrivate || user.UserId == user.UserId)
+                    user.Settings = settings.Where(setting => !setting.IsPrivate && setting.EntityId == user.UserId)
```
hishamco, line 69: > user.UserId == user.UserId

This was a big mistake that might have happened accidentally. Also, I think we need to use and operator instead of or

sbwalker:
@hishamco thank you for identifying this issue - it was mistakenly introduced as part of an earlier performance enhancement. It requires a different solution than what you suggested - which I have now merged as part of #4605

PR #4601, opened 2024-09-10
Add null operator in ModuleState.ModuleDefinition

hishamco:


Oqtane.Client/Modules/Admin/Modules/Settings.razor
```diff
@@ -167,7 +167,7 @@
     {
         SetModuleTitle(Localizer["ModuleSettings.Title"]);
 
-        _module = ModuleState.ModuleDefinition.Name;
+        _module = ModuleState.ModuleDefinition?.Name;
```
hishamco, line 170: Already checked in line 187

```csharp
if (ModuleState.ModuleDefinition != null)
{
...
}
```

sbwalker:
@hishamco this is a valid issue. I am guessing you were able to trigger this issue is by deleting a module assembly from the /bin for a module which is already in use within a site - which essentially creates an "orphaned module" scenario (hence the ModuleDefinition property is null).

In regards to fixing the problem, my preference would be to move the _module assignment to within the condition if block where all of other assignments which depend on ModuleDefinition are already located. It makes the logic clearer.

```
if (ModuleState.ModuleDefinition != null)
{
>>   _module = ModuleState.ModuleDefinition.Name; <<
   _permissionNames = ModuleState.ModuleDefinition?.PermissionNames;
```

if you make this change I will merge this PR.

PR #4602, opened 2024-09-10
Fix CreateTenant()

hishamco:
Fixes bug in line 181

https://github.com/oqtane/oqtane.framework/blob/e1949717279d331bc27e20c7f9e8f631658630d2/Oqtane.Server/Infrastructure/DatabaseManager.cs#L173-L196

### DIFF
--- a/Oqtane.Server/Infrastructure/DatabaseManager.cs
+++ b/Oqtane.Server/Infrastructure/DatabaseManager.cs
@@ -342,9 +342,9 @@ namespace Oqtane.Infrastructure
 
                         _cache.Remove("aliases");
                     }
-            }
 
-            result.Success = true;
+                result.Success = true;
+            }
 
             return result;
         }


sbwalker:
@hishamco please explain the bug that this PR fixes - I will need to be able to reproduce this bug locally in order to validate that this PR is a valid.

hishamco:
If you look closely at `CreateTenant` it will return `true` all the time, for that line 181 is always success

sbwalker:
You are correct that the current logic seems strange because it initializes Success to false and then always sets it to true. However the current logic actually behaves as expected ie. the installer completes successfully for all 4 possible scenarios.

```
       private Installation CreateTenant(InstallConfig install)
       {
           var result = new Installation { Success = false, Message = string.Empty };
           if (!string.IsNullOrEmpty(install.TenantName) && !string.IsNullOrEmpty(install.Aliases))
           {
               ...
           }

           result.Success = true;

           return result;
       }
```

The change in this PR actually changes the behavior and would result in Success being set to false if either the TenantName or Aliases are null - which would break the installer for 2 out of the 4 possible scenarios.

hishamco:
What if the `tenant` is `null` in the line

```csharp
tenant = db.Tenant.FirstOrDefault(item => item.Name == install.TenantName);
```

sbwalker:
@hishamco if the tenant is null it means that you have a completely broken Oqtane installation and an exception should have been thrown way before it reached this line of code. Can you please explain how you encountered this issue? It would be more important to understand how to reproduce this problem (ie. the root cause) so that it can be prevented... as checking for null at this point would not be a solution, it would just be a bandaid which would result in an exception in subsequent operations.

sbwalker:
closing this PR as it negatively affects a number of installer scenarios

PR #4608, opened 2024-09-13
Fix issue in toggle edit mode

hishamco:


sbwalker:
@hishamco  I tested the current logic and it does not actually seem to be an "issue"? This line of code is only executed for personalized pages - and the functionality works as expected. The only "problem" appears to be that it sets PageState.EditMode when its not required - however this does not affect the behavior of the application. Please explain if I am missing something based on your own testing.

Oqtane.Client/Themes/Controls/Theme/ControlPanel.razor
```diff
@@ -147,8 +147,7 @@
         {
             if (PageState.Page.IsPersonalizable && PageState.User != null && UserSecurity.IsAuthorized(PageState.User, RoleNames.Registered))
             {
-                PageState.EditMode = true;
```
hishamco, line 150: I'm not sure how you set the `EditMode` then check it in the line after?!!

hishamco: As I said in another PR I'm trying to test the functionality from within the code to improve the quiality

sbwalker: I am not sure what you mean by "testing the functionality from within the code"? Does this mean that you are not actually testing the code at run-time, but instead using static code analyzers to identify potential issues?

hishamco: I'm not testing the code at the runtime, sometimes I identifying the bugs by observation or static code analyzers

sbwalker: Ok it sounds like the answer to my earlier question is that you did not experience any functional problems with the application ie. this is not an "issue" according to the standard definition. This is an optimization to code which is already working properly. In general my philosophy in these cases is "if it aint broke then dont fix it" as there are often unexpected side effects. In this case the change is very straightforward and easy to test/validate so I will merge the PR.

hishamco:
@sbwalker shall I start some unit tests that examine several scenarios that might break the code

PR #4677, opened 2024-09-26
sign out the principal when it is rejected due to security stamp changes

sbwalker:


Oqtane.Server/Security/PrincipalValidator.cs
```diff
@@ -7,13 +7,15 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Authentication;
 
 
 namespace Oqtane.Security
 {
     public static class PrincipalValidator
     {
-        public static Task ValidateAsync(CookieValidatePrincipalContext context)
+        public static async Task ValidateAsync(CookieValidatePrincipalContext context)
```
thabaum, line 18: I was looking for this, nice catch.

PR #4705, opened 2024-10-04
Hard deletion of page more robust use of entity framework contexts

maurocavallin:
The issue:

Step to Reproduce:
- I have a simple page with a simple Html module in it.
- First I soft-delete the page.
- Then I hard-delete the same page from Recycle Byn.

Spot the issue:
All seems to go well on UI but looking at the record in the db I see a stale record of that page still present (module and permission are correctly deleted, but not the record in the page db table).

Since the page on db has no permission linked to it I won't see that page anywhere in the UI. But it will remain in database forever.

This happens in SqlServer and SqlLite (an i guess in any other db, its an entity framework context thing).

The problem is the use of multiple opened ef context in deletion of module of the page (one ef dbcontext) and the page itself (another ef dbcontext): the fix in code is self-explanatory.

### DIFF
--- a/Oqtane.Server/Repository/PageRepository.cs
+++ b/Oqtane.Server/Repository/PageRepository.cs
@@ -91,18 +91,29 @@ namespace Oqtane.Repository
         public void DeletePage(int pageId)
         {
             using var db = _dbContextFactory.CreateDbContext();
-            var page = db.Page.Find(pageId);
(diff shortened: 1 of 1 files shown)


maurocavallin:
@dotnet-policy-service agree

sbwalker:
@maurocavallin thank you for the detailed explanation of the problem

Issue #4716, opened 2024-10-09
[ENH] Add Sort On Deleted date to the Recycle Bin


leigh-pointer:
### Oqtane Info

Version - 2.3.2#
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the enhancement
Locating a deleted page or module can often be a tedious process, particularly when the module names are set to their default values. To streamline this task, we propose implementing a sorting feature based on the Deleted Date. This enhancement will significantly improve the efficiency of tracking down deleted modules or pages, allowing users to quickly identify and recover the information they need.

PR #4717, opened 2024-10-09
Add Sort On Deleted date to the Recycle Bin #4716

leigh-pointer:


sbwalker:
@leigh-pointer I like the idea of sorting the items based on DeletedOn (with most recent first) but I would prefer to not implement the ascending/descending sort options in the UI

sbwalker:
replaced with #4731

Issue #4692, opened 2024-10-01
[ENH] Image Resizing and Manipulation with Query String


mdmontesinos:
### Oqtane Info

Version - 5.2.3
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the enhancement

Similar to how images can be resized and manipulated with the File API endpoint by id and route parameters (i.e):

```
api/file/image/1010/1500/1500/crop/center/transparent/0/false
```

It would interesting for some use cases when the id is not available to be able to achieve the same, retrieving the image by its path and with a query string. This is the way most libraries and websites I've seen work:

```
files/folder/subfolder/myImage.jpg?w=1500&h=1500&mode=crop&position=center&background=transparent&rotate=0&recreate=false
```

This should be updated in the GET action in the [Files Page](https://github.com/oqtane/oqtane.framework/blob/dev/Oqtane.Server/Pages/Files.cshtml.cs). Also, the generated image should be cachable with the ETAG, or another mechanism, to avoid regenerating it multiple times and reducing CPU and memory consumption.

tvatavuk:
We've been using an [Imageflow](https://www.imageflow.io/)-based extension for Oqtane in our [2sxc](https://github.com/2sic/2sxc) module to handle image resizing and manipulation via query strings. Imageflow is an outstanding solution—it's extremely optimized, super fast, and packed with features.

The extension can also be installed as a standalone module in Oqtane, making it versatile for various use cases. 

Here are some resources you might find helpful:

- [Imageflow Query String API Documentation](https://docs.imageflow.io/querystring/introduction.html#querystring-api)
- [2sic's Oqtane Imageflow Extension on GitHub](https://github.com/2sic/oqtane-imageflow)
- [2sic's Oqtane Imageflow Module](https://www.oqtane.net/!/tosic.imageflow.oqtane/2sxc/?type=module&price=free)

mdmontesinos:
@tvatavuk Thanks for the info about the Imageflow-based Oqtane module! Is it free for commercial use or do I need to purchase an imageflow license? Is it compatible with Oqtane > 5? (the last commit was on May)

Also, I still believe there should be an in-framework solution, as Oqtane already bundles with SixLabors.ImageSharp and provides the [functionality](https://github.com/oqtane/oqtane.framework/blob/dev/Oqtane.Server/Controllers/FileController.cs#L746), just in a different format. In this way, there's no need for extra dependencies or modules.

[...]

PR #4723, opened 2024-10-13
Image Resizing and Format via QueryString in Files Page (enhances #4692)

mdmontesinos:
- Extracts the image creation into a service
- Refactors Files Page GET action for better readability and cyclomatic complexity
- Allows changing the image encoding: webp or png (default)
- Includes all parameters in the Etag calculation

Related to #4692

### DIFF
--- a/Oqtane.Server/Controllers/FileController.cs
+++ b/Oqtane.Server/Controllers/FileController.cs
@@ -685,14 +685,16 @@ namespace Oqtane.Controllers
                     {
                         if (!bool.TryParse(recreate, out _)) recreate = "false";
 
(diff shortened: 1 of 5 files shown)


sbwalker:
@mdmontesinos as part of the work on caching I discovered that the .NET string.GetHashCode() method is non-deterministic ie. each time an application is restarted, the  GetHashCode() method will return a different value. This is not optimal for eTags as their value should remain consistent until a resource changes. The logic which was introduced in this PR appears to be  attempting to create a unique eTag value based on the various parameters which are passed to the page. I am going to modify this logic so that eTag values are deterministic.

sbwalker:
#5032 adds deterministic hashing and also simplifies logic by creating the hash based on the querystring value

mdmontesinos:
My apologies for the oversight!

PR #4730, opened 2024-10-14
Localization Cookie Static Conditional HttpOnly=True

### LINKED ISSUE #4714: [BUG] 5.2.3 - Cookies with HttpOnly=True unable to be accessed while using Interactive Render Mode (SignalR + Javascript Interop).
### Oqtane Info

Version - 5.2.3
Render Mode - Interactive
Interactivity - Server/WebAssembly
Database - SQL Server

### Describe the bug
LanguageSwitcher does not switch language (Interactive Server / Interactive Webassembly) because of httpOnly culture cookie set to `true` and SignalR using JavaScript interop not allowing Oqtane to able to access these cookies directly.

Taken from discussion #4703

### Expected Behavior
Culture (and visitor?) Cookie can be accessed while 

### Steps To Reproduce
Attempt switching languages via the LanguageSwitcher.razor control in the menu while using interactive render mode site setting.

### Anything else?
Set 'HttpOnly=false' if render mode for the site is `Interactive` 

Works OK while in Static Render Mode.

Can these cookies be created using middleware to avoid JavaScript?

https://learn.microsoft.com/en-us/dotnet/api/microsoft.net.http.headers.setcookieheadervalue.httponly?view=aspnetcore-8.0

https://learn.microsoft.com/en-us/dotnet/api/microsoft.net.http.headers.setcookieheadervalue?view=aspnetcore-8.0

thabaum:
Fix #4714 
#4703 
#4728

This adds a conditional checks in `LangaugeSwitcher.razor` component and `App.razor` component to check if current rendermode is static and set `HttpOnly=True` if it is for the culture cookie when created.  Otherwise sets to false.

### DIFF
--- a/Oqtane.Client/Themes/Controls/Theme/LanguageSwitcher.razor
+++ b/Oqtane.Client/Themes/Controls/Theme/LanguageSwitcher.razor
@@ -61,7 +61,7 @@
                     Expires = DateTimeOffset.UtcNow.AddYears(365),
                     SameSite = Microsoft.AspNetCore.Http.SameSiteMode.Lax, // Set SameSite attribute
                     Secure = true, // Ensure the cookie is only sent over HTTPS
-                    HttpOnly = false // cookie is updated using JS Interop in Interactive render mode
+                    HttpOnly = PageState.RenderMode == RenderModes.Static // sets the cookie HttpOnly=true only if the rendermode is not updated using JS Interop in Interactive render mode
                 });
 
             }
--- a/Oqtane.Client/Themes/Controls/Theme/LanguageSwitcher.razor
+++ b/Oqtane.Client/Themes/Controls/Theme/LanguageSwitcher.razor
@@ -61,7 +61,9 @@
                     Expires = DateTimeOffset.UtcNow.AddYears(365),
                     SameSite = Microsoft.AspNetCore.Http.SameSiteMode.Lax, // Set SameSite attribute
                     Secure = true, // Ensure the cookie is only sent over HTTPS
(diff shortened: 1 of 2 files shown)


sbwalker:
I am not going to merge this PR as there is the real possibility that if a site starts with Static render mode (the default) and later changes to Interactive - the users of the site would run into  issues setting the language cookie and would need to manually remove the cookie from their browser in order to get it to work properly.

maurocavallin:
> I am not going to merge this PR as there is the real possibility that if a site starts with Static render mode (the default) and later changes to Interactive - the users of the site would run into issues setting the language cookie and would need to manually remove the cookie from their browser in order to get it to work properly.

I can confirm: whit this PR a cookie that starts httpOnly in Static mode stay unchanged when switching to interactive, then the user have to manullay remove cookie to go on.

I agree language cookie can be httpOnly=false for any mode, has no sensitive content, even being manipulated by any allowed actor on client side, it can make no harm.

thabaum:
Could we expire the cookie, then create a new one?

sbwalker:
@thabaum If a cookie has HttpOnly disabled it does not mean your site is vulnerable to attack. HttpOnly is only an extra layer of defense (for those browsers that actually support it). XSS requires an initial attack vector - the injection of executable JavaScript into a page. If your site allows a malicious user to inject JavaScript then they could include a script which can read the value of cookies. HttpOnly prevents scripts from reading cookie values. So it's useful as an additional safeguard for authentication cookies or cookies which store other confidential information. But if your site allows malicious scripts to be injected then you obviously have much bigger problems to worry about than just cookies.

PR #4782, opened 2024-10-24
Fixes #4781: Adds Edit + Add Page ScrollToTopPage() On Settings Tab Form Error Messages

### LINKED ISSUE #4781: [ENH] Page Management: Scroll To Top Of Page on Error for Edit and Add Page
### Oqtane Info

Version - 6.0.0
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the enhancement
Currently, when an error occurs during the editing or adding of pages in the control panel/dashboard, the user remains at their current scroll position. This can lead to confusion, as users may not immediately see the error messages displayed at the top of the page.

### Steps To Reproduce
1. Navigate to the Edit Page or Add Page section of the control panel/dashboard.
2. Enter invalid data or trigger an error (e.g., missing required fields) and click save (be sure to have a small enough window so that you do not see the top of the screen/page).
3. Observe the error message displayed at the top of the page without any scrolling action.

### Expected Behavior
- When an error occurs during the save operation, the page should automatically scroll to the top to bring the error messages into view.

### Actual Behavior
- The page remains at the user's current scroll position, potentially causing the user to miss important error messages.

thabaum:
Fixes #4781: 
- Adds Edit + Add page ScrollToTopPage() on error messages.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Pages/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Pages/Add.razor
@@ -473,10 +473,6 @@
                 else
                 {
                     AddModuleMessage(Localizer["Message.Required.PageInfo"], MessageType.Warning);
-                    if (string.IsNullOrEmpty(_name))
-                    {
-                        AddModuleMessage(Localizer["Message.Required.PageName"], MessageType.Warning);
-                    }
                     await ScrollToPageTop();
                 }
 
@@ -491,10 +487,6 @@
         else
         {
             AddModuleMessage(SharedLocalizer["Message.InfoRequired"], MessageType.Warning);
-            if (string.IsNullOrEmpty(_name))
-            {
-                AddModuleMessage(Localizer["Message.Required.PageName"], MessageType.Warning);
-            }
             await ScrollToPageTop();
         }
     }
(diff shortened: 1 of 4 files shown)


sbwalker:
@thabaum the page name validation is already handled in the existing code:

There is a required attribute on the field:

```
<input id="name" class="form-control" @bind="@_name" maxlength="50" required />
```

and a form validator in the Save method:

```
if (await interop.FormValid(form))
```

This handles all of the required fields.

thabaum:
@sbwalker I guess I was thinking about this error message

'
                    AddModuleMessage(Localizer["Message.Required.PageInfo"], MessageType.Warning);
                    if (string.IsNullOrEmpty(_name))
                    {
                        AddModuleMessage(Localizer["Message.Required.PageName"], MessageType.Warning);
                    }
                    await ScrollToPageTop();
'

Just seemed more friendly to have the error message tell what is the issue instead of looking for a red x in the form for the blank page which I think is the only one you can really mess up beside the effective date on the form.
"You Must Provide Page Name, Theme, and Container"

This message however does not get fired off when the page is blank.

Instead you get this one:
`AddModuleMessage(SharedLocalizer["Message.InfoRequired"], MessageType.Warning);`
"Please Provide All Required Information"

A more generic message for this error which I understand is probably preffered.

I can remove the message enhancements I suppose it is a bit more simple to keep it just a message that makes a user review the entire form.

thabaum:
@sbwalker I went ahead and removed the added error message for blank page name.  This just focuses on scroll to top of page functionality on error messages.

sbwalker:
@thabaum the reason why I took so long to merge this PR is because it contains compilation errors - which means that the changes were likely implemented manually outside of an IDE and never tested. I went ahead and merged the PR and will resolve the errors and test now. In the future it would be best to submit PRs which build successfully.

The compilation error was:

The 'await' operator can only be used within an async method. Consider marking this method with the 'async' modifier and changing its return type to 'Task'.

PR #4798, opened 2024-11-03
Use Microsoft.IO.RecyclableMemoryStreamManager

### LINKED ISSUE #4797: [ENH] Use Microsoft.IO.RecyclableMemoryStream
### Describe the enhancement
In the code, new `MemoryStream()` could be optimized with pooling by using the [Microsoft.IO.RecyclableMemoryStream](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream) nuget.

adamradocz:
Closes #4797 

Initial commit switching from `MemorySteam` to `Microsoft.IO.RecyclableMemoryStream`.
I haven't changed the `LoadFromStream` functions, as I don't know how that works, whether its inner stream should be disposed or not.

### DIFF
--- a/Oqtane.Client/Program.cs
+++ b/Oqtane.Client/Program.cs
@@ -15,6 +15,7 @@ using Microsoft.AspNetCore.Components.WebAssembly.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.JSInterop;
 using Oqtane.Documentation;
+using Oqtane.Helpers;
 using Oqtane.Models;
 using Oqtane.Modules;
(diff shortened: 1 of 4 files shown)


sbwalker:
Oqtane only uses MemoryStream when unzipping packages - which is a very limited use case and therefore does not result in any application performance issues. One of the items in the Oqtane Philosophy (https://www.oqtane.org/blog/!/20/oqtane-philosophy) is to minimize dependencies... and adding an additional dependency on Microsoft.IO.RecyclableMemoryStreamManager does not seem necessary.

PR #4799, opened 2024-11-04
Simplify InstallationController.cs

adamradocz:
Empty return is simplified.


sbwalker:
Although the GetZIP() returns a byte array, the client caller is expecting the byte array to be in a specific format ie. a ZIP archive. Returning an empty byte array will result in an exception in the client caller. So this "optimization" will actually cause an issue.

PR #4804, opened 2024-11-07
Fix #4778: update the tabstrip view after render.

### LINKED ISSUE #4778: [BUG] Page Management - Edit: When saving and there is an error, the Settings tab is selected but teh tab panel is not updating the UI.
### Oqtane Info

Version - 6.0.0
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Description
When attempting to save changes within a module (edit page settings), if an error occurs, the "Settings" tab becomes selected, but the corresponding tab panel does not update to reflect the latest user inputs. This leads to user confusion, as they see the "Settings" tab highlighted but the content remains unchanged.

### Steps to Reproduce
1. Navigate to the module with tabbed settings.  (Control Panel > Edit Page)
2. Make changes to the "Settings" input fields that create an error. (blank page name)
3. Change tabs to "Permissions" tab.
4. Attempt to save the changes.
5. Observe that the "Settings" tab is selected after the error, but the content in the UI for that tab does not refresh.

### Expected Behavior
- After encountering a save error, the "Settings" tab should display the most recent input values from the user.
- The tab panel should refresh to show the updated state of the inputs, even if a save error occurs.

### Actual Behavior
- The "Settings" tab appears active, but the tab panel retains the previous tab input instead of reflecting the current state.

### Anything else?

Also notice the 2 tab strip tab panels information when you switch from the "Settings" to "Theme Settings" for example.
image

zyhfish:


sbwalker:
@zyhfish the problem with this solution is that it only works withing an Interactive rendered component - whereas the TabStrip component can also be used within a Static rendered component. If possible it would best to avoid JavaScript to support both render modes.

zyhfish:
Hi @sbwalker , I think this will not be a problem in static render mode as it will full refresh the page when post back, currently this behavior is occur in the client side so can only correct it by javascript.

sbwalker:
Static render mode does not do a full page refresh in Blazor... it uses Enhanced Navigation to only updates the elements which were changed.

zyhfish:
Hi @sbwalker , please check the attached package to verify the static render mode. (you may need to rename it with nupkg extension)
[Owner.Module.Test.1.0.0.zip](https://github.com/user-attachments/files/17691364/Owner.Module.Test.1.0.0.zip)

PR #4805, opened 2024-11-07
parsedCulture capture null

leigh-pointer:
Returns null on a fresh installation of windows os and Oqtane.

`Shared.CookieRequestCultureProvider.ParseCookieValue(culture).Culture.Name;`

### DIFF
--- a/Oqtane.Server/Components/App.razor
+++ b/Oqtane.Server/Components/App.razor
@@ -215,7 +215,11 @@
                     // set language for page
                     if (!string.IsNullOrEmpty(culture))
                     {
-                        _language = Shared.CookieRequestCultureProvider.ParseCookieValue(culture).Culture.Name;
+                        var parsedCulture = Shared.CookieRequestCultureProvider.ParseCookieValue(culture);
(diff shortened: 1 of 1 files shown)


sbwalker:
@leigh-pointer I am confused why this logic would fail. The ParseCookieValue() method simply parses a culture string (ie. "c=en|uic=en") into a RequestCulture object. If the culture string is null or empty string it will not attempt to parse it. So I am wondering what value the culture string contains in your installation? Its important to be able to understand this in order to validate that this is the correct solution.

```
                    if (!string.IsNullOrEmpty(culture))
                    {
                        _language = Shared.CookieRequestCultureProvider.ParseCookieValue(culture).Culture.Name;
                    }
```

sbwalker:
@leigh-pointer I deleted the culture cookie in browser and reloaded the page and realized that in this scenario the system loads the default Language for the site (ie. "en") however it then needs to convert it into a Culture string (ie. "c=en|uic=en"). PR #4811 resolves this issue. Thank you for reporting!

Issue #4866, opened 2024-11-25
[DOCS] Move Release History to docs


iJungleboy:
### Oqtane Info

Version - 6.0

iJungleboy:
1. docs updated to contain the latest history
2. pull request to shorten readme on oqtane repo

sbwalker:
I would prefer to not remove all 5.x releases from the repo home page at this point.

PR #4867, opened 2024-11-25
Move change history up to v6 to Docs #4866

iJungleboy:
https://github.com/oqtane/oqtane.framework/issues/4866


sbwalker:
I would prefer to not eliminate all 5.x change history from the repo home page, as it makes it appear that this project is not continuously maintained

PR #4828, opened 2024-11-14
Fix for Tabpanel is not updating the UI. #4778

leigh-pointer:
Updated the TabStrip and TabPanel so when the ActiveTab parameter is changed the corresponding TabPane is selected.

PR #4871, opened 2024-11-26
Rework for Tabstrip regression issue

leigh-pointer:
Fix for TabPanel is not updating the UI. [#4778](https://github.com/oqtane/oqtane.framework/issues/4778) [#4828](https://github.com/oqtane/oqtane.framework/issues/4828)

### DIFF
--- a/Oqtane.Client/Modules/Admin/Modules/Settings.razor
+++ b/Oqtane.Client/Modules/Admin/Modules/Settings.razor
@@ -302,13 +302,13 @@
             }
             else
             {
-                _activetab = "Settings";
+                //_activetab = "Settings";
                 AddModuleMessage(Localizer["Message.Required.Title"], MessageType.Warning);
             }
         }
         else
         {
-            _activetab = "Settings";
+            //_activetab = "Settings";
             AddModuleMessage(SharedLocalizer["Message.InfoRequired"], MessageType.Warning);
         }
     }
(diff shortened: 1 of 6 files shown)


sbwalker:
@leigh-pointer there are a lot of third party modules which use the TabStrip component... I am concerned that these recent changes will result in regression issues / broken functionality.

sbwalker:
After more testing (without applying #4871) there are also UX issues with adding new pages and editing existing pages. I am going to rollback #4828 as there are too many ripple effects to this change.

leigh-pointer:
@sbwalker i have been extremely careful about backward compatibility. I have tested with the framework and external modules.

sbwalker:
@leigh-pointer if you select Add Page from the Control Panel and then select to change the Parent field to a different page, all of the fields on the Settings tab disappear. Do you see this issue as well?

leigh-pointer:
@sbwalker no, everything works fine.

sbwalker:
after applying this PR... if you select Add Page from the Control Panel and then select to change the Parent field to a different page, the Settings tab is no longer highlighted:

image

sbwalker:
#4872 also seems to have been caused by the recent changes to the TabStrip component.

sbwalker:
@leigh-pointer I rolled back #4871 and #4828 as there are too many regression problems

PR #4878, opened 2024-11-27
User Settings should only be accessible to individual users or administrators

sbwalker:


sbwalker:
A CVE 2024-55470 has been attributed to this pull request:

"Oqtane Framework 6.0.0 is vulnerable to Incorrect Access Control. By manipulating the entityid parameter, attackers can bypass passcode validation and successfully log into the application or access restricted data without proper authorization. The lack of server-side validation exacerbates the issue, as the application relies on client-side information for authentication."

However both the severity level and description of the vulnerability is inaccurate. Specifically, "attackers can bypass passcode validation and successfully log into the application" is 100% false. The security issue was investigated and feedback was provided to the contributor of the issue, however they proceeded to publicize the vulnerability without justification.

PR #4892, opened 2024-12-02
Fix #4885: set the authorization header to http client.

### LINKED ISSUE #4885: [BUG] JWT Authentication broken after 5.2.2 SecurityStamp changes
### Oqtane Info

Version - 5.2.2 to 6.0.0
Render Mode - Interactive & Static
Interactivity - Server
Database - SQL Server

### Describe the bug
Cannot use JWT bearer token for authentication.  Returned user is always null.  Issue started with addition of SecurityStamp in version 5.2.2.

### Expected Behavior
Should be able to issue JWT bearer token and use for API authentication.

### Steps To Reproduce
- Perform Fresh install
- Go to User Management > Settings > Token Settings
- Set "Issuer" value & save
- Select "Create Token"

Use created token from external application (Postman for easy testing).  If running in debugger, you will see exception at line 71 of JwtMiddleware.cs.   Problem is line 61/62 always return null for user.  External reply is always a 200 response.

### Anything else?
I created fresh installs of 5.2.0, 5.2.1, 5.2.2, 5.2.3, 5.2.4 & 6.0.0.  This functionality works fine for 5.2.0 & 5.2.1, but fails in all versions from 5.2.2 and on.

image

image

zyhfish:


sbwalker:
@zyhfish I do not see how this PR has anything to do with the introduction of the SecurityStamp? The issue reports that JwT Tokens worked fine prior to the introduction of SecurityStamp... so I would expect that the fix would be related to SecurityStamp.

zyhfish:
Hi @sbwalker , with this fix I can request to the private pages with JWT token, there have 2 issues:
1. in the JWT middleware, it called to wrong GetUser overload method;
2. in the API requests, the http client didn't send the authorization header so that API request failed when getting user information.

sbwalker:
@zyhfish the standard (internal) HttpClient should NOT pass a bearer token, as it relies on standard cookies for authentication

sbwalker:
@zyhfish #4885 could be resolved by simply changing this 1 line:

```var user = _users.GetUser(userid, alias.SiteId); // cached```

to

```var user = _users.GetUser(int.Parse(userid), alias.SiteId); // cached```

zyhfish:
Hi @sbwalker ,
we already parsed the user id at line #48, so don't need to parse again.

the other scenario is if you request to a page which only available to authenticated users, you will be redirected to  the login  page without the change in the http client area.

sbwalker:
@zyhfish generally it is a good practice to minimize code changes when fixing bugs so that the root cause of the issue is easy to identify, and there is less chances of introducing other regression issues to surrounding logic. So I would prefer if we simply changed the one line of code in this case to resolve #4885 

```var user = _users.GetUser(int.Parse(userid), alias.SiteId); // cached```

In regards to the other changes in this PR to OqtaneServiceCollectionExtensions.cs, they should be removed as they are not related to the issue reported in #4885. As mentioned previously, #4885 is related to accessing the back-end public API using a JwT token - not accessing the front-end UI. The following diagram describes the backend-for-frontend (BFF) security model used in Oqtane:

image

When accessing the front-end UI, cookies are used for auth. When accessing remote downstream APIs, Jwt tokens are used. And when remote clients access the Oqtane API, Jwt tokens are used.

zyhfish:
thx so much @sbwalker , I will close this PR as #4894 already submitted.

PR #4895, opened 2024-12-04
Add ShowEditMode parameter to ControlPanel to allow hiding the Edit Mode toggle button

### LINKED ISSUE #4883: [ENH] Oqtane ControlPanel component forces edit button unto page
### Oqtane Info

Version - 5.x up to 6.0
Render Mode - all
Interactivity - Server
Database - SQL Server

### Describe the bug

The razor component `ControlPanel` should be responsible for making the control panel available.
But it also forces a visible edit button unto the page, which is not desirable, because this makes it impossible to create your own.

Here's an example: in both cases there is a customized toolbar (custom languages, custom buttons) but the forced edit still appears (languages could be disabled).

image

image

### Expected Behavior

Either

1. The edit button is a separate control which can be added as needed
2. or the ControlPanel has a `ShowEditToggle` parameter (just like `ShowLanguageSwitcher`) to allow not having that button

My recommendation _would_ be to keep concerns separate, and not combine these controls. But because I assume "the ship has sailed" we should probably just add the toggle. 

### Anything else?

I would be happy to contribute the fix, but must be sure that it would be accepted, and that the naming for this Parameter is according to your expectations.

tvatavuk:
# Fix #4883 [ENH] Oqtane ControlPanel component forces edit button unto page

This enhancement addresses issue #4883 by introducing a ShowEditMode parameter to the ControlPanel component. This allows theme developers to hide the Edit Mode toggle button when desired.

### Changes Made
•	Added a new `ShowEditMode` parameter to `ControlPanel.razor` with a default value of `true` to maintain existing behavior.
•	Updated the `ControlPanel` logic to conditionally display the Edit Mode toggle button based on the `ShowEditMode` value.

### Testing Performed
Extensive testing was conducted to ensure this change works as intended across various scenarios.

#### Default Themes
- **Themes Tested**: `Default` themes of **Oqtane** and **Blazor**.
- **Users**: Anonymous and Admin.
- **Render Combinations**:
  - Static, Server
  - Static, Client
  - Static, Auto
  - Interactive, Server
  - Interactive, Client
  - Interactive, Auto
- **Results**:
  - **Anonymous Users:** Edit Mode toggle button is not visible (as expected).
  - **Admin Users:** Edit Mode toggle button is visible (current behavior preserved).

#### Modified Themes with `ShowEditMode="false"`
- **Created** new theme variations named `DefaultHideEditToggle.razor` for both **Oqtane** and **Blazor** themes.
- **Modification**:
  ```razor
    <ControlPanel LanguageDropdownAlignment="right" ShowEditMode="false" />
  ```
- **Users**: Anonymous and Admin.
- **Render Combinations**: Same as above.
- **Results**:
  - All Users: Edit Mode toggle button is not visible.
  - Confirmation: Enhancement allows hiding the Edit Mode toggle as intended.

### Conclusion
The addition of the `ShowEditMode` parameter provides theme developers with the flexibility to hide the Edit Mode toggle button when necessary. The default behavior remains unchanged for existing themes, ensuring backward compatibility.

**Note**: All tests confirm that the enhancement functions correctly without affecting current Oqtane behavior in default themes.

### DIFF
--- a/Oqtane.Client/Themes/Controls/Theme/ControlPanel.razor
+++ b/Oqtane.Client/Themes/Controls/Theme/ControlPanel.razor
@@ -9,7 +9,7 @@
     <LanguageSwitcher ButtonClass="@ButtonClass" DropdownAlignment="@LanguageDropdownAlignment" />
 }
 
-@if (_showEditMode || (PageState.Page.IsPersonalizable && PageState.User != null && UserSecurity.IsAuthorized(PageState.User, RoleNames.Registered)))
+@if (ShowEditMode && (_showEditMode || (PageState.Page.IsPersonalizable && PageState.User != null && UserSecurity.IsAuthorized(PageState.User, RoleNames.Registered))))
 {
     <form method="post" class="app-form-inline" @formname="EditModeForm" @onsubmit="@(async () => await ToggleEditMode(PageState.EditMode))" data-enhance>
         <input type="hidden" name="@Constants.RequestVerificationToken" value="@SiteState.AntiForgeryToken" />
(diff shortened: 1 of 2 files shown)


iJungleboy:
@sbwalker would be great if this could make it into 6.0.1 ;)

Issue #4897, opened 2024-12-05
[ENH] Dynamic TargetFramework Support for Debug Scripts


leigh-pointer:
### Oqtane Info

Version -6.0.0
Render Mode - any
Interactivity - any
Database - any

### Describe the enhancement
Implement dynamic TargetFramework support for our build and debug scripts across different project types (modules and themes) to improve build flexibility and future-proof our development process.

### Anything else?
**Key Improvements**
- Dynamic Framework Versioning: Replace hardcoded framework versions with a dynamic TargetFramework parameter
- Cross-Platform Support: Update both Windows (.cmd) and Unix-like (.sh) debug scripts
- Simplified Maintenance: Reduce manual updates when changing .NET versions

**Benefits**
- Easier framework version upgrades
- Reduced risk of manual configuration errors
- Improved build process consistency
- Better support for multi-targeting scenarios

iJungleboy:
@leigh-pointer This is quite awesome!

Any chance you could also parameterize the name of the module, so that the cmd files could be identical in all projects?

I think that would be another great step forward, since it would remove any variance in the copies of the scripts.

Having identical scripts would also make it easier to "update" a script to a newer model, since there would be nothing inside it that is project specific.

PR #4909, opened 2024-12-13
Implement Dynamic ProjectName Parameter Across Build Process"

leigh-pointer:
This pull request introduces a dynamic ProjectName parameter across our build process, enhancing flexibility and reducing hardcoding in our module and theme development workflow. The changes affect the project file, NuSpec file, command scripts.


iJungleboy:
This is truly awesome!

Just what I hoped for.

Thanks @leigh-pointer !!!

iJungleboy:
Now probably an small add-on would be great to update the theme template to do the same, so future modules/themes are "pure" 😉

iJungleboy:
Just for the docs: related to #4897 and #4898

PR #4928, opened 2024-12-18
[Enh] Displays page list according to the sort order preferences applied to pages.

W6HBR:
Updates \Admin\Pages Add, Edit & Index to display pages according to the sort order preferences applied to pages.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Pages/Add.razor
+++ b/Oqtane.Client/Modules/Admin/Pages/Add.razor
@@ -252,6 +252,7 @@
         try
         {
             _pages = await PageService.GetPagesAsync(PageState.Site.SiteId);
+            _pages = _pages.OrderBy(item => item.ParentId).ThenBy(item => item.Order).ToList();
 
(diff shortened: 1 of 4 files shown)


W6HBR:
FYI - the last file here was supposed to be a separate pull request - Oqtane.Client\Modules\Admin\Roles\Index.razor.
(I didn't realize it would add in to the previous pull request for the Page Mgmt updates.)

sbwalker:
@W6HBR the issue with the sort order for Pages was caused by PR #4412. Prior to this optimization, various components were relying on the PageState.Pages collection. The PageState properties had to be optimized for performance reasons in Static Blazor which meant that module components needed to load the Pages collection from the PageService rather than relying on PageState.Pages. However, the PageService GetPages method does not return the list of pages in the same order as the PageState.Pages collection. This difference in behavior is what caused the issue with sort order in Page Management. I am going to close this PR as the better way to resolve this problem is included in #4931.

W6HBR:
Copy that... thanks.

PR #4932, opened 2024-12-19
[Enh] Update PermissionGrid.razor to be visually consistent with other editing screens/controls

W6HBR:
- Moves checkbox controls to the left side and role name/descriptions to the right.
- The movement of the checkbox controls is more consistent with the layout of controls on the Page, User, Profile and Role Management pages (e.g. action controls on left, description to the right).
- Adds role description column to display to be more informational.
- This also makes it easier to visually align the controls with the name/description on full screen monitors.

### DIFF
--- a/Oqtane.Client/Modules/Controls/PermissionGrid.razor
+++ b/Oqtane.Client/Modules/Controls/PermissionGrid.razor
@@ -15,22 +15,24 @@
             <table class="table table-borderless">
                 <tbody>
                     <tr>
-                        <th scope="col">@Localizer["Role"]</th>
                         @foreach (var permissionname in _permissionnames)
                         {
 							<th style="text-align: center; width: 1px;">@((MarkupString)DisplayPermissionName(permissionname).Replace(" ", "<br />"))</th>
 						}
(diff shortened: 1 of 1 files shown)


W6HBR:
Sample screens:
(before)
image

(after)
image

W6HBR:
Here's a sample screen in actual use for a site with multiple roles that control access to modules.

image

sbwalker:
@W6HBR I am not going to merge this change for a number of reasons:

1. The admin UI is trying to minimize the amount of information displayed in the default view so that it is possible to be used on a mobile device. Adding the role description does not align with this goal.
2. In addition to roles, the permission grid also allows you to define permissions for individual users. These are displayed beneath the roles and display the users display name. It would be best if the information for roles and users is consistent.
3. The permission grid is not always displayed in its own dedicated tab. It is also often displayed on edit forms as a standard form field. In this layout it is very important that it does not consume too much horizontal space.
4. I personally prefer the checkboxes to be on the right, as they are configuration not actions. Oqtane generally puts actions (ie. buttons) on the left and displays data entry/configuration options to the right

PR #4942, opened 2024-12-20
fix reload script to use static array rather than a live HtmlCollection

sbwalker:


sbwalker:
this is necessary to avoid an infinite loop which can happen if a script that is being reloaded contains logic to add script elements to the page

Issue #4929, opened 2024-12-19
[BUG] Deploy to Azure Button


ubidev:
### Oqtane Info

Version - 6.0.0
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the bug
I just successfully ran the "deploy to azure" button BUT when I go to the https://websitename.azurewebsites.net URL I get a default azure website status page instead of the oqtane initialization wizard

![image](https://github.com/user-attachments/assets/96373bae-2524-4094-a671-0ce6b8d58386)

### Expected Behavior
once deployed I should get the oqtane initialize wizard

### Steps To Reproduce
click the "deploy to azure" button
fill the form with acceptable values
create
wait for success
clic the website provided address

### Anything else?
I know there is a manuall way to do this documented here: https://blazorhelpwebsite.com/ViewBlogPost/1 but I really need that wizard to work because I plan to demonstrate/use oqtane during classes to noobs

Thanks in advance for taking the time. This project is simply Amazing

sbwalker:
@vnetonline I believe you were able to resolve the issues in the Deploy to Azure functionality in the past... are you able to investigate the latest issues?

vnetonline:
Yup I will have a look and investigate and fix

vnetonline:
I have recorded a FREE course on how to setup Oqtane on Local IIS 

[https://learnoqtane.com/course-page/!/4/how-to-setup-oqtane-on-local-iis](https://learnoqtane.com/course-page/!/4/how-to-setup-oqtane-on-local-iis)

If you are not a member of [LearnOqtane.com](https://LearnOqtane.com) ..  you can [Join Now](https://LearnOqtane.com/join-now) for FREE

vnetonline:
The Button to deploy to Azure is working properly, however when it has finished getting the source files from the master it fails because of a Build error see below 

`Could not load file or assembly 'System.Text.Json, Version=8.0.0.4, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified. [C:\home\site\repository\Oqtane.Server\Oqtane.Server.csproj]`

```
NuGet Config files used:
      D:\DWASFiles\Sites\#120241221-1542-amazing\AppData\NuGet\NuGet.Config
  
  Feeds used:
      https://api.nuget.org/v3/index.json
  
  Installed:
[...]

PR #4953, opened 2024-12-23
add back System.Text.Json to Shared project (#4929)

sbwalker:
This ensures an older version of System.Text.Json is not pulled in via a transitive dependency


ADefWebserver:
This works? I had this problem with another project and tried this but I still kept getting the warnings :(

Issue #4955, opened 2024-12-29
[ENH] Implement the ManagePackageVersionsCentrally feature to enable centralized version management


leigh-pointer:
### Oqtane Info

Version - 6.0.1
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the enhancement
Implement the ManagePackageVersionsCentrally feature to enable centralized version management for NuGet packages across multiple projects within the Oqtane solution. This feature allows for consistent versioning of dependencies across all projects, simplifying the process of managing and updating package versions. The enhancement will:

- Enable central management of NuGet package versions through a Directory.Packages.props file.
- Eliminate the need for individual package version definitions in each project file, reducing version discrepancies and maintenance overhead.
- Improve consistency and reduce errors when upgrading package versions across multiple projects in a solution.

### Anything else?
This enhancement would help streamline the development process, especially for our solution with multiple projects that depend on the same set of packages. Proper implementation would prevent version conflicts and improve dependency management in Oqtane-based projects.

sbwalker:
@leigh-pointer is the ManagePackageVersionsCentrally feature only supported in Visual Studio - or is it also supported in other IDEs? I am asking because one of the overarching goals in Oqtane is to support multiple IDEs.

leigh-pointer:
@sbwalker ManagePackageVersionsCentrally is not exclusive to Visual Studio. This feature is supported across various .NET development environments and tools.

I identify several development environments that support this feature:

1. Visual Studio
2. JetBrains Rider
3. Visual Studio for Mac (implied support for solution-level NuGet package management)
4. Command-line tools:
   - dotnet CLI
   - NuGet CLI

It's important to note that while these IDEs and tools support Central Package Management, there might be variations in how they implement or interact with the feature. For example, some users have reported issues with the dotnet CLI, and updating packages after switching to central package management can be challenging in certain IDEs.

To use Central Package Management, you generally need:
- NuGet 6.2.0 or later
- .NET SDK 6.0.300 or later
- If using Visual Studio, version 2022 17.2 or later

sbwalker:
[...]

PR #4956, opened 2024-12-29
For #4955 Implement Centralized Package Version Management for Solution 

leigh-pointer:
This PR introduces Centralized Package Version Management for the solution by utilizing the Directory.Packages.props file. This change ensures consistency in package versions across all projects in the solution, streamlining the package update process and reducing the risk of version conflicts.
**Key Changes:** 

1. **Centralized Package Version Management:** Introduced Directory.Packages.props file to manage NuGet package versions at the solution level.

- All common NuGet package versions are now defined in this single file.
- Packages will be shared across all projects, ensuring the same version is used across the solution.

2. **Simplified Package Updates:** Now, package versions can be updated in one place, and all projects will automatically use the updated version when restoring packages.

3. **Reduced Manual Updates:** Eliminated the need to manually update NuGet package versions in each project's .csproj file.

**Benefits:**

- **Consistency:** Ensures that all projects within the solution use the same version of shared packages, preventing version mismatches and potential issues.
- **Efficiency:** Makes it easier to update NuGet package versions by only modifying the version in the Directory.Packages.props file, rather than in each individual project file.
- **Scalability:** Ideal for large solutions with multiple projects, as the package management is simplified and centralized.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,5 @@
+<Project>
+  <Target Name="Restore" BeforeTargets="Build">
+    <Exec Command="dotnet restore $(SolutionDir)Oqtane.sln" />
+  </Target>
+</Project>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,0 +1,35 @@
+<Project>
+  <PropertyGroup>
+    <!-- Enable Central Package Management -->
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- Centralized package versions -->
+    <PackageVersion Include="Microsoft.AspNetCore.Components.WebAssembly" Version="9.0.0" />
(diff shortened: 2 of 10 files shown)


thabaum:
After testing the build process things worked fine, however, when I went to update the package through Visual Studio Nuget Package Manager I ran into an error:

`There are 2 package sources defined in your configuration. When using central package management, please map your package sources with package source mapping (https://aka.ms/nuget-package-source-mapping) or specify a single package source. The following sources are defined: nuget.org, NuGetOrg`

Since I never used CPM before this was a new issue for me to resolve.

In order to resolve it I had to go to `%appdata%\Roaming\NuGet\NuGet.Config` and edit the file removing the extra NuGetOrg entry so it only had one package source value as shown below:

```
<configuration>
  <packageSources>
    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
  </packageSources>
</configuration>
```

We may need to add some troubleshooting support in documentation to help others down the road with the same issue who may be unfamiliar with using this and run into the same issue.

sbwalker:
I am closing this PR as I would prefer to not utilize centralized package management in Oqtane

PR #4981, opened 2025-01-13
Fix #4976: exclude deleted pages.

### LINKED ISSUE #4976: [BUG] In menudropdown we only see one level under the top-level.
### Oqtane Info

Version - 6.0.1

 In menudropdown we only see one level under the top-level.

image
 
After deleting the Pages TestLevel2 and TestLevel3 it still occour in the Parent combobox under Add Page.
 
image
image

zyhfish:


sbwalker:
@zyhfish this is not the standard approach we use in all other APIs - usually the caller filters deleted items. I would prefer to be consistent in the API approach unless there is compelling case to be made otherwise in this instance.

PR #4998, opened 2025-01-17
Update ActionDialog Add method to ensure consistent button sizing

leigh-pointer:
This PR introduces a new private method GetButtonSize() to enhance the consistency of button sizing within our UI. The method specifically checks for the presence of the "btn-sm" class in the Action Button's class list and applies the same sizing to the Cancel Button if found.

### DIFF
--- a/Oqtane.Client/Modules/Controls/ActionDialog.razor
+++ b/Oqtane.Client/Modules/Controls/ActionDialog.razor
@@ -24,7 +24,7 @@
                             {
                                 <button type="button" class="@Class" @onclick="Confirm">@((MarkupString)_iconSpan) @Text</button>
                             }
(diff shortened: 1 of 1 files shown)


sbwalker:
@leigh-pointer my preference for handling this scenario is to introduce new parameters for ConfirmClass and CancelClass which would enable a developer to override and specify their preferred CSS classes for each button. By default Class would be used for the non-modal button and would use "btn btn-primary", ConfirmClass would be used for the Confirm modal button and would use the same value as Class (unless overridden), and CancelClass would be used for the Cancel modal button and would use "btn btn-secondary" (unless overridden).

leigh-pointer:
@sbwalker changed as to your comment,

PR #5003, opened 2025-01-18
Update Oqtane.Server.csproj - MySQL deploy to Azure error

sdi2121:
Fixes Azure manual deployment build from local code build. May need an additional fix in MySQL library (Line 24 - MySQLDatabase.cs)


sdi2121:
@dotnet-policy-service agree

PR #5007, opened 2025-01-21
Fixes #5005: Add Browser Caching for Static Assets

### LINKED ISSUE #5005: [ENH] Browser Caching for Static Assets
### Oqtane Info

Version - 6.0.1
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the enhancement

As discussed in https://github.com/oqtane/oqtane.framework/discussions/4755#discussioncomment-11017118, Oqtane won't be able to take advantage of the new MapStaticAssets capability to improve performance for static assets loading.

Still, I believe Oqtane should try to include some of its functionalities, like setting caching headers so that browsers can cache them. Basically, adding "Cache-Control" header in the UseStaticFiles response, as stated in https://learn.microsoft.com/en-us/aspnet/core/fundamentals/static-files?view=aspnetcore-9.0#set-http-response-headers. 

This would be the easiest of the techniques to implement, it wouldn't affect framework performance and it improves Lighthouse performance metrics. For example, this was taken from the Oqtane.org site:

Image

For busting the cache, in https://github.com/oqtane/oqtane.framework/discussions/4745 it was discussed that a framework level hashing/fingerprinting was not necessary, and developers should add querystrings with versioning for this purpose.

I assume this caching should only affect static assets and exclude the ones uploaded via the framework File Manager, but it's still an improvement.

RahulKaushik007:
This PR addresses issue #5005 by adding the Cache-Control header for static file responses. This enhancement will allow browsers to cache static assets such as images, JavaScript, and CSS, improving performance by reducing the number of network requests on subsequent visits. The change is aimed at improving Lighthouse performance metrics for the application.

Testing:

Changes have been tested locally to confirm the Cache-Control headers are being correctly applied to static files.
Performance testing has shown an improvement in load times when the caching headers are applied.

Screenshot 2025-01-21 171303
Screenshot 2025-01-21 171116

Note:
Since this is my first PR, please let me know if there are any suggestions for improvements if I have missed anything or if I have done it totally wrong!!. I'm open to feedback!

### DIFF
--- a/Oqtane.Server/Startup.cs
+++ b/Oqtane.Server/Startup.cs
@@ -205,6 +205,7 @@ namespace Oqtane
                 ServeUnknownFileTypes = true,
                 OnPrepareResponse = (ctx) =>
                 {
+                    ctx.Context.Response.Headers.Append("Cache-Control", "public, max-age=604800");
                     var policy = corsPolicyProvider.GetPolicyAsync(ctx.Context, Constants.MauiCorsPolicy)
                         .ConfigureAwait(false).GetAwaiter().GetResult();
                     corsService.ApplyResult(corsService.EvaluatePolicy(ctx.Context, policy), ctx.Context.Response);
--- a/Oqtane.Server/appsettings.json
+++ b/Oqtane.Server/appsettings.json
@@ -2,10 +2,10 @@
   "RenderMode": "Interactive",
   "Runtime": "Server",
   "Database": {
-    "DefaultDBType": ""
+    "DefaultDBType": "Oqtane.Database.SqlServer.SqlServerDatabase, Oqtane.Database.SqlServer"
   },
   "ConnectionStrings": {
-    "DefaultConnection": ""
+    "DefaultConnection": "Data Source=(LocalDb)\\MSSQLLocalDB;AttachDbFilename=|DataDirectory|\\Oqtane-202501210838.mdf;Initial Catalog=Oqtane-202501210838;Integrated Security=SSPI;Encrypt=false;"
   },
   "Installation": {
     "DefaultAlias": "",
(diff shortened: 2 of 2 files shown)


sbwalker:
@RahulKaushik007 Oqtane contains a file server which handles static files that have a path prefix of "/files". The file server handles folder permissions, 404s, generating thumbnails, eTags, caching, etc... So any requests for files with a path prefix of "/files" need to be excluded from the Cache-Control policy you added in this PR.

RahulKaushik007:
@sbwalker Thank you for the clarification! I’ll update the PR to exclude requests with the /files path prefix from the general Cache-Control policy, ensuring it doesn’t interfere with Oqtane’s file server logic. I’ll test the changes and push the updates shortly. 

Please let me know if there’s anything else I should consider!

RahulKaushik007:
@mdmontesinos Since files under the `/Content` folder are being served via the `/files` route (handled as a `PhysicalFileResult`), excluding this folder from the cache-control policy, seems like the appropriate approach. Please let me know if you would like me to proceed with this change.

mdmontesinos:
@RahulKaushik007 I believe that would the right approach, excluding the "/Content" request path from the cache-control policy, but not "/files", as it's a razor page.

Still, @sbwalker has a deeper knowledge on how the system works, so he should be able to provide more details.

sbwalker:
@mdmontesinos upon further investigation you are correct that the "/files" route is excluded from static asset routing. And since it is possible (although not recommended) to serve files directly from within the "/content" folder (ie. without using the Files server) then it would be beneficial if they were included in the cache policy - so there is no need to exclude "/content".

sbwalker:
@RahulKaushik007 I will merge this PR and then make some modifications

PR #5059, opened 2025-02-05
fix: #5058 - chunk file upload concurrency limit and customizable site settings

### LINKED ISSUE #5058: [BUG] Server overload when uploading big files
### Oqtane Info

Version - 6.0.2
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the bug

This is a "continuation" of the discussion in #5044 (the original error was already fixed)

I've been exploring the code behind the upload process and analyzing further more and it's clear that I'm essentially performing a small DDoS against the server when uploading big files. Here's the Azure metrics during the upload:

Image

You can see the massive spikes in all I/O operations and, most worrying one, the response time goes from dozens of milliseconds to 2.83s (I've seen it even reach 6 seconds while testing). Basically all users trying to access the site will be waiting for the upload to complete.

> If you have a 40 MB file, the browser will break it into 1 MB chunks and make 40 POST requests to the server (this is to avoid the server limitations related to file upload size, timeouts, and leverage multi-threading)

I believe this is too extreme, as browsers and servers are capable of handling much bigger file sizes. Would you consider increasing the chunk size to something like 4MB (which I believe is the default max size of a http request in asp.net core)? (Perhaps even more) This would greatly reduce the number of requests. Even if you are "not" leveraging multi-threading as much, you also have to take into account the overhead for each request, like retrieving the folder from db, checking permissions, etc. I think this would increase overall performance.

> Since the browser can only support a specific number of threads, it will queue them and work its way through the queue as each POST request completes

This is also problematic. Browsers can handle a lot of requests if you have enough memory/cpu yourself. For example, I have 32 GB of RAM and 14 cpu cores (20 threads). So basically the browser is not queuing the requests but rather sending all of them almost at once. Here's the waterfall:

Image

As you can see, almost of all them started at the same time and are just waiting for the server response, which took almost 40s in some cases. This is 40s to upload a single 1MB file. Clearly the server was overloaded.

Perhaps the requests should be queued programatically rather than leaving it to the browser. For example, only have 5 active requests at a time, and when one of them finishes, queue the next one. This would greatly reduce the overload in the server.

To clarify, this was with no throttling, a network upload speed of 75 Mbps and Azure app service P0v3. And as I mentioned, it's a 40MB video. Imagine if I try to upload a 500 MB file, the server will probably crash with 500 nearly simultaneous post requests.

EDIT: I've been researching recommendations about file upload chunking, specially from file providers. For example, [Uploadcare](https://uploadcare.com/blog/file-uploader-by-uploadcare/#3.-chunking-files-over-20-mb) has determined after experimentation that the optimal upload speed is reached with chunks of 5MB each, with a batch of 4 chunks being uploaded simultaneously, and only for files over 20MB. [FastPix](https://www.fastpix.io/blog/how-to-upload-large-video-files-efficiently-using-chunking#how-to-implement-chunking-using-fastpix-upload-sdk) sets the minimum chunk size of 5MB, while it defaults to 16MB. [Transitload](https://transloadit.com/devtips/optimizing-online-file-uploads-with-chunking-and-parallel-uploads/#best-practices-and-optimization-tips) also recommends setting up a concurrency limit to the parallel uploads to avoid overloading the server. The [huge-uploader](https://github.com/Buzut/huge-uploader) js library defaults to a chunk size of 10MB

EDIT: Perhaps the UploadSpeed, ChunkSize and MaximumConcurrentRequests properties should be configurable via site settings, as the host user would know the capacity of the server and typical users to optimize the upload process

### Expected Behavior

ChunkSize should be customizable to allow developers to decide which option best suits their use case.

A concurrency limit to the upload process should be implemented to avoid server overload. It should also be customizable for the same reason as above.

### Anything else?

I'll try to submit a PR

mdmontesinos:
Fixes #5058 by implementing a concurrency limit for the file chunk upload process.

The concurrency limit is configurable with site settings. Also added settings for the chunk size. Both of them are validated in the js code to avoid negative values or huge chunk sizes.

The concurrency limit can be disabled by setting its value to 0.

The default behaviour remains the same to avoid being a breaking change (chunk size of 1MB and concurrency limit is 0)

The Interop method without the new parameters has been deprecated in favour of the new one (not deleted for retrocompatibility) but keeps the same behaviour if used.

### DIFF
--- a/Oqtane.Client/Modules/Admin/Site/Index.razor
+++ b/Oqtane.Client/Modules/Admin/Site/Index.razor
@@ -156,6 +156,18 @@
                         <input id="uploadableFileExt" spellcheck="false" class="form-control" @bind="@_uploadableFiles" />
                     </div>
                 </div>
+                <div class="row mb-1 align-items-center">
+                    <Label Class="col-sm-3" For="maxChunkSizeMB" HelpText="Enter the maximum size in MB of a chunk for file uploads" ResourceKey="MaxChunkSize">Maximum File Chunk Size (MB): </Label>
+                    <div class="col-sm-9">
+                        <input id="maxChunkSizeMB" type="number" min="1" max="50" step="1" class="form-control" @bind="@_maxChunkSizeMB" />
+                    </div>
+                </div>
+                <div class="row mb-1 align-items-center">
+                    <Label Class="col-sm-3" For="maxConcurrentChunkUploads" HelpText="Enter the maximum concurrent number of file chunk uploads. If set to 0, no concurrency limit is applied" ResourceKey="MaxConcurrentChunkUploads">Maximum Concurrent File Chunk Uploads: </Label>
+                    <div class="col-sm-9">
+                        <input id="maxConcurrentChunkUploads" type="number" min="0" step="1" class="form-control" @bind="@_maxConcurrentChunkUploads" />
+                    </div>
+                </div>
             </div>
         </Section>
         <Section Name="PageContent" Heading="Page Content" ResourceKey="PageContent">
@@ -432,6 +444,8 @@
     private string _textEditor = "";
     private string _imageFiles = string.Empty;
     private string _uploadableFiles = string.Empty;
+    private int _maxChunkSizeMB = 1;
+    private int _maxConcurrentChunkUploads = 0;
 
     private string _headcontent = string.Empty;
     private string _bodycontent = string.Empty;
@@ -530,6 +544,8 @@
                 _imageFiles = (string.IsNullOrEmpty(_imageFiles)) ? Constants.ImageFiles : _imageFiles;
                 _uploadableFiles = SettingService.GetSetting(settings, "UploadableFiles", Constants.UploadableFiles);
                 _uploadableFiles = (string.IsNullOrEmpty(_uploadableFiles)) ? Constants.UploadableFiles : _uploadableFiles;
+                _maxChunkSizeMB = int.Parse(SettingService.GetSetting(settings, "MaxChunkSizeMB", "1"));
+                _maxConcurrentChunkUploads = int.Parse(SettingService.GetSetting(settings, "MaxConcurrentChunkUploads", "0"));
 
                 // page content
                 _headcontent = site.HeadContent;
@@ -736,6 +752,8 @@
                         settings = SettingService.SetSetting(settings, "TextEditor", _textEditor);
                         settings = SettingService.SetSetting(settings, "ImageFiles", (_imageFiles != Constants.ImageFiles) ? _imageFiles.Replace(" ", "") : "", false);
                         settings = SettingService.SetSetting(settings, "UploadableFiles", (_uploadableFiles != Constants.UploadableFiles) ? _uploadableFiles.Replace(" ", "") : "", false);
+                        settings = SettingService.SetSetting(settings, "MaxChunkSizeMB", _maxChunkSizeMB.ToString(), false);
+                        settings = SettingService.SetSetting(settings, "MaxConcurrentChunkUploads", _maxConcurrentChunkUploads.ToString(), false);
 
                         await SettingService.UpdateSiteSettingsAsync(settings, site.SiteId);
 
--- a/Oqtane.Client/Modules/Controls/FileManager.razor
+++ b/Oqtane.Client/Modules/Controls/FileManager.razor
@@ -412,7 +412,7 @@
                     }
                     else
                     {
-                        await logger.LogInformation("File Upload Failed {Files}", uploads);
+                        await logger.LogError("File Upload Failed {Files}", uploads);
                         _message = Localizer["Error.File.Upload"];
                         _messagetype = MessageType.Error;
                     }
--- a/Oqtane.Client/Modules/Controls/FileManager.razor
+++ b/Oqtane.Client/Modules/Controls/FileManager.razor
@@ -121,6 +121,9 @@
     private MessageType _messagetype;
     private bool _uploading = false;
 
+    private int _maxChunkSizeMB = 1;
+    private int _maxConcurrentUploads = 0;
+
     [Parameter]
     public string Id { get; set; } // optional - for setting the id of the FileManager component for accessibility
 
@@ -173,6 +176,9 @@
         _fileinputid = "FileInput_" + _guid;
         _progressinfoid = "ProgressInfo_" + _guid;
         _progressbarid = "ProgressBar_" + _guid;
+
+        int.TryParse(SettingService.GetSetting(PageState.Site.Settings, "MaxChunkSizeMB", "1"), out _maxChunkSizeMB);
+        int.TryParse(SettingService.GetSetting(PageState.Site.Settings, "MaxConcurrentChunkUploads", "0"), out _maxConcurrentUploads);
     }
 
     protected override async Task OnParametersSetAsync()
@@ -383,7 +389,7 @@
                         StateHasChanged();
                     }
 
-                    await interop.UploadFiles(posturl, folder, _guid, SiteState.AntiForgeryToken, jwt);
+                    await interop.UploadFiles(posturl, folder, _guid, SiteState.AntiForgeryToken, jwt, _maxChunkSizeMB, _maxConcurrentUploads);
 
                     // uploading is asynchronous so we need to poll to determine if uploads are completed
                     var success = true;
--- a/Oqtane.Client/Resources/Modules/Admin/Site/Index.resx
+++ b/Oqtane.Client/Resources/Modules/Admin/Site/Index.resx
@@ -438,4 +438,16 @@
   <data name="System" xml:space="preserve">
     <value>System</value>
   </data>
+  <data name="MaxChunkSize.HelpText" xml:space="preserve">
+    <value>Enter the maximum size in MB of a chunk for file uploads</value>
+  </data>
+  <data name="MaxChunkSize.Text" xml:space="preserve">
+    <value>Maximum File Chunk Size (MB):</value>
+  </data>
+  <data name="MaxConcurrentChunkUploads.HelpText" xml:space="preserve">
+    <value>Enter the maximum concurrent number of file chunk uploads. If set to 0, no concurrency limit is applied</value>
+  </data>
+  <data name="MaxConcurrentChunkUploads.Text" xml:space="preserve">
+    <value>Maximum Concurrent File Chunk Uploads:</value>
+  </data>
 </root>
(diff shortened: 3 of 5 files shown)


sbwalker:
@mdmontesinos some feedback on the PR....

1. I have refactored the uploadFiles() JavaScript method to use fetch, promises, error handling, recursion, headers, etc... so I will need to merge your changes with mine
2. The approach I took for options was to include some additional parameters to the FileManager component so that developers have the ability to configure it based on the specific needs of their module. For all other cases, the default options should be optimized so that the file upload works well without requiring any configuration (as most users will use the defaults and not even understand what these settings mean).
3. If the system was optimized to handling throttling of file uploads by default, I am not sure if there needs to be a Maximum Concurrent File Chunk Uploads option. Too many options can often cause more harm than benefit in the long term - especially if the settings are very esoteric and require an intimate knowledge of the inner workings of the system to understand their purpose. 
4. If there is a broader need for UI settings (which I am not sure is true), Oqtane generally tries to include them in the Admin UI to which they relate ie. we are trying to avoid overloading the Site Settings UI with too many random options. For example all Site Settings related to Users are in User Management - Settings tab. So we would want to move these new settings related to Files to File Manager - Settings tab (this tab does not yet exist).

mdmontesinos:
@sbwalker Thanks for the feedback!

> I have refactored the uploadFiles() JavaScript method to use fetch, promises, error handling, recursion, headers, etc... so I will need to merge your changes with mine

I didn't really modify how the requests are generated, I just enqueued them and sent them with the concurrency limit, so there shouldn't be many problems.

> The approach I took for options was to include some additional parameters to the FileManager component so that developers have the ability to configure it based on the specific needs of their module

Right, this should work mostly fine, although it might require reconfiguring loads of modules that use it if there's no way to globally change these settings. Also, I wouldn't be able to modify the settings in the File Management admin page, so the problem of uploading a large video through there and saturating the server still remains.

> If the system was optimized to handling throttling of file uploads by default, I am not sure if there needs to be a Maximum Concurrent File Chunk Uploads option

I believe a concurrency limit is still required to avoid a server overload. Even if handling slow/throttled uploads, sending 40 simultaneous requests is probably too much.

> For example all Site Settings related to Users are in User Management - Settings tab. So we would want to move these new settings related to Files to File Manager - Settings tab (this tab does not yet exist).

I thought so as well, but as the settings tab for the file management wasn't created yet I wasn't sure.

sbwalker:
#5062 contains some changes for file upload...

- uses Fetch API rather than XHR
- Passes file part identifiers to server using headers rather than as part of the file name (to better hide implementation details from client)
- uses Promises to manage the entire process and return whether or not the upload process succeeded or failed
- since the status of the entire upload process is available on the client, there is no need for polling anymore (eliminating one of the most problematic aspects of the upload process)
- Promises also prevent the server from being flooded with requests
- chunk size is configurable (however increasing its value from the default of 1 MB can result in timeouts in low latency environments)

sbwalker:
@mdmontesinos the client-side implementation was changed significantly which will result in completely different run-time behaviors. Please test these latest changes and provide your feedback.

mdmontesinos:
@sbwalker The new upload process is looking much much better without the polling and using modern APIs like Fetch and Promises! Also, the progress bar works betters as it has a reliable way to determine when each upload has completed.

Let me add some comments to it:

- I tested the upload in localhost (as I can't test it in an Azure environment) by throttling the connection to Fast 4G and, after a while, I got this error in the event log:

```
System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at Microsoft.JSInterop.JSRuntime.InvokeAsync[TValue](Int64 targetInstanceId, String identifier, Object[] args)
   at Oqtane.Modules.Controls.FileManager.UploadFiles() in oqtane.framework\Oqtane.Client\Modules\Controls\FileManager.razor:line 390
```

It seems like a timeout awaiting the task of the js interop, which I'm not sure why it ocurred so soon (~1min). Nonetheless, the uploads kept going and the progress bar was advancing until it reached 100% and all the uploads were completed. As the task was cancelled previously, no success value is received and I ended up with a successfully completed upload (I checked the folder) but the module error message shown before:

{7ED0E93C-BD20-4ADF-A610-6D2C6DE7DD2A}

- With the new system probably the server won't overload because you are essentially sending the requests sequentially, without any parallelism. You can see the waterfall here:

{59ECC333-4D13-4850-9018-06A12005FD98}

I believe this will result in slower uploads overall, even in fast upload speed environments, as you need to wait for the server response before sending a new one. In this way, you have some idle time for both the server and the client while waiting for their respective response/request each roundtrip.

It took ~400s to complete the ~41MB file upload (throttled to Fast 4G)

For comparison, this is the waterfall with the approach from the PR, using the previous process but limiting the concurrent requests to 5 (taking advantage of some parallelism)

{1C2830AE-C64B-4593-B75D-01BE2EE383E8}

The same file upload, with the same chunk size of 1MB and Fast 4G throttling, took ~260s, so it's quite faster (although the UX was worse because of the network calculations that inmediately throw an error and the progress bar disappear). This time difference would grow exponentially with the file size.

Perhaps it would be a good idea combining your improved upload method with parallel requests and a concurrency lock.

Let me know your thoughts about it.

mdmontesinos:
I also did a comparison without any throttling in localhost.

**Merged version in the PR with parallelism and concurrency lock:**

{9AC613F7-E175-4D38-B4A7-EA83BE8F903C}

Time to upload: ~300-400ms

**Your version in the dev branch:**

{0F930564-A74C-40E1-9F33-0944882A6605}

Time to upload: ~650-800ms

Obviously the tests are not truly accurate and comprehensive, but an orientation to see that the upload times decrease around 50%.

sbwalker:
I am fine with adding throttling support - however I would want the default configuration to be 1 concurrent upload (sequential) as I think the predictability and smooth progress animation is more important than faster upload time - especially since the majority of files which users upload are small files - not large files ie. the system needs to perform well if a user selects 40 files to upload which are 1 MB in size each.

The error message "A task was canceled" needs to be investigated further. If its coming from the JS Interop call it will be difficult to diagnose, as it will depend on the render mode/interactivity. If you are using Static rendering with Server (SignalR) Interactivity then if the SignalR connection drops temporarily, it will result in a "task was cancelled" error. However the SignalR disconnect might also be caused by the debugger in Visual Studio.

mdmontesinos:
> I would want the default configuration to be 1 concurrent upload (sequential) as I think the predictability and smooth progress animation is more important than faster upload time

You are totally right. What approach should I take to implement the configuration? Add a parameter to the file manager component like the chunk size?

> One other item I want to test is if an upload fails for a chunk within a file, does it short circuit the upload process or does it continue to proceed with uploading the remaining chunks

I believe the current parallel implementation does not short circuit, although it should be simple enought to fix

> The error message "A task was canceled" needs to be investigated further

Are you able to reproduce the error? I'm using static render mode with server side interactivity

sbwalker:
In regards to short circuiting, I had to modify the server method in #5065 to return actual error codes (rather than NoContent). Once this was done, if an error occurs on the server it immediately short circuits the file upload process. I tested this by adding some temp code to the UploadFile method to throw an error on the first part received:

```
            if (partCount == 1)
            {
                return StatusCode((int)HttpStatusCode.Forbidden);
            }
```

sbwalker:
@mdmontesinos in regards to "Currently it has a "limitation" because the concurrency lock applies to each file (i.e. if you upload 2 files, you will have 10 simultaneous requests)." - are you saying that if I uploaded 40 small files which are 1 MB in size each, I would get 40 10 simultaneous requests?

mdmontesinos:
@sbwalker Yes, that's exactly what would happen, I'm fixing it right now. I wanted to know first if you were ok with the approach before going deeper into the implementation.

sbwalker:
I would personally be fine if the file upload did not support throttling as it would keep the implementation simpler and easier to maintain in the long term. The ability to modify the chunksize would provide a performance boost for environments which can handle larger payloads.

mdmontesinos:
> By the way; I believe the same happens with the current implementation in the dev branch. The requests in each file are processed sequentially (one at a time), but the files itself are processed in parallel. Therefore, with 40 1MB files you would get 40 simultaneous requests.

@sbwalker I just verified that this happens, so even if we don't implement throttling, we should modify the current behaviour so that files are also processed sequentially and not in parallel to avoid the issue described in #5058.

sbwalker:
@mdmontesinos this looks good. The interesting thing I noticed is that when I set maxConcurrentUploads to 1 it actually seems to perform better than when maxConcurrentUploads = 5. I would want the default to be set to 1 (for the reasons I mentioned above)... but it is interesting that parallel processing does not seem to provide much benefit in my local environment.

mdmontesinos:
@sbwalker I tested again and I can also confirm that the performance is now worse with 5 concurrent instead of 1. But the reason is not the uploads (as they are faster with concurrency), it's the server response. Look at this waterfall:

{70890C0E-ED95-489A-90BD-6D46612FEB0B}

With concurrent requests, it's often the case that the last chunk is received before the previous one (because it might be smaller). So the chunk number 39 (out of 40) is the last one that the server receives and someweird happens that makes the response take 15seconds!! even in localhost. IIRC, this didn't happen before your last commit with the server validation

sbwalker:
I am fine to ignore concurrency at this point

mdmontesinos:
It's interesing that this behaviour does not occur when the network is throttled with Fast 4G.

And it also didn't happened when I tested in https://github.com/oqtane/oqtane.framework/pull/5059#issuecomment-2639591286

sbwalker:
I think the problem is that on the server, it needs to wait until all files are unlocked before it can combine them. With parallel processing, there would be multiple files which are locked when it tries to combine them. Let's

mdmontesinos:
But this was actually how it was working before, when I reported the issue, with many parallel requests being sent at once. Why is it failing now?

sbwalker:
Concurrency issues are very difficult to understand as they are almost impossible to reproduce in a way where you can collect insights about what is happening at run-time. I think we should use sequential processing here so that we do not have to deal with the unpredictability.

mdmontesinos:
@sbwalker Right, I understand. For now, concurrency can be skipped, but the two last commits of the PR should be included:

- A fix to the log level in the file upload error
- Process files sequentially to avoid each of them sending a request simultaneously, potentially leading to server overload

With your changes and these ones, #5058 should be fixed

When the Oqtane new update is available, I'll do some more testing to ensure the upload speed is enough with sequential upload in production environments and try to get concurrency eventually available.

sbwalker:
LogLevel is now included. It will take some time to sort through the various iterations of interop.js to determine which specific changes are necessary to ensure sequential processing.

mdmontesinos:
I believe you can use your implementation and simply change the Promise.All in the end to the for loop of my last commit. That should process the files sequentially.

mdmontesinos:
I can submit another PR if you prefer.

mdmontesinos:
@sbwalker I just submitted the PR #5069. Now files and the chunks in them are all processed sequentially

sbwalker:
Closing this PR, as I moved the File Settings to File Management in #5071

PR #5083, opened 2025-02-10
Fix #5054: parse string to number with invariant culture.

### LINKED ISSUE #5054: Can't Add Page without parent while Using Persian (Iran) Language
Oqtane Info

Version -6.0.2.0
Render Mode - Static 
Interactivity - Server
Database - SQL Server

application returning error whenever in language section i select Iran(Fa)  and try to add new page Parent:<site Root> 

Error Loading Page The input string '-1' was not in a correct format.

exception generated by the system

System.FormatException: The input string '-1' was not in a correct format.
   at System.Number.ThrowFormatException[TChar](ReadOnlySpan`1 value)
   at System.Int32.Parse(String s)
   at Oqtane.Modules.Admin.Pages.Add.<OnInitializedAsync>b__34_0(Page item) in D:\Project\oqtane.framework-dev\oqtane.framework-dev\Oqtane.Client\Modules\Admin\Pages\Add.razor:line 276
   at System.Linq.Enumerable.ListWhereIterator`1.MoveNext()
   at Oqtane.Modules.Admin.Pages.Add.OnInitializedAsync() in D:\Project\oqtane.framework-dev\oqtane.framework-dev\Oqtane.Client\Modules\Admin\Pages\Add.razor:line 276

{"Error":"The input string \u0027-1\u0027 was not in a correct format."}

### Steps To Reproduce
1- Adding Persian (Iran) for localization and enable it
2- Add new Page 
3- Select <site Root > as Parent Page

Image

zyhfish:


mdmontesinos:
This PR is dangerous as it modifies the expected behaviour of all the integer parsing.

You are replacing int.Parse with int.TryParse and setting a default value of 0 when the string can't be parsed. The current behaviour is that a FormatException would be thrown. Now a value of 0 would be used instead. I believe this is extremely dangerous.

sbwalker:
@zyhfish a few comments...

Were you able to do a full regression test to ensure ALL of these changes in this PR do not affect any other functionality? I am hesitant to merge this PR into the 6.1.0 release as the changes are very widespread. We may need a more thorough regression testing strategy before this is included in a release?

Parsing integers was only one scenario identified recently which is problematic in RTL languages. However, is this the only scenario which is problematic, or are there also other data type conversions in Oqtane which do not support RTL languages? I guess what I am asking is similar to above ie. dp we need a more thorough regression testing strategy for RTL languages?

I am not sure I understand the purpose of introducing an abstraction of SharedConverter.ParseInteger()? ie. what benefit does it provide over simply using the native .NET ParseInteger(value, CultureInfo.InvariantCulture)? The native method is more explicit and is a clear signal to someone reading the code that it was written to support cultural differences. Anyone else reading this read would wonder why a "SharedConverter" class is required in order to do something as simple as int.Parse(). Is the abstraction intended to contain additional methods in the future? 

It appears that this PR targeted int.Parse() scenarios... however there are also references to Int32.Parse() in the framework which suffer from the same issue. For consistency, we should probably use int.Parse() throughout the framework. This PR would need to be updated to included these scenarios.

leigh-pointer:
>  I am hesitant to merge this PR into the 6.1.0 release as the changes are very widespread. We may need a more thorough regression testing strategy before this is included in a release?

I agree, this needs a thorough regression testing.

zyhfish:
Hi @sbwalker ,
yeah, I also think we should do a full test on this change but sorry that I don't have enough time to do.

the purpose of SharedConverter class is I think we will need to change more areas around parse side for RTL languages, like boolean, long, etc. we can move all the parse functions in to this class so that it will be easier to maintenance.

zyhfish:
closed this PR as when  I'm trying to verify the changes, I found more issues other than string parsing when language is Iran. for example as below:
1. go to Admin > Schedule Jobs, try to edit a job.
2. exception will thrown.

we need to test with Iran language more to find out a better solution.

PR #5094, opened 2025-02-12
Fix #5054: resolve the issue in fa-IR language.

### LINKED ISSUE #5054: Can't Add Page without parent while Using Persian (Iran) Language

zyhfish:


sbwalker:
@zyhfish this looks interesting... it appears that when the UI Culture is set, it overrides various methods for type conversions, etc... so that standard logic such as int.Parse() behaves in a predictable manner - correct? However does this also mean that the framework would need to include specific culture classes for every RTL language (ie. not just fa-IR)? I am trying to understand the bigger picture so that the can be compared to simply using Culture Neutral conversions in code for various types.

zyhfish:
Hi @sbwalker , I think this should be better than use the overload method of int.Parse, and we may also have other parsing method references like long, double, etc. I will try to verify other RTL languages later. thx

zyhfish:
Hi @sbwalker , I checked and looks all RTL languages will have the issue. updated PR to handle all of them.

PR #5114, opened 2025-02-22
Fix #4936: add the cookie consent theme control.

### LINKED ISSUE #4936: [ENH] GDPR Compliance
### Oqtane Info

Version - 6.0.0
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the enhancement

Oqtane should have a mechanism to make a site GDPR compliant by showing a cookie consent banner and blocking cookies/scripts until the user gives the consent.

To better track the enhancement, I'm opening this issue to replace the discussion in #4791, where the current situation and needs are defined.

The main issues currently were described at https://github.com/oqtane/oqtane.framework/issues/4812#issuecomment-2553441729, and a possible work around in https://github.com/oqtane/oqtane.framework/issues/4812#issuecomment-2554152360.

zyhfish:


sbwalker:
@zyhfish functionally this PR accomplishes the goal for obtaining cookie consent. However there are some modifications I would suggest:

1. The CookieConsent component should have a parameter for Enabled=True/False so that it could be toggled in a theme
2. Rather than including the default markup for the CookieConsent component in the RESX file, I would prefer to have the default markup in the razor component... and only the localizable text should be in the RESX file. This makes it easier for translators to manage.
3. For the Oqtane Theme there is a ThemeSettings.razor component which allows administrators the ability to dynamically configure the theme for their site. It would be useful to have an option in the ThemeSettings which allows the administrator to enable/disable Cookie Consent for their site (default should be disabled).
4. The CookieConsent should have a parameter for ShowPrivacy=True/False. When it is enabled it should display a Privacy Policy button(secondary) which links to a /privacy page by convention. The default site template could include this hidden page by default when creating a site.

PR #5119, opened 2025-02-25
update the cookie consent control.

zyhfish:


sbwalker:
@zyhfish one question I have is related to the configuration of the cookie policy options. It appears that they are set at the installation - not at an individual site level. And it appears that cookie consent is required for the entire installation. So if I have a site that does not require cookie consent (ie. an intranet) and I do not include the CookieConsent component in my theme, then how will the cookies get set properly for the site?

```
            //add cookie consent policy
            services.Configure<CookiePolicyOptions>(options =>
            {
                options.CheckConsentNeeded = context => true;
                options.ConsentCookieValue = Constants.CookieConsentCookieValue;
            });
```

zyhfish:
Hi @sbwalker , I didn't check the internal implement of this flag, but seems it's only a flag to indicate whether the consent is applied but no limit to use cookie without grant the consent. I test with login without apply the consent, it still works.

sbwalker:
@zyhfish so if the CookieConsent component is active for a site and a user chooses to NOT click it (ie. they do NOT consent to allowing cookies), then the system still creates cookies for them?

sbwalker:
@zyhfish after doing a bit more research, it is fine to store "essential" cookies even if the user does not choose to consent to cookies. Essential cookies would be login, etc... However an app is supposed to disable "non-essential" cookies (ie. tracking cookies) if a user does not consent. In Oqtane's case I believe this means the visitor cookie logic should be disabled if cookie consent is enabled for a site and a user has not consented to it. However the only way for Oqtane to know if cookie consent is enabled for a site is to add another option. I woul suggest adding it to the Visitor Management - Settings UI.

zyhfish:
I think the cookie consent is just a flag, and need to be checked when creating a essential cookie. for example, we should only create visitor cookie when CookieConsentService.CanTrack is true, so if site admin want to enable visitor tracking, he should eable the cookie consent bar in theme settings.

sbwalker:
@zyhfish my point is that CookieConsentService.CanTrack will never be true if a site is not using CookieConsent. There is a scenario where a site will choose NOT to include the CookieConsent component in their theme - but will still want to use Visitor Tracking.

zyhfish:
Hi @sbwalker , yes I'm also thinking about this scenario, I think we can add a setting in visitor management named "Ignore Cookie Consent", if this setting is true, we will always save visitor cookie without checking the cookie consent status, otherwise we will only save visitor cookie when cookie consent is granted, does this make sense?

sbwalker:
@zyhfish I realize that it is hard to hit a moving target, so I have tried to outline some high level requirements for this feature below... please let me know if anything does not make sense:

- an Administrator should have the ability to specify Cookie Consent for a Site in an Oqtane installation:
  - Disabled (default)
  - Opt-In
  - Opt-Out
The help text should indicate that Cookie Consent must be used in conjunction with a Theme which supports Cookie Consent (ie. contains the CookieConsent component). This option should be located in Visitor Management - Settings (as we are trying to avoid placing too many random options in the Site Settings UI).

- if Cookie Consent is Disabled for a Site then a CookieConsent component will never be displayed in a Theme and cookies should always be created

- if Cookie Consent is Opt-In or Opt-Out for a Site:
  - if Theme contains a CookieConsent component:
    - if a visitor has never specified their cookie consent preference, the CookieConsent component UI will be displayed as a banner across the bottom of the screen which allows them to specify their preference for non-essential cookies. The message should be "I agree to using cookies to provide the best user experience for this site" and there should be a checkbox to the left where the default value is set based on the Opt-In/Out-Out site setting. The primary button on the right should be "Confirm". A secondary "Privacy" button should link to a /privacy page.
  - if a visitor has specified their cookie consent preference previously, the CookieConsent component should display an icon at the lower left of the screen which allows a visitor to display the full UI and change their preference. Note that changing the preference from enabled to disabled will NOT remove any cookies which have already been created.
  - if Theme does not contain a  CookieConsent component there will be no way for visitors to specify their preference (which means it will fallback to the Opt-In/Opt-Out setting)

- system could include a property in PageState (set in app.razor) for AllowCookies. It would be set based on the combination of the Site setting and Cookie Consent preference for the current visitor. This would allow themes/modules to reference PageState.AllowCookies to determine if non-essential cookies should be created or not.
- app.razor should use this to determine if a Visitor cookie should be created

zyhfish:
thx so much @sbwalker ! I  just have concern to add this setting into visitor management. if we have another module need consent in the future, such as google tracking, it will need to enable consent but this option is under visitor management, it will confuse the user. I still think this need to be added in site settings.

also I think we don't need Opt-Out option for it, consent should be an opt-in option.

sbwalker:
I included opt-out as a requirement because in my research, I found that some implementations use an opt-in approach and others use opt-out - so I would prefer if Oqtane supported both approaches.

zyhfish:
in this case we will not able to use the .net core built-in ITrackingConsentFeature functions, as this is only works for opt-in option.

PR #5136, opened 2025-03-03
modify cookie consent text

sbwalker:


sbwalker:
This PR also attempts to support the display in mobile (ie. by stacking buttons):

image

PR #5253, opened 2025-04-21
Files server optimization

mdmontesinos:
Implements some optimizations for the /files server:

- Early return when 404 or 304, before evaluating query string
- Improves ETag compute by always including the ModifiedOn property and using MD5 to reduce probability of collisions (important when using CDNs)

### DIFF
--- a/Oqtane.Server/Pages/Files.cshtml.cs
+++ b/Oqtane.Server/Pages/Files.cshtml.cs
@@ -137,14 +137,14 @@ namespace Oqtane.Pages
             string downloadName = file.Name;
             string filepath = _files.GetFilePath(file);
 
+            var etagInput = $"{file.ModifiedOn.Ticks}:{file.Size}";
+
             if (Request.QueryString.HasValue)
             {
-                etag = Utilities.GenerateSimpleHash(Request.QueryString.Value);
+                etagInput += $":{Request.QueryString.Value}";
             }
-            else
-            {
-                etag = Convert.ToString(file.ModifiedOn.Ticks ^ file.Size, 16);
-            }
+
+            etag = Utilities.GenerateHashMD5(etagInput);
 
             var header = "";
             if (HttpContext.Request.Headers.TryGetValue(HeaderNames.IfNoneMatch, out var ifNoneMatch))
--- a/Oqtane.Server/Pages/Files.cshtml.cs
+++ b/Oqtane.Server/Pages/Files.cshtml.cs
@@ -133,18 +133,9 @@ namespace Oqtane.Pages
                 }
             }
 
-            string etag;
+            string etag = Convert.ToString(file.ModifiedOn.Ticks ^ file.Size, 16);
             string downloadName = file.Name;
             string filepath = _files.GetFilePath(file);
-
-            if (Request.QueryString.HasValue)
-            {
-                etag = Utilities.GenerateSimpleHash16($"{file.ModifiedOn.Ticks}:{file.Size}:{Request.QueryString.Value}");
-            }
-            else
-            {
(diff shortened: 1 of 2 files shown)


sbwalker:
@mdmontesinos computing an MD5 hash is expensive and impacts performance for all file server requests... which is much worse than the rare chance of eTag collections.

mdmontesinos:
@sbwalker You're right, I thought it wasn't that bad but I did some benchmarks and it was 10x slower. 

I replaced it with a longer simple hash with 16 characters.

sbwalker:
@mdmontesinos this PR still impacts performance when compared to the current logic... as it now calculates a simple hash for ALL standard file requests, whereas previously a hash was only computed for file requests related to dynamic image manipulation (ie. thumbnails). The performance impact would need to be evaluated against the "problem" it is intended to resolve. Can you please share some more specific details about the problem (ie. how often is the existing algorithm generating collisions) as I have my doubts that this issue is large enough to warrant the performance impact being proposed (the existing method which uses ticks and file size is the same method used in the standard .NET Core static file middleware).

mdmontesinos:
@sbwalker The hash collisions are not the issue itself, just a proposed improvement. The reason why I compute the hash for every file is mainly consistency between image manipulation and regular requests. When an image manipulation request is made, the hash is only computed with the query string, but no real information about the original file. If the original file is modified, the ETag should change as well for manipulated images.

mdmontesinos:
@sbwalker Right, I agree with you. I added another commit to only compute the hash when there's a query string

sbwalker:
@mdmontesinos thank you. Can you please explain the need for the new Hash method. If you review the implementation of the existing hash method it will calculate the hash value using the entire string of the text parameter passed in... and returns an 8 digit hash. An 8 digit hash should be sufficiently unique... unless you have encountered scenarios where it is not (and if so I would like to know about them as the GetSimpleHash() method is used in a number of other similar scenarios in other areas of the framework). I just want to make sure we introduce new methods for a valid reason.

mdmontesinos:
@sbwalker I've reviewed the uses of the current hash function, and it's only used for fingerprints of entities which don't really have that many items: installations, module definitions, themes and sites (perhaps some apps can have something like 50 module definitions?). An 8 digit hash is evidently enough for this use case.

However, files can be in the thousands (I currently have around 9k in a site), so the probability of a collision with such a short hash is much higher. Let me note that I haven't detected a collision myself, but I wanted to be cautious and that's why I introduced the 16 digit version.

sbwalker:
@mdmontesinos please note that the hash function in relation to an eTag has nothing to do with the number of files in a site - it has to do with the number of variations of a specific file. When a request is received for a specific file it computes the eTag for that specific file and compares it to the HTTP header to determine if it has changed. So the chance of collisions is low as it is not likely that you have 1000's of variations of the same file. Therefore I do not see a reason to include the new hash method.

sbwalker:
@mdmontesinos this discussion has been useful... it reminded me that eTags are for a specific resource. So from a client side perspective, if you have the following 2 resources:

file.png?param=a
file.png?param=b

the browser treats them as independent resources and maintains the eTag for each of them. So on the server-side there is no need to consider the querystring parameters at all when creating the eTag - the only attributes which matter are those which indicate if the underlying resource has changed. This means that the Files Server can utilize the simple algorithm using Modified.Ticks and Size to calculate the eTag value for any request - whether it is for a standard file or a dynamic image.... there is no need for the hashing in this scenario.

sbwalker:
@mdmontesinos correct - and the call to GenerateSimpleHash can be removed so that the eTag uses the simple calculation in all scenarios

PR #5270, opened 2025-04-29
GetUrlParameters crash

leigh-pointer:
The _urlparametersstate variable is not initialized so in GetUrlParameters it causes a crash

### DIFF
--- a/Oqtane.Client/Modules/ModuleBase.cs
+++ b/Oqtane.Client/Modules/ModuleBase.cs
@@ -17,7 +17,7 @@ namespace Oqtane.Modules
     public abstract class ModuleBase : ComponentBase, IModuleControl
     {
         private Logger _logger;
-        private string _urlparametersstate;
+        private string _urlparametersstate = string.Empty;
         private Dictionary<string, string> _urlparameters;
         private bool _scriptsloaded = false;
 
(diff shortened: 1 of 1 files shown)


sbwalker:
@leigh-pointer can you please explain how to reproduce this problem. If _urlparametersstate is not explicitly set it means that it has a value of null... and since the condition checks for null it should work fine? Is there a specific Url format which causes a problem?

leigh-pointer:
@sbwalker yes i called GetUrlParameters because its public thinking it would return the parameters, but when you have no parameters you get a crash.

sbwalker:
It is a public property... so it is intended to be referenced from module components:

```
            if (UrlParameters.ContainsKey("level"))
 ```

how did you call it?

**Comment by leigh-pointer** [comment]
> It is a public property... so it is intended to be referenced from module components:
> 
> ```
>             if (UrlParameters.ContainsKey("level"))
> ```
> 
> how did you call it?

```
    protected override async Task OnParametersSetAsync()
    {
        await base.OnParametersSetAsync();

       Dictionary<string,string> modParams = base.GetUrlParameters();
```

Because the name of the method dictates Get Url Parameters

```
System.NullReferenceException: 'Object reference not set to an instance of an object.'
This exception was originally thrown at this call stack:
    Oqtane.Modules.ModuleBase.GetUrlParameters(string) in ModuleBase.cs

```

PR #5276, opened 2025-05-02
Updated to Bootstrap 5

leigh-pointer:
Updated to Bootstrap 5.3.5
Update bootswatch Cyborg to 5.3.5 using https://cdn.jsdelivr.net because it is not available at https://cdnjs.com/libraries


thabaum:
https://github.com/oqtane/oqtane.framework/blob/dev/Oqtane.Server/wwwroot/Themes/Templates/External/Client/ThemeInfo.cs

This file should be updated as well, correct?

leigh-pointer:
Yes but I was going to wait on this pr being accepted first, but guess it should be update as well! I will action it.

________________________________

leigh-pointer:
@sbwalker I waould say different 

image

plus Bootswatch also resides there.

image

PR #5277, opened 2025-05-02
Modify the module's server-side resource path.

ohba-ikuo:
I will modify the server-side resource path as follows.

before:
[Owner].Module.[Module].Server.Oqtane

after:
[Owner].Module.[Module].Server


sbwalker:
@ohba-ikuo I believe the changes to Oqtane.Server.csproj were included by mistake?

sbwalker:
@leigh-pointer does this PR appear to be valid? The Oqtane Framework specifies the following in its Oqtane.Server\AssemblyInfo.cs:

```
using Microsoft.Extensions.Localization;

[assembly: RootNamespace("Oqtane")]
```

This PR is intended to provide localization support for the Server assembly of a custom module... but I am not sure if the RootNamespace value is correct.

PR #5306, opened 2025-05-14
fix #5200 - sort folders alphabetically, display folders hierarchically

sbwalker:


ProjectMarco:
Thank you so much! We will update to the new release!

ProjectMarco:
Hi Shaun, we have updated Oqtane to the latest version and indeed the page hierarchy is now correct:

immagine

I want to point out, however, that - even BEFORE the upgrade to 6.1.3 - both the Host and the end user see dozens of folders (which I don't understand what they are):

immagine

sbwalker:
@ProjectMarco I have no idea why you have so many folders - I have never seen that in any Oqtane installation. Are you using any third party / custom modules?

ProjectMarco:
Hi Shaun, we are using first of all modules from the Oqtane ecosystem taken from the Marketplace (open source and paid).
But, above all, our modules developed specifically for which my colleague with the nickname "pnicosia80" is talking to you on GitHub.

sbwalker:
@ProjectMarco I believe those folders are being created by third party/custom modules - not by the Oqtane framework. You will need to investigate.

Issue #5332, opened 2025-05-25
[BUG] ModuleBase ReplaceTokens


leigh-pointer:
### Oqtane Info

Version - 6.1.2
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the bug
The code as written only finds and replaces the first token in the string. This happens because the logic for finding tokens is not inside a loop that continues searching through the entire string

### Expected Behavior
To replace all the tokens in the string

### Steps To Reproduce
Add this string to an HTMLText module with DynamicTokens enabled
Hello [ModuleState:Title], user: [PageState:User:Email]
Only the first token in replaced.

sbwalker:
@leigh-pointer can you please explain how to reproduce this issue? The logic evaluates the content and identifies all tokens. And then it loops through the tokens and replaces them with values.

sbwalker:
@leigh-pointer yes please go ahead

PR #5333, opened 2025-05-25
Fix for ModuleBase ReplaceTokens #5332

leigh-pointer:
Replaced the ReplaceTokens logic to replace all tokens in the string


sbwalker:
I will need to investigate this further. This method was purposely implemented without Regex for performance reasons. Regex is notorious for being very slow and because the HtmlText module is so widely used, it is important that the run-time performance is not impacted. The current logic looks for tokens and exits immediately if none are identified.

PR #5353, opened 2025-06-06
fix #5352: remove requests to cookie consent service when not enabled

### LINKED ISSUE #5352: [ENH] Remove CookieConsent api calls if not enabled
### Oqtane Info

Version - 6.1.3
Render Mode - Interactive
Interactivity - Client
Database - SQL Server

### Describe the enhancement

When Cookie Consent is not enabled, there shouldn't be any requests to the CookieConsent service.

Image

mdmontesinos:
Fixes #5352 with an early return in the CookieConsent control initialization method if cookie consent is not enabled.

### DIFF
--- a/Oqtane.Client/Themes/Controls/Theme/CookieConsent.razor
+++ b/Oqtane.Client/Themes/Controls/Theme/CookieConsent.razor
@@ -103,6 +103,9 @@
     {
         var cookieConsentSetting = SettingService.GetSetting(PageState.Site.Settings, "CookieConsent", string.Empty);
         _enabled = !string.IsNullOrEmpty(cookieConsentSetting);
+
+        if (!_enabled) return;
+
         _optout = cookieConsentSetting == "optout";
         _actioned = await CookieConsentService.IsActionedAsync();
 


sbwalker:
@mdmontesinos this looks good however I did notice a few items in my code review (which are probably questions for @zyhfish as he created the cookie consent feature):

1. CookieConsent.razor has an assignment of PageState.AllowCookies = _canTrack; which is not optimal as PageState is intended to be a read-only cache and should never be mutated outside of SiteRouter as it affects performance of all components which subscribe to PageState. This should be removed and replaced with an alternative approach.

2. What is the purpose of the PageState.AllowCookies property? It appears to be intended as a simple mechanism to determine if cookies are enabled - however it is never utilized. It seems that it could be used in CookieConsent.razor rather than checking var cookieConsentSetting = SettingService.GetSetting(PageState.Site.Settings, "CookieConsent", string.Empty);

Basically we are trying to ensure the cookie consent logic is optimized so that it does not affect run-time performance. In sites where cookie consent is disabled, there should be no API calls related to this feature. If it is enabled (optin or optout) then it is acceptable for API calls to be made to retrieve the visitor's preference (ie. cookie).

zyhfish:
Hi @sbwalker ,
1. update the value of PageState so that other component on the page will update the status without page refresh needed;
2. components can use this property to know whether cookies is allowed.

Issue #5354, opened 2025-06-08
[BUG] Scheduled Jobs UI


leigh-pointer:
### Oqtane Info

Version - 6.1.3
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the bug

```
An Unexpected Error Has Occurred In Oqtane.Modules.Admin.Jobs, Oqtane.Client: Nullable object must have a value.
System.InvalidOperationException: Nullable object must have a value.
   at System.Nullable`1.get_Value()
   at Oqtane.Modules.Admin.Jobs.Edit.SaveJob() in D:\source\_Oqtane Instances\Oqtane.GitHub\Oqtane.Client\Modules\Admin\Jobs\Edit.razor:line 180
   at Microsoft.AspNetCore.Components.ComponentBase.CallStateHasChangedOnAsyncCompletion(Task task)
   at Microsoft.AspNetCore.Components.RenderTree.Renderer.GetErrorHandledTask(Task taskToHandle, ComponentState owningComponentState)
### Expected Behavior
```

### Steps To Reproduce
Navigate to Scheduled Jobs
Edit Purge Job or any job
set Next Execution to a day behind
the click Save

sbwalker:
resolved in #5356 - this is a fairly serious issue as it prevents a host user from managing scheduled jobs in the admin UI

PR #5356, opened 2025-06-09
Fix for Scheduled Jobs UI #5354

leigh-pointer:
This PR addresses an issue where null date/time values could cause exceptions when processing job scheduling. 

**Changes Made**:
- Added proper null checks for _startDate, _startTime, _endDate, _endTime, _nextDate, and _nextTime
- Improved parsing safety for _retentionHistory using int.TryParse()
- Added validation to fail early with meaningful error messages

**Impact**:
Prevents NullReferenceException and InvalidOperationException when date/time fields are missing


sbwalker:
@leigh-pointer wouldn't it make sense to implement the null check within the LocalToUtc method?

Issue #5358, opened 2025-06-10
[BUG]


leigh-pointer:
### Oqtane Info

Version - #.#.#
Render Mode - Static 
Interactivity - Server
Database - SQL Server

ModuleRepository.GetModules does not populate the IDeletable properties and leads to false infomation

### Describe the bug
```
        #region IDeletable Properties

        [NotMapped]
        public string DeletedBy { get; set; }
        [NotMapped]
        public DateTime? DeletedOn { get; set; }
        [NotMapped]
        public bool IsDeleted { get; set; }
```
running this returns all modules regardless.

```
            var allModules = moduleRepository.GetModules(site.SiteId);
            return allModules.Where(m => m.ModuleDefinitionName == DiscussionSettings.MODULE_DEFINITION_NAME && !m.IsDeleted)
                .ToList();
```

### Expected Behavior
In my situation only modules that have not been deleted should be returned!

sbwalker:
You should use: ```IEnumerable<PageModule> GetPageModules(int siteId)``` if you want to access the PageModule domain properties

PR #5359, opened 2025-06-10
Fix GetModules not populating the IDeletable propertie #5358

leigh-pointer:
This PR updates GetModules() to populate deletable properties (DeletedBy, DeletedOn, IsDeleted) from the related PageModule table using an efficient separate query approach.


sbwalker:
You should use: ```IEnumerable<PageModule> GetPageModules(int siteId)``` if you want to access the PageModule domain properties

Issue #5372, opened 2025-07-03
[ENH] Support OAuth authentication for SMTP configuration


mdmontesinos:
### Oqtane Info

Version - 6.1.3
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the enhancement

It seems like Office365 (Microsoft) is deprecating Basic Authentication (username/password) for client SMTP so the current available configuration in Oqtane is insufficient, resulting in errors like:

```
The SMTP server requires a secure connection or the client was not authenticated. The server response was: 5.7.57 Client not authenticated to send mail. Error: 535 5.7.139 Authentication unsuccessful, user is locked by your organization's security defaults policy. Contact your administrator.
```

(SMTP AUTH was already enabled for the account)

The options are disabling security configuration in the Office organization, which is totally not recommended, using paid services like Azure Communication Services Email or using an OAuth flow.

https://learn.microsoft.com/en-us/exchange/mail-flow-best-practices/how-to-set-up-a-multifunction-device-or-application-to-send-email-using-microsoft-365-or-office-365#client-smtp-submission-authenticate-mail-sent-from-your-device-or-application-through-a-cloud-mailbox

Therefore, it would be necessary that Oqtane implements the OAuth flow required by Microsoft.

https://learn.microsoft.com/en-us/exchange/client-developer/legacy-protocols/how-to-authenticate-an-imap-pop-smtp-application-by-using-oauth

Perhaps there's some other way to properly configure the email service that I'm not aware of. Any help is appreciated.

sbwalker:
This article provides more context:

https://isoc.net/microsoft-is-ending-smtp-basic-authentication-what-it-means/

Personally I use SendGrid for SMTP as it provides much better capabilities (and supports Basic Authentication). However, it does make sense that Oqtane will need to be enhanced to support OAuth for SMTP.

mdmontesinos:
I can't actually use another provider as I need send the emails from the Office 365 organization email account of my client. Hence why I require the OAuth support.

sbwalker:
Yes, that makes sense. I believe that in order to support OAuth, Oqtane will need to utilize MailKit as System.Net.Mail does not support OAuth. Integrating MailKit was discussed previously and does make sense as Microsoft is no longer actively maintaining System.Net.Mail. This should not affect existing installations as Oqtane uses the Notification abstraction layer for sending email (for this exact reason).

mdmontesinos:
[...]

PR #5376, opened 2025-07-04
feat: replace System.Net.Mail with MailKit (#5372)

mdmontesinos:
Replaces System.Net.Mail with the MailKit library to send emails in the NotificationJob.

The only question I have is related to the UseSSL option. I believe it can be removed from the SiteSettings UI as the connect method should use MailKit.Security.SecureSocketOptions.Auto, although I'm not 100% sure of the implications

### DIFF
--- a/Oqtane.Server/Infrastructure/Jobs/NotificationJob.cs
+++ b/Oqtane.Server/Infrastructure/Jobs/NotificationJob.cs
@@ -57,7 +57,7 @@ namespace Oqtane.Infrastructure
 
                         client.Connect(host: settingRepository.GetSettingValue(settings, "SMTPHost", ""), 
                                        port: int.Parse(settingRepository.GetSettingValue(settings, "SMTPPort", "")),
-                                       options: MailKit.Security.SecureSocketOptions.Auto);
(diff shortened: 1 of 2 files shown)


sbwalker:
@mdmontesinos MailKit has a Connect method that takes a SecureSocketOptions argument instead of a simple bool for SSL.

The options are as follows:

- None: Don't use any form of SSL (or TLS).
- Auto: Automatically decides which type of SSL mode to use based on the specified port. Note: This only works reliably when the port is a standard defined port (e.g. 25, 587 or 465 for SMTP).
- SslOnConnect: This specifies that MailKit should connect via an SSL-wrapped connection.
- StartTls: Use the STARTTLS method for SSL/TLS encryption. If the server doesn't support the STARTTLS command, abort the connection.
- StartTlsWhenAvailable: Use the STARTTLS method for SSL/TLS encryption if the server supports it, otherwise continue using an unencrypted channel.

I believe the current Oqtane options for SSL map as follows:

- false = None
- true = StartTls

mdmontesinos:
@sbwalker I tested and the mapping you suggested seems to work fine.

PR #5387, opened 2025-07-09
feat: handle timezones and conversions with NodaTime

mdmontesinos:
Integrates NodaTime according to what was discussed in #5348.

- NodaTime package was added to Shared project, as it's only directly referenced there
- TimeZone list method added to Utilities file
- Added overloads for all the Utilities time conversion methods using timezoneIds and NodaTime instead of TimeZoneInfo
- Modified ModuleBase time conversions to use new utility methods

For the utility methods, I basically duped the logic from the current ones and adapted it to NodaTime. However, I could modify the current ones to convert .NET TimeZoneInfo to a NodaTime DateTimeZone and integrate the logic in a base method to avoid code duplication. Thoughts about this?

Also, the reset of the timezone field on upgrade is required (I leave that for you).


sbwalker:
@mdmontesinos i don't think we need to be concerned about backward compatibility in this case as the timezone support was only introduced in 6.1.3. I also merged the .NET 9.0.7 changes which caused a conflict in this PR.

mdmontesinos:
@sbwalker I just resolved the conflict. About compatibility, I also disregarded it for the same reason you are mentioning. I don't believe it had great adoption yet.

PR #5644, opened 2025-09-20
Package Updates

leigh-pointer:
Updated Client Radzen.Blazo to 7.3.5
Updated Server HtmlAgilityPack 1.12.3


sbwalker:
as if Oqtane 6.2.0, all package updates need to be done in the *.csproj as well as the corresponding *.nuspec files

PR #5669, opened 2025-09-27
build(deps): downgrade to .NET 8 and update packages

mrizwan1974:
Migrate target framework from net9.0 to net8.0 across all projects. Update related NuGet packages from 9.x to 8.x versions for compatibility. Downgrade Swashbuckle.AspNetCore to 6.9.0 and adjust database providers. Add .gitignore entry for Codacy instructions file.


sbwalker:
Closing as we are not moving the framework back to .NET 8.

PR #5688, opened 2025-10-02
exclude package assets in solution.

zyhfish:


sbwalker:
@zyhfish can you please explain why this folder needs to be excluded.... it should only contain static assets from Nuget packages.

zyhfish:
Hi @sbwalker , the solution will built failed if there is files with same name of referenced third package, it will be a rarely case but possible a problem.

PR #5724, opened 2025-10-17
Fix #5714: sort profile by categories.

### LINKED ISSUE #5714: [BUG] Profile fields will not display by category
### Oqtane Info

Version - 6.2.1
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the bug
After I tried to add some custom profile definitions to existing category, such as "Name" or "Contact". the fields will not render together by the category but render separately when edit a user's profile:
Image

### Expected Behavior
The profile fields should be rendered by category

### Steps To Reproduce
1. add custom profile definitions to existing category "Name".
2. edit a user and switch to the "Profile" tab.

zyhfish:


sbwalker:
@zyhfish I resolved this issue in #5728 by changing the sort order of the GetProfilesAsync() method

sbwalker:
@zyhfish I would prefer to keep this simple and not introduce a ProfileCategory entity

PR #5796, opened 2025-11-12
update Swashbuckle to 10.0.0

leigh-pointer:
update Swashbuckle to 10.0.0 in csproj and nuspec


sbwalker:
@leigh-pointer I can no longer build the solution after including this PR - it is giving me a compilation exception. Were you able to build the solution?

sbwalker:
fixed in #5800

PR #5832, opened 2025-11-20
Fix #5649: handle not found request.

### LINKED ISSUE #5649: [ENH] Return 404 status code for non-existent routes
### Oqtane Info

Version - 6.2.1
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the enhancement
Currently, when a URL is not found, the application issues a 302 redirect to the 404 page, and that page then returns a 200 status code. From an SEO perspective, this is not considered a best practice. It would be more appropriate to return a proper 404 status code while rendering the custom 404 page directly, without performing a redirect. Am I correct in this understanding?

zyhfish:


sbwalker:
@zyhfish instead of checking for "/api/" it should use the Constants.ReservedRoutes similar to TenantMiddleware.cs

I am not a fan of the hard-coded file extension logic. Could we simply check if the last segment in the path contains a "." and if so, assume it is a request for a static file? I realize this is not 100% accurate (as it is possible that someone could specify a page path of "x.y") but a hard-coded list of file extensions is not accurate either  as there will be extensions we forget to include or there will be new extensions created in the future which the logic does not support.

zyhfish:
Hi @sbwalker , thanks so much for review, I updated the PR to remove hard coded values.

sbwalker:
@zyhfish I noticed one strange behavior with this PR:

If I am logged in to a site and I enter a url in the browser of "/does-not-exist" it will retain the Url and display the 404 page content, however the other theme components on the page make it appear that I am no longer logged in:

image

However if I click on Login it tells me that I am already logged in (and displays the theme components correctly).

image

PR #5847, opened 2025-12-01
Fix SMTPRelay condition for email sender

W6HBR:
Prior change was leaving sender null and not properly setting "From" address when used in a relay configuration.  This caused emails to go to the deleted state and not be delivered.


W6HBR:
Need to add check for empty sender and only apply default relay sender if no sender was already specified.

PR #5877, opened 2025-12-15
Offers cache stampede protection

MarkCiliaVincenti:
Also minor change of lock to use the new(ish) `System.Threading.Lock`.

### DIFF
--- a/Oqtane.Server/Extensions/CacheExtensions.cs
+++ b/Oqtane.Server/Extensions/CacheExtensions.cs
@@ -1,3 +1,4 @@
+using AsyncKeyedLock;
 using Microsoft.Extensions.Caching.Memory;
 using System;
 using System.Collections.Generic;
@@ -8,13 +9,21 @@ namespace Oqtane.Extensions
 {
     public static class CacheExtensions
     {
+        private static AsyncKeyedLocker<string> _locker = new();
         private static string _cachekeys = "cachekeys";
 
         public static TItem GetOrCreate<TItem>(this IMemoryCache cache, string key, Func<ICacheEntry, TItem> factory, bool track)
         {
-            if (!cache.TryGetValue(key, out object result))
+            if (cache.TryGetValue(key, out TItem result))
             {
-                using ICacheEntry entry = cache.CreateEntry(key);
+                return result;
+            }
+
+            using var _ = _locker.Lock(key);
+
+            if (!cache.TryGetValue(key, out result))
+            {
+                using var entry = cache.CreateEntry(key);
                 result = factory(entry);
                 entry.Value = result;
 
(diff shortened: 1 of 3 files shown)


sbwalker:
@MarkCiliaVincenti thank you for submitting a PR. The Oqtane framework is managed according to a high level philosophy outlined here: https://www.oqtane.org/blog/!/20/oqtane-philosophy. You will notice that one of the items is "Minimizing Dependencies". 

Adding a new package reference to "AsyncKeyedLock" would only be considered if there is a clear and compelling reason to do so. Can you please provide evidence that the existing caching approach is problematic, as this has not been reported previously for any production deployment that I am aware of.  

To provide some additional context, there was previously a discussion about using HybridCache (introduced in .NET 9) as it aligns with the "Microsoft Alignment" item in the philosophy - however has not been implemented because it currently requires a full async approach (https://github.com/dotnet/extensions/issues/5679) which would result in major refactoring and breaking API changes. HybridCache was not being considered for its stampede protection (although it does possess this) but rather for its scale-out (ie. distributed cache) capability.

MarkCiliaVincenti:
Hi @sbwalker, essentially it's about protection from cache stampede. The current implementation offers no such protection, meaning that if a key does not exist and 10 concurrent threads try to GetOrCreate it at the same time, the factory will get called 10 times. One of the motivating reasons why one would use a cache is because the data is expensive to get, so you'd want to keep it in memory. That is why it is very important to offer cache stampede protection. All the good caching libraries offer cache stampede protection out of the box.

In the scenario I made up, with this implementation what would happen would be that the first thread to get the lock for that key would enter and call the function while the other 9 stay waiting. Once it's ready, the other 9 would enter the lock one at a time, recheck the cache, find that it's created already and return it. That function would get called just once.

MarkCiliaVincenti:
I wanted to add that I am the author of AsyncKeyedLock and it's a mature library used even in official Microsoft products such as https://github.com/microsoft/kiota and https://github.com/microsoft/Vipr.

sbwalker:
@MarkCiliaVincenti so you are saying that you have not actually encountered this in production with Oqtane... it is a general best practice. I do agree that stampede protection is a best practice, however since there are no Oqtane installations which are encountering this issue currently, my preference is to tackle this item as part of the larger caching initiative which I alluded to in my previous post (ie. scale-out support).

MarkCiliaVincenti:
An alternative to HybridCache is https://github.com/ZiggyCreatures/FusionCache, a mature library.

sbwalker:
Yes, I was aware of FusionCache. As per the product philosophy I was trying to maintain "Microsoft Alignment" which is why I was hoping HybridCache would be improved - see https://github.com/dotnet/extensions/issues/5679. However with Marc Gravell's departure from Microsoft I do not believe this item will be prioritized. FusionCache may be the best alternative anyways based on:

https://github.com/ZiggyCreatures/FusionCache/blob/main/docs/MicrosoftHybridCache.md

sbwalker:
closing this PR, as FusionCache has now been integrated into the framework for the next release

PR #5917, opened 2025-12-23
Fix: resolve Radzen JS loading issue ('Radzen' was undefined) modules

Bhushan02-dot:


sbwalker:
@Bhushan02-dot thank you for submitting this PR... however Oqtane is a framework which means that it provides extensibility points for integrating third party components (rather than direct integration into core services).

PR #5932, opened 2025-12-28
Fix SMTPRelay condition for fromEmail validation

W6HBR:
Use default from email for non-relay as well as relay configured where from email was not supplied.

### DIFF
--- a/Oqtane.Server/Infrastructure/Jobs/NotificationJob.cs
+++ b/Oqtane.Server/Infrastructure/Jobs/NotificationJob.cs
@@ -186,7 +186,7 @@ namespace Oqtane.Infrastructure
                                     var mailboxAddressValidationError = "";
 
                                     // sender
-                                    if ((settingRepository.GetSettingValue(settings, "SMTPRelay", "False") == "True") && string.IsNullOrEmpty(fromEmail))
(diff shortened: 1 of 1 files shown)


Oqtane.Server/Infrastructure/Jobs/NotificationJob.cs
```diff
@@ -186,7 +186,7 @@ await client.AuthenticateAsync(settingRepository.GetSettingValue(settings, "SMTP
                                     var mailboxAddressValidationError = "";
 
                                     // sender
-                                    if ((settingRepository.GetSettingValue(settings, "SMTPRelay", "False") == "True") && string.IsNullOrEmpty(fromEmail))
+                                    if ((settingRepository.GetSettingValue(settings, "SMTPRelay", "False") != "True") || (settingRepository.GetSettingValue(settings, "SMTPRelay", "False") == "True") && string.IsNullOrEmpty(fromEmail))
```
zyhfish: Looks like the bracket put in wrong place, IMO this can be changed to:
```
if (settingRepository.GetSettingValue(settings, "SMTPRelay", "False") != "True" || string.IsNullOrEmpty(fromEmail))
```

sbwalker:
fixed by #5937

PR #5943, opened 2026-01-05
Fix #5942: set the next execution time  correctly.

### LINKED ISSUE #5942: [BUG] schedule job will be executed twice if change the execution time manually
### Oqtane Info

Version - 10.0.3
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the bug
When I try to change the execution time of a job to let it been executed manually,  the job will be executed twice.

### Expected Behavior
The job should only been executed once and update the next execution time correctly.

### Steps To Reproduce
1. edit "purge job", change the date of "next execution" to some value before today;
2. wait for 2 more minutes, check the log of "purge job", you will notice the job has been executed twice.
Image

zyhfish:


sbwalker:
@zyhfish this PR affects the behavior of scheduled jobs. If we use the PurgeJob as an example....

Currently it is scheduled to run every day at 3:00 AM. It does not matter if the Purge Job takes 5 minutes to run or 1 hour to run... it will always run the following day at 3:00 AM. This is handled by the NextExecutionDate logic which remembers the time it last ran and it adds the interval (+1 day).

With this PR, the time when a job runs will slowly "drift" and will not execute at a precise time. If the job executes at 3 AM and takes 15 minutes, then tomorrow it will run at 3:15 AM. This is not the expected behavior of a "schedule".

I think the real problem is that users should not be allowed to edit/modify a job when it is active or executing... as essentially this is like refueling the plane in mid-air (which I know is technically possible, but also has many risky side effects).

zyhfish:
Hi @sbwalker , I already considered this scenario, so I logged the start time when the job started to be executed, the next execution time will be based on the start time, but not the end time of the job.

PR #5966, opened 2026-01-18
Fix user folder creation and prevent duplicate user roles

vnetonline:
Updated user folder creation to use a unique path per user and ensure the folder is created if it does not exist. Added a check to prevent assigning duplicate auto-assigned roles to users.


sbwalker:
replaced by #5969

PR #5980, opened 2026-01-22
Image optimization to webp without manipulation

### LINKED ISSUE #5979: [ENH] Image optimization to webp without resizing
### Oqtane Info

Version - 10.0.4
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the enhancement

Currently, there's no way to simply optimize images without having to resize them via the files server or api. Even if you don't specify the width or height, you need the "Image sizes" property of the folder to match the current size. Therefore, it's impossible to perform optimizations to images when you don't know/restrict the dimensions when the user uploads them.

I've submitted a PR that adds this functionality to Oqtane.

mdmontesinos:
This pull request introduces new functionality for image optimization, allowing images to be served as optimized WebP files with configurable quality (closes #5979).

**Image Optimization Functionality:**

- Added a new `GetOptimizedImage` endpoint to `FileController` that serves the optimized WebP images.
- Updated the `IImageService` interface to include a new `OptimizeImageToWebp` method and defined a `DefaultQuality` constant for image optimization.
- Implemented the `OptimizeImageToWebp` method in `ImageService`, using the specified quality. The WebP encoder now uses the provided quality value instead of a hardcoded default.

**File Handling and Query Parameter Improvements:**

- Enhanced the file handling logic in `Files.cshtml.cs` to support an `optimize` query parameter, which triggers the image optimization with optional `quality` and `recreate` parameters. Image manipulation (resize, crop, ...) takes precedence over optimization if both queries are included.

(Only the PR summary was generated by Copilot, not the code)

### DIFF
--- a/Oqtane.Server/Controllers/FileController.cs
+++ b/Oqtane.Server/Controllers/FileController.cs
@@ -778,6 +778,71 @@ namespace Oqtane.Controllers
             return System.IO.File.Exists(errorPath) ? PhysicalFile(errorPath, MimeUtilities.GetMimeType(errorPath)) : null;
         }
 
+        [HttpGet("image/{id}/optimize/{quality}/{recreate}")]
+        public IActionResult GetOptimizedImage(int id, int quality, string recreate)
+        {
+            var file = _files.GetFile(id);
+
+            var _ImageFiles = _settingRepository.GetSetting(EntityNames.Site, _alias.SiteId, "ImageFiles")?.SettingValue;
+            _ImageFiles = (string.IsNullOrEmpty(_ImageFiles)) ? Constants.ImageFiles : _ImageFiles;
+
+            if (file != null && file.Folder.SiteId == _alias.SiteId && _userPermissions.IsAuthorized(User, PermissionNames.View, file.Folder.PermissionList))
+            {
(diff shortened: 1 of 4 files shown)


sbwalker:
@mdmontesinos I have not merged the associated PR #5980 because as part of the research for achieving the "scale out" enhancement on the roadmap, it will be necessary to abstract the storage system so that the location of files are determined at runtime. The existing Folder abstraction was designed with this in mind... however I noticed that the IImageService does not support this abstraction as it includes file paths directly in the interface methods. Since the IImageService interface will need to be overhauled, I would prefer to avoid adding additional methods at this point which will further complicate our future plans.

mdmontesinos:
@sbwalker Right, no problem! I will adjust and submit the PR again when the new abstractions are implemented. Is there any estimate for the folder providers functionality?

sbwalker:
@mdmontesinos #6064 contains the initial version of the Folder Provider enhancement. It is very large and requires a detailed review so it may not be included in the 10.1.0 release.

mdmontesinos:
@sbwalker Great, I'll keep an eye out for updates

PR #5996, opened 2026-01-30
Move I[Module]Service to Shared

leigh-pointer:
Relocate the I[Module]Service interface from the client Services file to a new Shared/Interfaces/I[Module]Service.cs so the interface becomes a shared contract. Remove the duplicate interface from the client-side [Module]Service implementation. It was move in error, release 6.1.5 PR add a new Visual Studio Project Template #5493

### DIFF
--- a/Oqtane.Server/wwwroot/Modules/Templates/External/Server/[Owner].Module.[Module].Server.csproj
+++ b/Oqtane.Server/wwwroot/Modules/Templates/External/Server/[Owner].Module.[Module].Server.csproj
@@ -27,7 +27,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\Client\[Owner].Module.[Module].Client.csproj" />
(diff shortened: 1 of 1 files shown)


sbwalker:
@leigh-pointer one thing to note is that there are "external" and "internal" module templates now. The "internal" template is used by the Oqtane Application Template and the original goal was to keep it as consistent with the "external" template as possible to simplify future maintenance (obviously it cant be identical because "internal" templates do not have projects, solutions, packages, etc...). With this PR, the Service interface will be in the Shared project for "external" modules and in the Client project for "internal" modules. In hindsight I believe this is why I moved the Service interface to the Client project - so that they would be consistent. The other approach would be to modify the Oqtane Application Template to be consistent with this PR (by including the interface in the Shared project). This could be tackled as a subsequent PR.

leigh-pointer:
@sbwalker ok i will modify the Application project on another PR

PR #6013, opened 2026-02-06
Rename client and server module services

leigh-pointer:
Rename the client service class to Client[Module]Service and the server service file to Server[Module]Service, updating all references accordingly. Updated DI registration in ClientStartup to register Client[Module]Service, and updated Edit.razor and Index.razor to inject and call Client[Module]Service instead of [Module]Service. This clarifies client vs server implementations and avoids naming collisions.


sbwalker:
@leigh-pointer I am going to review these changes in more detail before merging. The "dual service" approach has existed in its current form in the default module template since Oqtane 5.1 (early 2024). It has been discussed many times previously, and it has been working correctly on all render modes/hosting models. 

The recent change to move the interface to the Shared project seems to have resulted in a bunch of additional changes to the module template, and I am concerned that this may have caused a regression.

image

leigh-pointer:
@sbwalker 
Thanks for taking a closer look at this.
To clarify: this PR is purely about naming and does not change the dual‑service pattern or any behavior. The server service remains Server[Module]Service; the client implementation has been renamed from [Module]Service to Client[Module]Service, and all usages/DI registrations were updated accordingly.
The interface location in the Shared project is unchanged by this PR, and there are no logic, API, or render‑mode changes – just the class/file names to make the client/server distinction explicit and avoid naming collisions.
If you see anything that suggests a behavioral change or regression, that would be unintentional, please point it out and I’ll fix it immediately.

sbwalker:
@leigh-pointer after merging these changes and creating a new module using the default template, the module does not compile:

image

I am going to roll these changes back as it seems my regression concerns were valid.

sbwalker:
@leigh-pointer the problem was that the new solution file (*.slnx) did not specify that Oqtane.Server should not be rebuilt - this has been fixed in https://github.com/oqtane/oqtane.framework/pull/6029. I was able to apply your changes and test again, and the module compiles and functions fine.

PR #6064, opened 2026-02-21
implement the folder provider.

zyhfish:
1. add the folder provider interface and folder provider factory;
2. implement the default folder provider;
3. update the core system to use folder provider instead of  file system directly.

attached a demo project to verify the folder provider works with azure storage.
[Oqtane.FolderProviders.zip](https://github.com/user-attachments/files/25455348/Oqtane.FolderProviders.zip)


sbwalker:
@zyhfish I will merge this into the next release which will focus on "scale out" scenarios

mdmontesinos:
@sbwalker Just checking, is there any estimate on this feature? This would allow using Deployment Slots in Azure to avoid downtime during updates

sbwalker:
@mdmontesinos this will be the major feature included in the 10.3 release. Note that this is limited to content folders - it does not support storing the Oqtane application files in an external location. It is already possible to configure Oqtane using deployment slots in Azure (perhaps @zyhfish can share some info on how he has done this in the past)

zyhfish:
Hi @mdmontesinos ,
as what I analyzed before, azure web app is using storage container to store the website files, zip / unzip process will take a long time when it contains too much files, which cause the upgrade process not able to be completed in azure.
what I did for azure env is to use FTP service to upload the upgrade files instead of upgrade directly in oqtane.

mdmontesinos:
@zyhfish That is a general issue with automatic upgrading in Azure, which rather requires the manual approach to mentioned that I also use, but I don't think it has anything to do with slots. However, the specific issue which prevents using Oqtane with Deployment Slots is the content folders.

As far as I know, when you create a deployment slot (i.e. staging), it does not copy the app content to it, so all the user uploaded images wouldn't be available in that slot and everything would show as broken links. Then, if I swap with the production slot, all the content would actually be moved to the staging one. Additionally, there's no easy way to copy all the content between slots manually, and in my case, it would be around ~50 GB.

Therefore, with the folder providers I could use Azure Storage to detach the content from the app's filesystem so there wouldn't be an issue with deployment slots. As for the Oqtane application files, if you are using the newer application template approach, it would be easier, as all the framework and module updates are handled like a normal deployment. However, with the "old" approach, there might still be some issues and you would be required to manually copy upgrade zips from both the framework and module between slots, but still it would be much less that the content files.

Is this right? Or am I missing something?

zyhfish:
Hi @mdmontesinos , I think the folder provider is useful for you case. the content files can be shared between different slots by using remote storage, you will need to add the folder configs on both slots to sync the contents.

mdmontesinos:
@zyhfish but have you been able to use deployment slots to achieve no-downtime when updating Oqtane, or a module?

zyhfish:
Hi @mdmontesinos , no, I haven't done this before. IMO we need to match below requirements:
1. when swap the slots, the database schema need to be updated, the process time should not be long;
2. the files between prod and staging slots should be same or keep synced. 

I guess #2 is the hard point for your case. after we implement the folder providers and move  your content files into remote storage, I guess this may becomes much easier, as this can reduce the website file's size.

mdmontesinos:
@zyhfish Yes, syncing the app's content is the main issue with using slots, that's why the folder providers solves a huge chunk of it. Although it would still be problematic to sync the framework and module packages, any ideas for this?

I believe that being able to deploy without users experiencing any downtime is an important functionality to have for the framework

zyhfish:
Hi @mdmontesinos ,
I think the best way is to refresh the staging env with latest prod env every time before the upgrade process. this can make sure the upgrade will not be failed when swap slots. (specially to database, if the database hasn't been fresh for a while, the schema maybe different after several upgrade.)

mdmontesinos:
@zyhfish Sure, but Azure does not provide any way to copy between or sync deployment slots, how do you refresh it?

zyhfish:
Hi @mdmontesinos , you can use azure pipeline to call to azure FTP service to download / upload assets from prod to staging env, it can also running commands to export/import database. you can reference https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/ftp-upload-v2?view=azure-pipelines to get more information.

rohzart:
@zyhfish Hello Ben, found an issue in the migration. 

```
The ALTER TABLE statement conflicted with the FOREIGN KEY constraint "FK_Folder_FolderConfig". The conflict occurred in database "****_db", table "dbo.FolderConfig", column 'FolderConfigId'.
```

Just to get through with the install, migration completed by allowing null for `FolderConfigId` in the `Folder` table.
`folderEntityBuilder.AddIntegerColumn("FolderConfigId", true, 0);`

Admin/Files won't load since there's no record in the `FolderConfig` table.

I'm not sure atm whether the application should fall back to the default file provider when there's no provider in the FolderConfig or the default file provider data must be seeded (and if the data is seeded, then the default value might need be 1). I'm trying to figure what to do next.

Btw, it's really nice code. :)

zyhfish:
Hi @rohzart ,
Thanks so much for testing this function, the default folder  provider and existing folder updating code are located in 10_1_0 method in this mean time, because this PR submitted months ago. if you have versions installed newer than this version, the upgrade code will not be executed so that you meet the default folder provider missing issue. I will update the upgrade method path after the which version is decided to include this PR, thx.

here is the related code:
https://github.com/oqtane/oqtane.framework/pull/6064/changes#diff-78f7759b9f27b23d8e6c469aac194935391ed1dc5b757cfeadd461ce7f2362e6R650

rohzart:
Hi @zyhfish Ben, thank you. I was able to make tweaks to run `UpgradeFolderConfigs(...)`. 
The default provider seeds in the folder config table but the folder table records did not get updated with the folder config id.

The `folders.GetFolders(..)` returns null. I couldn't figure it out why.
Even `db.Folder` returns null.
```
public IEnumerable<Folder> GetFolders(int siteId, int userId)
{
    using var db = _dbContextFactory.CreateDbContext();
    var folders_0 = db.Folder.ToList();
...
}
```
I suspect the default null value assigned to the folder config id is posing an issue in EF. 

I'm guessing you figured that out; the next change made in that function:
` FolderConfigId = _folderProviderFactory.GetDefaultConfigId(folder.SiteId),`
introducing `IFolderProviderFactory` in `FolderRepository` caused a cyclic reference.

Runtime error:
```
'Error while validating the service descriptor 'ServiceType: Oqtane.Services.IImageService Lifetime: Scoped ImplementationType: Oqtane.Services.ImageService': A circular dependency was detected for the service of type 'Oqtane.Providers.IFolderProviderFactory'.
Oqtane.Services.IImageService(Oqtane.Services.ImageService) -> Oqtane.Providers.IFolderProviderFactory(Oqtane.Providers.FolderProviderFactory) -> Oqtane.Repository.IFolderConfigRepository(Oqtane.Repository.FolderConfigRepository) -> Oqtane.Repository.IFolderRepository(Oqtane.Repository.FolderRepository) -> Oqtane.Providers.IFolderProviderFactory'
```

I'm wondering if using `migrationBuilder.Sql(...)` during migration itself to seed data and set required defaults would be a good idea.

PS: on my local branch, I have the latest of your branch as well as the latest from upstream/dev.

rohzart:
@zyhfish Hello Ben,

I got it working like this (below); tested over SQL Server only. Please review:

```c#
protected override void Up(MigrationBuilder migrationBuilder)
{
    var folderProviderEntityBuilder = new FolderConfigEntityBuilder(migrationBuilder, ActiveDatabase);
    folderProviderEntityBuilder.Create();

    migrationBuilder.Sql($@"
        INSERT INTO FolderConfig
        (SiteId, Name, Provider, CreatedBy, CreatedOn, ModifiedBy, ModifiedOn)
        VALUES (1, '{Constants.DefaultFolderProvider}', '{Constants.DefaultFolderProvider}', '', CURRENT_TIMESTAMP, '', CURRENT_TIMESTAMP)
    ");

    var folderEntityBuilder = new FolderEntityBuilder(migrationBuilder, ActiveDatabase);
    folderEntityBuilder.AddIntegerColumn("FolderConfigId", false, 1);
    folderEntityBuilder.AddStringColumn("MappedPath", 512, true);
    folderEntityBuilder.AddForeignKey("FK_Folder_FolderConfig", "FolderConfigId", "FolderConfig", "FolderConfigId", ReferentialAction.NoAction);
}
```

If this works out, the `UpgradeFolderConfigs(...)` won't be needed but the looping through all the folders makes me think there are requirements I'm unaware of.

Btw, the other issue: cyclic reference error persists.

sbwalker:
@rohzart thank you for exploring. Please note the PR which @zyhfish created is still a draft and could be refactored significantly when it is merged. This could include refactorings of the data structures, interfaces, etc... I am just mentioning this so that it is clear that this is a work in progress and should not be considered a final result (ie. do not upgrade any production sites to use this PR or you may end up in a situation where you cannot upgrade)

rohzart:
@sbwalker That's so kind of you! Thanks for watching out! :) 
... I definitely won't go to production with anything unstable/unreleased and I am rethinking/scrutinizing my approach.

zyhfish:
Hi @rohzart ,
thanks so much for your contribution, I verified the upgrade scenario and it doesn't work. we are not recommend to use sql query directly because we need to consider the compatibility between several database providers but not only for SQL Server. I'm thinking on how to handle this scenario.

Thanks,

rohzart:
@zyhfish Hey Ben, you worked it out! Cheers! :D

sbwalker:
@zyhfish I merged the PR in its current form. I will review the implementation and we can work through any refactoring.

Issue #6086, opened 2026-02-26
[BUG] Pager control should respond to CurrentPage changes


markdav-is:
### Oqtane Info

Version - 10.0.1
Render Mode - interactive 
Interactivity - Server
Database - SQL Server

### Describe the bug
The `Pager` component initializes its internal page state from the `CurrentPage` parameter on first render but does not update when `CurrentPage` changes on subsequent renders. This means a parent component cannot programmatically reset the pager to page 1 (e.g., after clearing filters) by updating a bound value.

### Expected Behavior
The pager resets to page 1 and the first page of items is shown.

### Steps To Reproduce
1. Add a `Pager` component bound to a filtered list with `CurrentPage` tied to a parent-managed page variable:

    ```razor
    <Pager Items="@_filteredItems" PageSize="20" CurrentPage="@_page.ToString()" OnPageChange="OnPageChange">
    ```

2. Navigate to any page beyond page 1 using the pager controls.
3. From the parent, reset the page variable to `1` and call `StateHasChanged()`:

    ```csharp
    _page = 1;
    StateHasChanged();
    ```

4. Observe the pager UI.

## Workaround
Force the `Pager` to remount by using a `@key` directive tied to an integer that is incremented whenever a programmatic page reset is needed:

```razor
<Pager @key="_pagerKey" Items="@_filteredItems" PageSize="20" CurrentPage="@_page.ToString()" OnPageChange="OnPageChange">
```

```csharp
private int _pagerKey = 0;

private void ClearFilters()
{
    // ... reset filter state ...
    _page = 1;
    _pagerKey++;   // forces Pager to remount at page 1
}
```

Incrementing `_pagerKey` causes Blazor to tear down and remount the `Pager` component, which reinitializes its internal state from `CurrentPage`.

## Impact
Any page that uses `Pager` with dynamic filtering or any programmatic page reset will silently retain stale page state, leading to empty or confusing result displays without any user-visible error.

sbwalker:
@markdav-is I am unable to reproduce this behavior using the criteria you specified above ie. Interactive render mode

```
    <button type="button" class="btn btn-primary" @onclick="SetPage">Set Page</button>
    <Pager Items="@_items" CurrentPage="@_page.ToString()" OnPageChange="OnPageChange">
        ...
    </Pager>

    private void OnPageChange(int page)
    {
        _page = page;	
    }

    private async Task SetPage()
    {
        _page = 2;
    }
```

[...]

PR #6097, opened 2026-03-03
Prefer CurrentPage over query string in pager fix for #6086

leigh-pointer:
Update Pager.razor to prefer an explicit CurrentPage parameter from the parent and only fall back to PageState.QueryString['page'] if CurrentPage is not supplied. Parsing is consolidated into a nullable requestedPage and _page is set to requestedPage ?? 1, with a minimum enforced of 1. This prevents the query string from overriding an explicitly provided CurrentPage.

### DIFF
--- a/Oqtane.Client/Modules/Controls/Pager.razor
+++ b/Oqtane.Client/Modules/Controls/Pager.razor
@@ -374,6 +374,14 @@
     [Parameter]
     public string PaginationAlignment { get; set; } = "center"; // Alignment of the Page Numbering start, center, end
 
+    /// <summary>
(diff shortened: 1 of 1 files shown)


sbwalker:
@leigh-pointer currently some modules rely on the fact that the querystring overrides the CurrentPage value (ie. providing a deep link to a user which displays a specific page in the pager)  - so for those modules this PR would be a breaking behavioral change.

sbwalker:
@leigh-pointer I don't think this is the correct fix for #6086, as @markdav-is indicates in the issue description that he is using Interactive render mode... and in Interactive render mode there are no querystring parameters (querystring parameters are only used in Static render mode because there is no state).

PR #6106, opened 2026-03-05
add property to Module class to indicate how the IPortable interface is being invoked (Export Module, Import Module, Copy Page, Global Replace, Site Template)

sbwalker:
@iJungleboy this PR adds an IPortable property to the Module object which is passed to the IPortable interface and can be used to determine the scenario for which IPortable was invoked

### DIFF
--- a/Oqtane.Server/Controllers/PageController.cs
+++ b/Oqtane.Server/Controllers/PageController.cs
@@ -558,11 +558,13 @@ namespace Oqtane.Controllers
 
                             module = _modules.AddModule(module);
 
-                            // copy module content (includes settings)
-                            string content = _modules.ExportModule(pm.ModuleId);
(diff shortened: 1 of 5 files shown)


iJungleboy:
Awesome!

I find the naming a bit confusing - the property `IPortable` might not be the best choice. I thought I would point this out, as it's not out in the wild yet and could still be optimized.

Just as an idea: if you would instead use something like an `OperationContext` property which would be an object, then you would still have room to add more information later on, if more information is needed for certain actions. For example, something like

```csharp
class OperationContext
{
   public string OperationName {get;init;} // "Page Copy", etc.
   public TriggerEnum Trigger {get;init;} // something like "Manual" (human), "Schedule", "Automation" (eg. a clean-up routine)
  //  more props as needed in future
}
```

just an idea...

PR #6149, opened 2026-03-28
fix #6148 Localization site groups expose stale/incomplete Site.Langu…

### LINKED ISSUE #6148: [BUG] Localization site groups expose stale/incomplete `Site.Languages` per site, causing missing languages and incorrect `LanguageSwitcher` labels
### Oqtane Info

Version - 10.1.2
Render Mode - Static
Interactivity - Server
Database - SQL Server

### Describe the bug

In a localization site group, the `Site.Languages` collection is not consistent across all sites in the same group.

Example setup:
- primary site culture: `en-US`
- secondary site culture: `hr-HR`
- secondary site culture: `de-DE`

Image

Image

Image

The built-in `LanguageSwitcher` shows incorrect labels:

- on the primary site it shows English 3 times

Image

- on the Croatian site it shows Croatian once and English twice

Image

- on the German site it shows German, Croatian, and English correctly

Image

The links themselves point to the correct sites, but the language list and labels are inconsistent.

### Expected Behavior

All sites in the same localization site group should expose the same complete language list.

For the example above, every site in the group should show:
- English
- Croatian
- German

The `LanguageSwitcher` should display the correct language names on every site.

### Steps To Reproduce

1. Create 3 sites in the same tenant.
2. Set their culture codes to:
   - primary site: `en-US`
   - secondary site: `hr-HR`
   - secondary site: `de-DE`
3. Add all 3 sites to the same site group of type `Localization`.
4. Open each site in the frontend.
5. Check the built-in `LanguageSwitcher`.
6. Compare the languages shown on each site.

### Anything else?

It looks like `Site.Languages` is cached as part of the site snapshot, and when localization group membership or member site culture changes, not all site caches in the group are refreshed. This causes different sites in the same localization group to expose different snapshots of the language list.

This also affects any module or component which reads `PageState.Site.Languages`, not only the built-in `LanguageSwitcher`.

tvatavuk:
…ages per site, causing missing languages and incorrect LanguageSwitcher labels

Site languages can depend on localization-group membership, member cultures, and member aliases. Rebuild them for every request instead of relying on the cached site snapshot.

### DIFF
--- a/Oqtane.Server/Services/SiteService.cs
+++ b/Oqtane.Server/Services/SiteService.cs
@@ -101,6 +101,10 @@ namespace Oqtane.Services
             }
             site.Pages = pages;
 
+            // site languages can depend on localization-group membership, member cultures, and member aliases.
+            // Rebuild them for every request instead of relying on the cached site snapshot.
+            site.Languages = GetLanguages(site.SiteId, alias.TenantId);
+
             // get language display name for user
             foreach (Language language in site.Languages)
             {


sbwalker:
@tvatavuk I understand the problem however I do not believe this is the correct solution. Under normal circumstances the CultureCode for a Site will rarely ever be modified... so the solution of bypassing the cache and retrieving the languages on every request is not optimal as it imposes a significant performance impact for the entire framework rather than addressing this specific scenario. The real issue is that  the cache is scoped per Site, however when a CultureCode is changed for one Site it affects the list of Languages for other Sites (due to the grouping). I believe there is a better solution to this problem which I will submit as a new PR.

sbwalker:
replaced by #6152

PR #6152, opened 2026-03-30
fix #6148 - ensure cache is refreshed when CultureCode is modified

### LINKED ISSUE #6148: [BUG] Localization site groups expose stale/incomplete `Site.Languages` per site, causing missing languages and incorrect `LanguageSwitcher` labels

sbwalker:


sbwalker:
@tvatavuk this PR contains to identify when the CultueCode changes, and if so, it refreshes the cache for all Sites which are part of the current Tenant. This is still not as efficient as it could be, as it would be possible to query the SiteGroupMember to identify the specific Sites which are related to the current Site... however based on the fact that CultureCodes will rarely ever be modified, this solution should be fine in the majority of cases.

Issue #6190, opened 2026-04-24
[ENH] Improve Search User Experience in Pager


CliffConway:
### Oqtane Info

Version - any new version
Render Mode - any
Interactivity - any
Database - Sany

PR #6191, opened 2026-04-25
Improve pager search input handling - Fix for issue #6190

leigh-pointer:
Prevent default form submission and enable immediate input binding for the pager search. Replaced @bind with @bind-value/@bind-value:event="oninput" and added an @onkeydown handler to trigger Search() on Enter. Applied changes to both search form variants and added HandleSearchKeyDown to invoke Search when Enter is pressed and the query is non-empty.

### DIFF
--- a/Oqtane.Client/Modules/Controls/Pager.razor
+++ b/Oqtane.Client/Modules/Controls/Pager.razor
@@ -10,9 +10,16 @@
     {
         @if (!string.IsNullOrEmpty(SearchProperties))
         {
-            <form autocomplete="off">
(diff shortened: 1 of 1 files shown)


sbwalker:
@leigh-pointer this PR seems to have an issue which breaks the Search functionality in Static rendered components. See https://www.oqtane.org/blog for an example. I am investigating the cause.

sbwalker:
#6274 fixes this problem. The issue is that the name="_search" attribute was removed from the static rendered form.

Issue #6222, opened 2026-05-27
[ENH] Performance Improvement when handling Not Found


mdmontesinos:
### Oqtane Info

Version - 10.1.2
Render Mode - Static 
Interactivity - Server
Database - SQL Server

### Describe the enhancement

When a page is not found and the url mapping is retrieved, it currently makes an additional db call to get the site, only the determine whether to capture the broken url or not. For thousands of 404 requests made to a site (i.e. ai scrappers / bots), this stops being negligible.

You can see the traces from Azure in which this extra db call is shown:

<img width="884" height="426" alt="Image" src="https://github.com/user-attachments/assets/ee623b09-8739-42ef-962a-4a09d5334b14" />

<img width="1608" height="232" alt="Image" src="https://github.com/user-attachments/assets/91777243-dd79-4aa9-b8a6-f21f9ce5e429" />

However, as the url mapping for the page is retrieved in the App.razor of the Server, the site object is already available. Therefore, it can be passed as an argument to the url mapping repository to avoid this extra db call.

I've submitted PR #6223 to pass the already retrieved site's CaptureBrokenUrl option in the App.razor of the Server. 

Although the not found pages are the vast majority of 404 requests, I would like to also improve this for the Files server, but the site object is not available there. Perhaps a cache could be used to avoid requesting the site every time?

sbwalker:
@mdmontesinos based on the dramatic increase in bot traffic (since the introduction of GenAI) I think it makes sense to set Capture Broken Urls  to Disabled by default in new Oqtane installations. The original purpose of this feature was to provide a diagnostic tool for analyzing 404s... primarily after migrating to Oqtane from another CMS... which means that it really should be enabled on a periodic basis rather than enabled all the time. 

However, regardless of the default setting, the implementation still needs to be optimized as the method still needs to know if Capture Broken Urls is enabled or disabled. Your suggestion to pass this information to the repository method is a valid approach for avoiding the Site lookup. However, as you mentioned above, the information is not always available (ie. File handler). This brings up a higher level item which I classify as "technical debt"...

[...]

PR #6223, opened 2026-05-27
Avoid extra db call to retrieve site when page not found

mdmontesinos:
Issue described at #6222 

**URL Mapping Enhancements:**

* Updated the `GetUrlMapping` method in `IUrlMappingRepository` and its implementation to accept an optional `captureBrokenUrls` parameter, allowing explicit control over whether broken URLs should be captured when searching for a mapping

* Modified the logic in `GetUrlMapping` to use the provided `captureBrokenUrls` value if available, or fall back to the site's setting

* Changed the call to `UrlMappingRepository.GetUrlMapping` in `App.razor` to pass the site's `CaptureBrokenUrls` setting, avoiding the extra db call in this situation

* Incidentally, this gives more control to module developers, as they can use this parameter to explicitly override the site configuration when required for a specific functionality.

* This change does not affect url mappings retrieved from other places (PageController, UrlMappingController and Files server), which will still use the site's setting fallback.

### DIFF
--- a/Oqtane.Server/Components/App.razor
+++ b/Oqtane.Server/Components/App.razor
@@ -292,7 +292,7 @@
         string referrer = (Context.Request.Headers[HeaderNames.Referer] != StringValues.Empty) ? Context.Request.Headers[HeaderNames.Referer] : "";
 
         // page not found - look for url mapping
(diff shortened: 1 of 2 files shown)


sbwalker:
@mdmontesinos optional parameters should be avoided in API methods as although they appear convenient, they can cause longer term technical and architectural problems. I will merge the PR and remove the optional parameter.

PR #6298, opened 2026-07-11
Bump key NuGet dependencies

leigh-pointer:
Updates dependency versions across client, server, shared, and package specs to keep runtime and published package metadata aligned. This includes upgrading Radzen.Blazor, Swashbuckle.AspNetCore, Npgsql.EntityFrameworkCore.PostgreSQL, and NodaTime, and explicitly pinning SixLabors.ImageSharp to 3.1.12 for license compatibility.

### DIFF
--- a/Oqtane.Client/Oqtane.Client.csproj
+++ b/Oqtane.Client/Oqtane.Client.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk.BlazorWebAssembly">
+﻿<Project Sdk="Microsoft.NET.Sdk.BlazorWebAssembly">
 
   <PropertyGroup>
     <RootNamespace>Oqtane</RootNamespace>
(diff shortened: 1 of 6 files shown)


sbwalker:
@leigh-pointer this PR contains a regression issue - the Oqtane.Client.nuspec was modified with a Radzen.Blazor package version of 11.1.13 - which does not exist. This appears to be a typo as the Oqtane.Client.csproj was modified with Radzen.Blazor version 11.1.3. The impact of this issue is that Oqtane Applications which consume Oqtane.Client 10.2.2 as a Nuget dependency will not compile - they throw a build exception:

Unable to find package Radzen.Blazor with version (>= 11.1.13)
  - Found 1041 version(s) in nuget.org [ Nearest version: 11.1.7 ]
  - Found 0 version(s) in Microsoft Visual Studio Offline Packages
  - Found 0 version(s) in C:\Program Files\dotnet\library-packs

So the 10.2.2 release is essentially a DOA release, and will need to be replaced with a 10.2.3 release.

